PR #63, opened 2013-12-30
Database keeps changes through one class

GammaSQ:
Changes scope of fixture 'db' dynamically, depending on kwargs for pytest.mark.django_db.
Allows one class to keep all changes in databse and allows for incremental testing of database.

!!!! BUG !!!!

    @pytest.mark.django_db(class_db=True)
    class TestDB:
        # Tests now with classDB, but without transaction

        @pytest.mark.django_db(transaction=True)
        def test_smthg(self):
            # Tests with classDB and (newly activated) transactions:
       ##### mayjor breackdown!!!!!

    @pytest.mark.django_db(True, True)
    class TestBlah:
    ### no problem

Since I don't use transactions, I don't think I will fix this anytime soon. Furthermore it's a rather rare usecase I believe.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -60,7 +60,6 @@ def _django_db_setup(request, _django_runner, _django_cursor_wrapper):
 ################ User visible fixtures ################
 
 
-@pytest.fixture(scope='function')
 def db(request, _django_db_setup, _django_cursor_wrapper):
     """Require a django test database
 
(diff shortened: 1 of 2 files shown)


pelme:
I close this PR for now, having class scoped fixtures that can run in its own transaction/savepoint will be supported with the reworked internals of pytest-django. Once again - thanks for this patch and the discussions. Feel free to reopen this PR or open another issue if you have other comments or feedback!

PR #74, opened 2014-02-23
 multidb: let one connection be reused, and other not

mahomahomaho:
set per settings.py 

instead (besides) of set --reuse_db in cmdline, I can set TEST_REUSE in db setting, which let me reuse one db (postgres in this case) , which is big and expensive to create/copy, and create other (small sqlite).


blueyed:
Closing as per previous comment.

PR #98, opened 2014-06-06
xdist disable suffix

lestatcheb:
need if we want to use only one database for xdist


blueyed:
Should this be closed in favor of #336?

pelme:
this is fixed in pytest 3.0, see the documentation and raise new issues/PRs if needed!

PR #141, opened 2014-07-27
Add initial approach to use override_settings within 'settings' fixture. #136

Stranger6667:


pelme:
Closing this, a PR that uses override_settings is now merged into master.

Issue #85, opened 2014-04-28
Bug like `urls` marker does not restore original urlconf


AndrewPashkin:
Hi!
I've got bug using `urls` marker which alternate urlconf with secondary urlconf - tests that runs after test using it - cant reverse urls from main urlconf.
I dont know how to correctly test this thing where one test should be launched after another, and I just created test Django project that reproduces that error:

https://github.com/AndrewPashkin/pytest_urlsconf_weird

thedrow:
This should be closed no?

thedrow:
Actually I think there's a regression since I see this happening in pytest-django 2.9.1 as well.

blueyed:
@thedrow 
Can you provide more information, e.g. a failing test case?

thedrow:
@blueyed Clone https://github.com/getdrifter/drifter-server and run the tests using tox.

thedrow:
@blueyed Would you rather see the tests output first?

blueyed:
@thedrow 
Sorry, I have no time for this currently.
The idea would be to have a pull request for pytest-django with some failing test and then work on a fix from there.
Especially since the external information provided by @AndrewPashkin has vanished in the meantime.

thedrow:
I'll provide a failing test case but I already looked at it and I have no idea what the fix is.

thedrow:
@blueyed Sorry it took me so long. See #321

blueyed:
It looked to me like the failing test was wrong.  See https://github.com/pytest-dev/pytest-django/pull/355 for a followup, with what I think the test should look like.

I am closing this issue for now: let's continue on #355.

PR #148, opened 2014-07-29
Restoration after `urls` marker (#85)

Stranger6667:
I'm not sure is it the right way or not :( 

Ref: #85


pelme:
This is fixed in master!

PR #166, opened 2014-09-03
Skip monkey-patching for South with Django 1.7

blueyed:
This fixes the following error when South tries to access
`management._commands`, which has been dropped in Django 1.7:

    …/pytest_django/pytest_django/fixtures.py:41: in _django_db_setup
        _handle_south()
    …/pytest_django/pytest_django/fixtures.py:128: in _handle_south
        patch_for_test_db_setup()
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

        def patch_for_test_db_setup():
            # Load the commands cache
            management.get_commands()
            # Repoint to the correct version of syncdb
            if hasattr(settings, "SOUTH_TESTS_MIGRATE") and not settings.SOUTH_TESTS_MIGRATE:
                # point at the core syncdb command when creating tests
                # tests should always be up to date with the most recent model structure
                management._commands['syncdb'] = 'django.core'
            else:
    >           management._commands['syncdb'] = MigrateAndSyncCommand()
    E           AttributeError: 'module' object has no attribute '_commands'

    …/site-packages/south/management/commands/__init__.py:34: AttributeError

I wonder if we should output some warning here, because South is not meant to be used with Django 1.7?!

The patch/diff looks huge, but is mostly whitespace / indentation.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -94,39 +94,43 @@ def _django_db_fixture_helper(transactional, request, _django_cursor_wrapper):
 
 def _handle_south():
     from django.conf import settings
-    if 'south' in settings.INSTALLED_APPS:
-        # Handle south.
-        from django.core import management
 
-        try:
-            # if `south` >= 0.7.1 we can use the test helper
-            from south.management.commands import patch_for_test_db_setup
-        except ImportError:
-            # if `south` < 0.7.1 make sure it's migrations are disabled
-            management.get_commands()
-            management._commands['syncdb'] = 'django.core'
-        else:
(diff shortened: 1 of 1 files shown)


blueyed:
I've gone ahead and merged it.

PR #170, opened 2014-10-01
setup.py: Mention the 3-clause BSD license in our metadata

wking:
I don't know if PyPI has a convention for naming these, but that's
what the Open Source Initiative calls this one [1].  However, the text
is a bit different from the official upstream version which has:

  3. Neither the name of the copyright holder nor the names of its
     contributors may be used...

instead of the local:

  * The names of its contributors may not be used...

There's also an "COPYRIGHT HOLDER" -> "COPYRIGHT OWNER" change in the
local disclaimer.

It would be nice if we could stick to the standard upstream phrasing.

[1]: http://opensource.org/licenses/BSD-3-Clause

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ setup(
     maintainer="Andreas Pelme",
     maintainer_email="andreas@pelme.se",
     url='http://pytest-django.readthedocs.org/',
+    license='BSD-3-Clause',
     packages=['pytest_django'],
     long_description=read('README.rst'),
     install_requires=['pytest>=2.5'],


blueyed:
Thanks for your PR.

I think it's good to be merged as-is.

What's your opinion / recommendation on the differences you've found in the license file (https://github.com/pytest-dev/pytest-django/pull/170#issuecomment-57402153)?

wking:
On Mon, Mar 02, 2015 at 04:03:36PM -0800, Daniel Hahler wrote:
> I think it's good to be merged as-is.

Me too.

> What's your opinion / recommendation on the differences you've found
> in the license file

I'm not a lawyer, but I'd guess it doesn't really matter.  I'd
probably just leave it alone unless someone with a more legal
knowledge comes around.

wking:
Both the Open Source Initiative and Wikipedia agree on “Neither the
name of the {copyright holder|<organization>} nor the names of its
contributors…” though, so if it was my project, I'd use the “Neither
the name of the copyright holder nor the names of its contributors…”
wording (and while I was touching the file, I'd just replace it
entirely with the Open Source Initiative text).  Still, for this
project the copyright holders and contributors are the same thing, so
it probably doesn't really matter either.

blueyed:
Thanks for the additional information.

I have merged it, but re-open it for @pelme to review / accept it, too.

PR #179, opened 2014-10-10
Example for usage of rf, admin_user and class-based views

mfa:


blueyed:
Thanks, this looks good.
I'll merge it later (with the commits squashed).

docs/helpers.rst
```diff
@@ -148,6 +148,41 @@ case there is no "admin" user yet).
...
+        return thing_object
+
+    def test_detail_view_logged_in(rf, admin_user, thing):
+        # set kwargs for reverse and for view
+        kwargs_thing = {
+            'pk': thing.id,
+        }
+
+        url = reverse("thing_detail", kwargs=kwargs_thing)
+        # bind url to request factory
```
pelme: A newline before the comment lines would make this easier to read

pelme:
Thanks for the patch, it looks good! Examples that uses multiple fixtures is indeed very useful!

I think the example should be under a section called "Examples" or something like that (currently it is placed under the `admin_client` fixture which I guess is not really the indented place)

blueyed:
@mfa 
Can you address the comment from Andreas, please - and then squash it yourself already?

PR #182, opened 2014-10-17
Fix clobbering of real database by tests

qris:
Force disconnection if a database is already connected when patching for DB reuse.

DjangoCMS opens database connections very early, to set up URL routing, before
pytest has time to patch the database name to use the test database. Thus the
connection is already open and changing the NAME has no effect, so the real
database gets clobbered instead of the test one.

This isn't easy to trigger - I've only just noticed it and I've been using
DjangoCMS together with pytest-django for several projects. I don't yet know
what causes it. But because it's so catastrophic, I think we ought to put some
protection against it in place, hence this patch.

### DIFF
--- a/pytest_django/db_reuse.py
+++ b/pytest_django/db_reuse.py
@@ -20,6 +20,7 @@ def test_database_exists_from_previous_run(connection):
 
     orig_db_name = connection.settings_dict['NAME']
     connection.settings_dict['NAME'] = test_db_name
+    connection.connection = None # force reconnection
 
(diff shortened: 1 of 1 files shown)


pelme:
From what I understand, this PR/discussion deals with errors that is caused by doing database queries, url lookups and other stuff at import time. These problems has been fixed by Django and its app refactor.

We have also made the database checking stricter in the latest version.

I'm therefore going to close this PR, feel free to open new issues or a new PR against the latest master if these problems still persists.

PR #184, opened 2014-10-30
is_django_unittest gives false negatives during pytest_collection_modifyitems

paulcollinsiii:
`django_compat.is_django_unittest` would give false negatives during test collection. This made it hard for me to filter db tests from non-db test.

I'm not thrilled with what amounts to putting a test inside the conftest module, but I'm not sure if there's another way to test this? The `@pytest.fixture` I added was passing with the original code. Only the change to conftest exposed this problem (which I how I ran across it in the first place).

### DIFF
--- a/pytest_django/django_compat.py
+++ b/pytest_django/django_compat.py
@@ -1,6 +1,5 @@
 # Note that all functions here assume django is available.  So ensure
 # this is the case before you call them.
-import sys
 
 
 def is_django_unittest(item):
--- a/pytest_django/django_compat.py
+++ b/pytest_django/django_compat.py
@@ -10,13 +10,7 @@ def is_django_unittest(item):
     except ImportError:
         from django.test import TestCase
 
-    if not hasattr(item, 'obj'):
+    if not hasattr(item, 'cls') or item.cls is None:
         return False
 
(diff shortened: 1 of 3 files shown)


blueyed:
It looks like (the logic from your 799f42e9) has been added in fa186048 by @pelme later.

I've skipped it when rebasing your branch, and ended up with:
https://github.com/blueyed/pytest_django/compare/paulcollinsiii-is_unittest

The approach through conftest is interesting.

Wouldn't something using `runpytest` work, too?

paulcollinsiii:
@blueyed not sure. Since the logic seems to be in there though (I'd honestly forgotten about this PR!) I'm fine closing this.

blueyed:
@paulcollinsiii 
I think it might be useful to merge / use your test maybe, but am not sure.

@pelme What do you think?
https://github.com/blueyed/pytest_django/compare/paulcollinsiii-is_unittest

pelme:
Rather than focusing on unit testing `is_unittest` function itself, I think these kinds of tests is better served by explicit test cases which invokes pytest with testdir.

I think the root issue is fixed, if not, please open a new issue and I'd be happy to continue the discussion on how to best test this.

PR #188, opened 2014-12-19
Create --nomigrations option to disable Django 1.7 migrations

renanivo:
When there are several migrations to run, the database setup can be very slow. This pull request uses [a hack](https://gist.github.com/NotSqrt/5f3c76cd15e40ef62d09) to disable Django 1.7 migrations when the option `--nomigrations` is given.

It should not be a problem when you use `--reuse-db` in your workflow, but the `--nomigrations` flag should help who needs / prefers to create the database on every run.

It seems that [until Django 1.9](https://groups.google.com/forum/#!topic/django-developers/PWPj3etj3-U) is safe to use this hack -- I hope we have another way to do it then. 

There's also a [plugin for manage.py](https://github.com/henriquebastos/django-test-without-migrations) to do it.

My tests took 120 seconds before and are taking just 10 seconds now, so I hope it gets merged.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -135,7 +135,7 @@ def _handle_south():
 def _disable_native_migrations():
     from django import get_version
 
-    if get_version() > '1.7':
+    if get_version() >= '1.7':
         from django.conf import settings
         from .migrations import DisableMigrations
 
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -40,6 +40,9 @@ def _django_db_setup(request,
 
     _handle_south()
 
+    if request.config.getvalue('nomigrations'):
+        _disable_native_migrations()
+
     with _django_cursor_wrapper:
         # Monkey patch Django's setup code to support database re-use
         if request.config.getvalue('reuse_db'):
@@ -127,6 +130,16 @@ def _handle_south():
             south.hacks.django_1_0.SkipFlushCommand = SkipFlushCommand
 
             patch_for_test_db_setup()
+
+
+def _disable_native_migrations():
+    from django import get_version
+
+    if get_version() > '1.7':
+        from django.conf import settings
+        from .migrations import DisableMigrations
+
+        settings.MIGRATION_MODULES = DisableMigrations()
 
 
 # ############### User visible fixtures ################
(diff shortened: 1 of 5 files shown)


pelme:
Thanks for a very nice and complete PR! This patch looks truly exemplary :+1:  I've just gotten the master branch back to a usable state and hope to review these changes properly and merge+release them ASAP. Again, thanks for a great PR!

renanivo:
I'm really glad you like it. 

When you stabilize the master branch, tell me if you need any help to merge this PR.

renanivo:
Hello again @pelme,

I merged your last updates to master and all tests are passing :smiley:.
I took the opportunity to fix a typo that I introduced in the docs. Take a look to see if it fits.

pytest_django/fixtures.py
```diff
@@ -129,6 +132,16 @@ def handle_noargs(self, **options):
             patch_for_test_db_setup()
 
 
+def _disable_native_migrations():
+    from django import get_version
+
+    if get_version() > '1.7':
```
pelme: Shouldn't this be `get_version() >= '1.7'`? 1.7.1 will still be greater than 1.7, and I guess few people run 1.7, but still :)

pelme:
Sorry for the delay and thanks again for the great PR! Everything looks good! My week is busy, but my plan is to merge this and make a new release on sunday (18th january).

pytest_django/fixtures.py
blueyed: Maybe it could be changed to return for `< '1.7'` instead, or try to import `DisableMigrations`.

renanivo: I updated to `get_version() >= '1.7'`. The @blueyed suggestion seems to be good too. I can do it if you prefer.

renanivo:
Thanks for the code review. Take your time to merge it.

alvinchow86:
this will be really useful! unfortunately it doesn't seem to work with pytest-xdist 
ran these commands
 `py.test -n2 --create-db`
 `py.test -n2 --create-db --nomigrations`

both take the same amount of time (with --nomigrations it still seems to have that lengthy pause before tests start executing). I confirmed it does work without using xdist/multiple processes though.

I will see if I can do some debugging

alvinchow86:
oh wait actually I think it does work, my local environment was a bit messed up. (well at least it's tested with xdist now!)

pelme:
@renanivo Again, thanks for this PR, it is now merged and released on PyPI!

renanivo:
Good to know. Thanks.

Issue #146, opened 2014-07-27
Support settings DJANGO_SETTINGS_MODULE in pytest_configure


pelme:
See comment in #119, this should be possible:

```python
import os

def pytest_configure():
    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'myproject.settings')
```

julen:
> *configuring DJANGO_SETTINGS_MODULE in pytest_configure is not really supported/tested at the moment, but should be easy to support properly*.

How easy is it? I need this to switch to Django 1.7 + pytest-django 2.7, otherwise I'm forced to call `setup()` manually.

pelme:
Actually, I think this works and the issue should be closed. Could you give it a try if you haven't?

There doesn't seem to be a test for this case, so that should probably be added to make sure it doesn't accidentally break.

julen:
I still get the same `INTERNALERROR` traceback:
```python
INTERNALERROR> AssertionError: assert <django.conf.LazySettings object at 0x10eb0e4e0>.configured is True
```

I added a test in #192. If you import `django.setup` and call it after setting `DJANGO_SETTINGS_MODULE`, the test passes.

julen:
Any updates @pelme?

blueyed:
The problem appears to be that pytest-django sets up Django after any project's conftest:

    @pytest.mark.trylast
    def pytest_configure():
        if django_settings_is_configured():
            _setup_django()

([Source](https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/plugin.py#L203-206))

Should be fixed via PR #216.

PR #192, opened 2015-01-04
Added test for setting DSM in pytest_configure

julen:
This adds a test for #146.

### DIFF
--- a/tests/test_django_settings_module.py
+++ b/tests/test_django_settings_module.py
@@ -96,6 +96,25 @@ def test_ds_after_user_conftest(testdir, monkeypatch):
     result.stdout.fnmatch_lines(['*1 passed*'])
 
 
+def test_ds_in_pytest_configure(testdir, monkeypatch):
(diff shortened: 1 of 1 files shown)


julen:
Rebased on top of master, note that this test is meant to fail until #146 is fixed. @pelme can you have a look at this, please?

pelme:
Thanks for the test, it will make the fix much easier. I have a very busy week and will try to look into this as soon as I can. I wanted to get 2.8.0 out quick to support Django 1.8 so this unfortunately had to wait for the moment.

pelme:
(If you or anyone else want to work on this til I can look into it, please go ahead :))

blueyed:
The fix for this might be simple: just remove the assertion, which is not correct in this case.

I've created a PR with the test and the fix at https://github.com/pytest-dev/pytest-django/pull/216.

blueyed:
Closing this in favor of #216, which includes the test and fix.

julen:
Fair enough, and thanks for the fix!

PR #196, opened 2015-01-20
Add TestUnittestMethods

blueyed:
Not sure how useful this really is, but it ensures that the methods are
being called actually.


pelme:
I think this is a potentially useful test since the code that interacts with pytest/Django to handle this properly is very hard to comprehend. :)

Issue #190, opened 2014-12-29
InternalError and AssertionError with non-django tests


klvbdmh:
After installing pytest-django I cannot run my old tests on a simple scrapper. The scrapper's project consists of five files: `__init__.py`, `scrap.py`, `test_scrap.py`, `output.txt`, and `test.txt`. The test works by comparing the two text files. As you can see it's a very simple project and it doesn't use django at all.

When I run `test_scrap.py` with pytest I get the following error:

```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\main.py", line 80, in wrap_session
INTERNALERROR>     config.do_configure()
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\config.py", line 618, in do_configure
INTERNALERROR>     self.hook.pytest_configure(config=self)
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\core.py", line 413, in __call__
INTERNALERROR>     return self._docall(methods, kwargs)
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\core.py", line 424, in _docall
INTERNALERROR>     res = mc.execute()
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\core.py", line 315, in execute
INTERNALERROR>     res = method(**kwargs)
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\pastebin.py", line 14, in pytest_configure
INTERNALERROR>     __multicall__.execute()
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest-2.6.4-py3.4.egg\_pytest\core.py", line 315, in execute
INTERNALERROR>     res = method(**kwargs)
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest_django-2.7.0-py3.4.egg\pytest_django\plugin.py", line 203, in
pytest_configure
INTERNALERROR>     if django_settings_is_configured():
INTERNALERROR>   File "C:\Python\lib\site-packages\pytest_django-2.7.0-py3.4.egg\pytest_django\lazy_django.py", line 26,
 in django_settings_is_configured
[...]

PR #200, opened 2015-01-28
TBD: Add (failing) tests for "django" imported, but no DSM 

blueyed:
Ref: https://github.com/pytest-dev/pytest-django/issues/190


blueyed:
Merged to master in 7c84e7f.

PR #201, opened 2015-01-29
Cleanup interaction with `live_server` and `transactional_db`

blueyed:
This also removes a test marked as `xfail`, which is not supposed to fail as far as I understand.

@pelme 
Can you confirm that the test should not have been expected to fail in the first place?


pelme:
This looks good to me. I am not sure about the historical reasons, but I think these changes makes sense.

PR #203, opened 2015-02-08
WIP: use setup/teardown from Django (TransactionTestCase)

blueyed:
Use `_pre_setup` and `_post_teardown` from Django also for `TransactionTestCase`.

This would provide additional features, and makes the behavior more consistent.

It does not work with Django < 1.5 however, which is now special-cased.

I'm going to add some test(s) for the new behavior later.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -64,32 +64,44 @@ def _django_db_fixture_helper(transactional, request, _django_cursor_wrapper):
     if is_django_unittest(request):
         return
 
+    if not transactional and 'live_server' in request.funcargnames:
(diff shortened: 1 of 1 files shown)


pelme:
This looks very good, using TransactionTestCase directly is much better than doing the flushing by ourselves. Is there any need for new tests for this?

Shouldn't the existing tests for transactional tests should be enough?

blueyed:
> Shouldn't the existing tests for transactional tests should be enough?

I'm not sure if the existing tests are specific enough, but this can be improved / fixed when any issues are being reported from this.

I'll merge it for now.

PR #205, opened 2015-02-10
tests: add assertions for `ret` to runpytest calls

blueyed:


pelme:
Looks good to me! :)

blueyed:
Merged in https://github.com/pytest-dev/pytest-django/commit/c380ae23179209804c3102584489e69daca8851d (and fixed in 145299d).

PR #207, opened 2015-02-10
Travis: test all Django versions against Python 2 and 3

### LINKED ISSUE #206: Travis: tests should include all Django versions on Python 2 and 3
It looks like the matrix generation from generate-configurations does not make sure that every Django version is tested on Python 2 and 3: https://travis-ci.org/pytest-dev/pytest-django/builds/50237861

It would be better to see that Django 1.5 also fails on Python 2.7.

blueyed:
Fixes #206.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,10 @@ env:
   - TESTENV=python2.6-1.6-sqlite_file
   - TESTENV=python2.7-1.3-sqlite_file
   - TESTENV=python2.7-1.4-sqlite_file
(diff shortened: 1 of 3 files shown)


pelme:
Looks good to me, it doesn't really add that many more envs to Travis and it would be good to run them all. :-)

blueyed:
Cool.  Rebased and will merge it when tests pass.

pelme:
(The tests passed so I hit the merge button :-))

blueyed:
Well done! :)

PR #208, opened 2015-02-10
Travis improvements

blueyed:
`sudo: false` makes sense (to enable a container based environment, which is faster and has more features).

But I don't now if we want to allow failure of Django master?

Another interesting option (when we'll use the allowed failures) is `fast_finish`: http://docs.travis-ci.com/user/build-configuration/#Fast-finishing


blueyed:
@pelme 
> But I don't now if if want to allow failure of Django master?

What's you opinion on this?

blueyed:
I am going to merge this for now - the questionable part can be adjusted / reverted later.

PR #209, opened 2015-02-10
Pass on pytest.config.option.verbose to db setup / flush

blueyed:
This is a followup to a00a8e4.


pelme:
Looks good! The messages that appears has annoyed me a lot of times, but I've never gotten around to fixing it.

blueyed:
Re-based and fixed a test (looking for `Creating table app_item` now).

PR #213, opened 2015-02-26
db fixture: fixed docstring

julen:
It doesn't look like the `db` fixture docstring makes much sense. First it says it will setup the DB without transaction support and then it says it will roll back changes at the end of the transaction. My understanding is there are no transactions and therefore the second sentence of the existing docstring must go away.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -146,11 +146,12 @@ def _disable_native_migrations():
 def db(request, _django_db_setup, _django_cursor_wrapper):
     """Require a django test database
 
-    This database will be setup with the default fixtures and will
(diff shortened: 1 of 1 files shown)


blueyed:
It's a bit confusing, yes, but in general it means that there is an outer transaction that gets rolled back, and the tests cannot use transactions themselves.

julen:
Thanks for clarifying. I have updated the text accordingly, hopefully it's clearer now!

blueyed:
:+1: 

But I'll wait for someone else to confirm, given how tricky it is.. :)

pelme:
Sounds good to me, thanks for the fix! :)

PR #216, opened 2015-03-02
Test and fix for issue #146

### LINKED ISSUE #146: Support settings DJANGO_SETTINGS_MODULE in pytest_configure

blueyed:
Fixes #146.


blueyed:
@pelme 
I think this is good, but since it removes some assertion, I'd like you to verify it.

pelme:
I cannot recall why that assertion was added, but I think it should be fine since all tests passes. I'll merge it now, @julen please try it out and open an issue if it doesn't work properly.

PR #219, opened 2015-03-14
Use .tar.gz instead of .zip to work around pip bug

blueyed:
This works around https://github.com/pypa/pip/issues/1698, where pip
fails to install the zip with unicode filenames on Python 2.


blueyed:
Cherry-picked into master as 1e72766.

PR #221, opened 2015-03-19
WIP: use "db" with "live_server", not "transactional_db"

blueyed:
"transactional_db" is slower and has drawbacks.  Using "db" instead
seems to work.

See also https://code.djangoproject.com/timeline?from=2014-10-13T05%3A15%3A40-05%3A00&precision=second.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -156,8 +156,7 @@ def db(request, _django_db_setup, _django_cursor_wrapper):
     database setup will behave as only ``transactional_db`` was
     requested.
     """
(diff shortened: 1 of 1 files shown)


blueyed:
It is required after all.. but I've found some local tests that obviously don't need it.

blueyed:
FWIW, it might be possible for/with SQLite's in-memory-db, which then used `connection.allow_thread_sharing`.

Might be worth to improve it for this case, which appears to be often used in tests.

blueyed:
Closing this for now.

PR #222, opened 2015-03-22
Adds tests for invalid template variables

codingjoe:
Django catches all `VariableDoesNotExist` exceptions to replace
them in templates with a modifiable string that you can define in
your settings.
Sadly that doesn't allow you to find them in unit tests.

`_fail_for_invalid_template_variable` sets the setting
`TEMPLATE_STRING_IF_INVALID` to a custom class that not only fails
the current test but prints a pretty message including the template
name.

A new marker allows disabling this behavior, eg:

    @pytest.mark.ignore_template_errors
    def test_something():
        pass

This marker sets the setting to None, if you want it to be a string,
you can use the `settings` fixture to set it to your desired value.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -5,6 +5,8 @@ test database and provides some useful text fixtures.
 """
 
 import contextlib
+import inspect
+from functools import reduce
 import os
 import sys
 import types
@@ -27,6 +29,7 @@ from .lazy_django import django_settings_is_configured, skip_if_no_django
 
 SETTINGS_MODULE_ENV = 'DJANGO_SETTINGS_MODULE'
 CONFIGURATION_ENV = 'DJANGO_CONFIGURATION'
+INVALID_TEMPLATE_VARS_ENV = 'FAIL_INVALID_TEMPLATE_VARS'
 
 
 # ############### pytest hooks ################
@@ -62,6 +65,12 @@ def pytest_addoption(parser):
                   'Automatically find and add a Django project to the '
                   'Python path.',
                   default=True)
+    group._addoption('--fail-on-template-vars',
+                     action='store_true', dest='itv', default=False,
+                     help='Fail for invalid variables in templates.')
+    parser.addini(INVALID_TEMPLATE_VARS_ENV,
+                  'Fail for invalid variables in templates.',
+                  default=False)
 
 
 def _exists(path, ignore=EnvironmentError):
@@ -169,6 +178,14 @@ def pytest_load_initial_conftests(early_config, parser, args):
         _django_project_scan_outcome = _add_django_project_to_path(args)
     else:
         _django_project_scan_outcome = PROJECT_SCAN_DISABLED
+
+    # Configure FAIL_INVALID_TEMPLATE_VARS
+    itv = (options.itv or
+           os.environ.get(INVALID_TEMPLATE_VARS_ENV) in ['true', 'True', '1'] or
+           early_config.getini(INVALID_TEMPLATE_VARS_ENV))
+
+    if itv:
+        os.environ[INVALID_TEMPLATE_VARS_ENV] = 'true'
 
     # Configure DJANGO_SETTINGS_MODULE
     ds = (options.ds or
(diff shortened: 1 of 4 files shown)


codingjoe:
I honestly don't know why the tests for 1.8 are failing. Does anyone have a clue to why the marker does not work?

blueyed:
This changed in 1.8: https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/

It should only be deprecated, but maybe there's a bug with that, so that e.g. `TEMPLATE_STRING_IF_INVALID` does not get mapped properly.

blueyed:
The problem is likely that you would need to change the `TEMPLATE` setting for Django 1.8+, because you configure it after Django has applied its translation into the new setting.

codingjoe:
@blueyed thanks, I'll look into that and tie it up.

codingjoe:
@blueyed I fixed the django 18 issue and added documentation. Have fun!

pytest_django/plugin.py
```diff
@@ -327,6 +329,86 @@ def restore():
         request.addfinalizer(restore)
 
 
+@pytest.fixture(autouse=True, scope='session')
+def _fail_for_invalid_template_variable(request):
+    """Fixture that fails for invalid variables in templates.
+
+    This fixture will fail each test that uses django template rendering
+    should at template contain an invalid template variable.
```
blueyed: s/at/a/

pytest_django/plugin.py
```diff
@@ -327,6 +329,86 @@ def restore():
...
+
+    It does not raise an exception, but fails, as the stack trace doesn't
+    offer any helpful information to debug.
```
blueyed: This should mention that it fails with a message, possibly mentioning the template, but at least the variable.
My first impression from reading was that it would "just fail".

pytest_django/plugin.py
```diff
@@ -327,6 +329,86 @@ def restore():
...
+    This behavior can be switched of using the marker:
+    ``ignore_template_errors``
+    """
+    class InvalidVarException(object):
+
```
blueyed: Delete this line, please.

pytest_django/plugin.py
```diff
@@ -327,6 +329,86 @@ def restore():
...
+            if template:
+                msg = "Undefined template variable '%s' in '%s'" % (var, template.name)
+            else:
+                msg = "Undefined template variable '%s'" % var
+            if self.fail:
+                pytest.fail(msg, pytrace=False)
+            else:
+                return msg
+
+    if django_settings_is_configured():
```
blueyed: I suggest moving this into a function, and using it from _template_string_if_invalid_marker, too - giving a flag to enable/disable it.

blueyed:
Very nice, I've left a few comments.

pytest_django/plugin.py
codingjoe: I can't the problem is, that in the first time I set the setting. Than all template settings get copied into to engine. That's why you can't change the template setting during runtime.
That's why in the second very similar looking snippet, I use the hashable quality of an object to change it state. Which also changes the engines own store as well.

I could define a singleton tho. Than I would only need to change the state in the marker. I'll look into that.

codingjoe: Scratch that, It's even more ugly. I'd prefer to leave it as is.

codingjoe:
@blueyed thanks, I updated my commit.

pelme:
This would be very useful!

I think this should be opt-in (maybe as a pytest.ini setting+marker for local overrides?). Just enabling it for everyone will probably break a lot of existing tests.

Maybe this (`InvalidVarException`) can be contributed back to Django somehow too, maybe it can be useful outside of pytest too.

pytest_django/plugin.py
blueyed: Ok then.  Thanks for explaining and looking into it!

codingjoe:
@pelme I don't really think that Django has any interest in that patch. They actually have an exception that they always catch.

Regarding opt-in/out: I think this is an exceptable change for a minor version. And yes, some people will have to fix some templates upgrading. But only if there templates are wrong ;)

If you want me to make this an opt-in feature, let me know, I'll update my code and doc.

pelme:
This is similar to `DEBUG_PROPAGATE_EXCEPTIONS = True` that used to be set by pytest-django, but it broke peoples test suites: #26. Yes, Django has some bad defaults here, and everyone should be better of by fixing these problems, but just changing this and breaking existing test suites is not nice. :/

Therefore I think that this feature must be opt-in, but it is certainly something I'd like to use myself.

codingjoe:
@pelme sure, I'll update my PR

codingjoe:
@blueyed @pelme there you go. I added a command line option `--test-templates` to opt-in.

blueyed:
@codingjoe 
Thanks!  I'd like to see a (config file) option for this, too.

pelme:
@codingjoe  Very nice!

What about naming the option a bit less generic than `--test-templates`, maybe something like `--fail-on-template-vars` ?

codingjoe:
Sure, will do. Might actually be a good Idea as I'm thinking about writing more template tests like a py.test wrapper for tidy-html5.

codingjoe:
done

blueyed:
Repeating my comment from above: can we have a config/ini file option for this, too? (via `parser.addini`)
I think it makes sense to be able to enable this in your project config.

codingjoe:
@blueyed all done.

pytest_django/plugin.py
```diff
@@ -170,6 +179,14 @@ def pytest_load_initial_conftests(early_config, parser, args):
     else:
         _django_project_scan_outcome = PROJECT_SCAN_DISABLED
 
+    # Configure FAIL_INVALID_TEMPLATE_VARS
+    itv = (options.itv or
+           early_config.getini(INVALID_TEMPLATE_VARS_ENV) or
+           os.environ.get(INVALID_TEMPLATE_VARS_ENV))
```
blueyed: This should be changed to prefer the environment variable before the config, see https://github.com/pytest-dev/pytest-django/pull/199.

pytest_django/plugin.py
```diff
@@ -170,6 +179,14 @@ def pytest_load_initial_conftests(early_config, parser, args):
...
+
+    if itv:
+        os.environ[INVALID_TEMPLATE_VARS_ENV] = 'True' if itv else None
```
blueyed: There is a duplicated check for `itv` here.

codingjoe:
@blueyed I swapped the config.

codingjoe:
@blueyed oh yes, sorry, I fixed that.

blueyed:
> @blueyed oh yes, sorry, I fixed that.

I think it might be good if it would use "1" for the environment variable, and allow "0" to disable it.
Currently it seems to be "set and non-empty" => True, otherwise False.
Seems to be confusing to use "True", if "False" would be also True.

codingjoe:
@blueyed I hope that does fixes the issue. Please feel free to fix any typos while merging. I'm dyslexic and going into a 20th-ish review cycle is somewhat disencouraging.

codingjoe:
@blueyed @pelme anything that is stopping this from being merged?

pelme:
@codingjoe Thanks a lot for fixing all the comments and sticking with it. This is a great feature that I will use myself. Sorry for the slow response, but now it is merged and will be included in the next version (which probably will be out during/after this weekend).

blueyed:
@codingjoe 
I'm also sorry for forgetting about this.
And thanks indeed for sticking with me/us and providing this great feature! :tulip:

pytest_django/plugin.py
```diff
@@ -327,6 +344,88 @@ def restore():
...
+
+        def __mod__(self, var):
+            """Handle TEMPLATE_STRING_IF_INVALID % var."""
+            template = self._get_template()
+            if template:
+                msg = "Undefined template variable '%s' in '%s'" % (var, template.name)
+            else:
+                msg = "Undefined template variable '%s'" % var
+            if self.fail:
+                pytest.fail(msg, pytrace=False)
```
blueyed: @codingjoe 
Do you remember why you used `pytrace=False` here?
I think it is very valuable to have a traceback in case of errors, isn't it?
(it might need to be massaged a bit to make it shorter though).

codingjoe: Nope, I don't. I think back then there trace didn't give you much information, because the template engine worked differently. That was before you could switch engines, if I remember correctly.

blueyed: @codingjoe 
Thanks for your feedback, enabled it in 760fb18.

PR #224, opened 2015-03-29
Use `travis_retry` in tox (via install_command), not on tox

blueyed:


pelme:
Very nice!

However, won't this affect local tox runs too? I run tox locally from time to time to run a specific configuration

pelme:
Very nice with the wrapper!

blueyed:
> However, won't this affect local tox runs too?

Yes..

The wrapper idea turned to be more complicated than initially hoped, but should work now.

pelme:
I have a hard time doing any meaningful review by looking at the diff, but feel free to merge this if it works :)

PR #225, opened 2015-04-02
Add a runner consumable by TEST_RUNNER

joeyespo:
This allows you to use Django's standard `manage.py test` to run your tests instead of making your collaborators switch to a new command. This can also help open source projects by allowing contributors to use the existing Django conventions they're already used to, since the alternative is that they run `manage.py test` normally and it shows ""

The implementation provides a light-weight translation of `manage.py test` options to `py.test` options. This includes `--verbosity 0` => `--quiet`, `--verbosity 2` => `--verbose`, `--failfast` => `--exitfirst`, and the upcoming Django 1.8 `--keepdb` => this project's `--nomigrations` equivalent.

Use it by adding the following line to your Django settings.py:

```py
TEST_RUNNER = 'pytest_django.runner.PyTestRunner'
```

Now you can run the familiar `manage.py test` and switch to `py.test` when you're ready.

### DIFF
--- a/pytest_django/runner.py
+++ b/pytest_django/runner.py
@@ -1,0 +1,61 @@
+import os
+from optparse import make_option
+
+import pytest
+
+
+class PyTestRunner(object):
+    """
+    Runs py.test to discover and run tests.
+    """
+    option_list = (
+        make_option('-t', '--top-level-directory',
+            action='store', dest='top_level', default=None,
+            help='Top level of project for unittest discovery.'),
+        make_option('-p', '--pattern', action='store', dest='pattern',
+            default="test*.py",
+            help='The test matching pattern. Defaults to test*.py.'),
+        make_option('-k', '--keepdb',
+            action='store_true', dest='keepdb', default=False,
+            help='Preserves the test DB between runs.'),
+    )
+
+    def __init__(self, pattern=None, top_level=None, verbosity=1,
+                 interactive=True, failfast=False, keepdb=False, **kwargs):
+        if pattern and pattern != 'test*.py':
+            # TODO: implement
+            raise NotImplementedError('Testing with a file pattern is not '
+                                      'implemented.')
+        if top_level is not None:
+            # TODO: implement
+            raise NotImplementedError('Specifying the top_level is not '
+                                      'implemented.')
+        self.verbosity = verbosity
+        self.interactive = interactive
+        self.failfast = failfast
+        self.keepdb = keepdb
+
+        # TODO: Make this an option
+        self.ds = os.environ['DJANGO_SETTINGS_MODULE']
+
+    def run_tests(self, test_labels, extra_tests=None, **kwargs):
+        if test_labels:
+            # TODO: implement
+            raise NotImplementedError('test_labels is not implemented.')
+        if extra_tests is not None:
+            # TODO: implement
+            raise NotImplementedError('extra_tests is not implemented.')
+
+        # Translate arguments
+        argv = ['--ds', self.ds]
+        if self.verbosity == 0:
+            argv.append('--quiet')
+        if self.verbosity == 2:
+            argv.append('--verbose')
+        if self.failfast:
+            argv.append('--exitfirst')
+        if self.keepdb:
+            argv.append('--nomigrations')
+
+        # Run py.test
+        pytest.main(argv)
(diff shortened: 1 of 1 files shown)


blueyed:
Sweet!

> the upcoming Django 1.8 --keepdb => this project's --nomigrations equivalent

Wouldn't this be `--reuse-db` instead - or maybe both?

joeyespo:
@blueyed Thanks! And thanks for this project!

I'm not entirely sure. We were using [django-test-without-migrations](https://github.com/henriquebastos/django-test-without-migrations/) after doing [some research](http://stackoverflow.com/questions/25161425/disable-migrations-when-running-unit-tests-in-django-1-7) to speed up our local continuous test runners. After finding this project, I saw that `--nomigrations` did the same thing (since it points to the [same gist](https://gist.github.com/NotSqrt/5f3c76cd15e40ef62d09)). The upcoming `--keepdb` replaces that hack, so it seemed to be a good translation.

Re-reading the docs, it does seem like `--reuse-db` is a better fit. I just tested it and the running times are nearly equivalent. Is the difference just that `--reuse-db` sets up your database once, whereas `--nomigrations` does nothing? You probably know much more about the subtleties between them, and how 1.8's `--keepdb` plays into it. So if you think that's the right translation, I'll make the change :-)

blueyed:
> Is the difference just that --reuse-db sets up your database once, whereas --nomigrations does nothing?

I think so, yes.

pelme:
Thanks for the patch! Anything that makes it easier for new people to get started using pytest is a great thing! :)

Is there a way to make to command pass any arguments? It would be very bad if we're not able to pass all available arguments to pytest in a straightforward way.

If that cannot be solved, we should maybe think about putting a helpful text that is shown when pytest is invoked via manage.py that says something along the lines "This test suite uses pytest to run tests, to use all features of py.test, run the py.test command directly". I think we should make it clear that running `py.test` is the preferred way.

We also would need a test which uses `django_testdir` to invoke `manage.py test` and make sure it works.

tomviner:
@pelme @joeyespo yes you can pass any pytest arg after `--`, like this:

`manage.py test tests/test_utils.py --settings=settings.tdd_settings -v1 -- --tb=short`

this passes in a tuple of args like this in to `PyTestRunner.run_tests`, all in the `test_labels` argument:
`('tests/test_utils.py', '--tb=short')`

This is [the code that calls your runner from Django](https://github.com/django/django/blob/master/django/core/management/commands/test.py#L89-L90):

        test_runner = TestRunner(**options)
        failures = test_runner.run_tests(test_labels)

Note: the args Django knows about, like `--settings` and `-v`, it pulls out. The `--` is no longer visible in the tuple passed in, but the args after it are **not** validated against the test command option list.

So editing your patch @joeyespo a minimal test runner would be this: https://github.com/pytest-dev/pytest-django/compare/master...tomviner:test-runner-helper?expand=1

pytest_django/runner.py
```diff
@@ -0,0 +1,64 @@
...
+                                      'implemented.')
+        self.verbosity = verbosity
+        self.interactive = interactive
+        self.failfast = failfast
+        self.keepdb = keepdb
+
+        # TODO: Make this an option
+        self.ds = os.environ['DJANGO_SETTINGS_MODULE']
+
+    def run_tests(self, test_labels, extra_tests=None, **kwargs):
```
tomviner: This is [the code that calls your runner from Django](https://github.com/django/django/blob/master/django/core/management/commands/test.py#L89-L90):

        test_runner = TestRunner(**options)
        failures = test_runner.run_tests(test_labels)

So while it may be wise to leave `**kwargs` for compatibility with future Django changes, there is no extra_tests passed in.

ekohl:
Any plans to finish this PR? This would make it a lot easier for me to integrate in our existing setup.

tomviner:
Good timing @ekohl, maybe we can finish this at the Pytest Sprint starting on Monday @pelme?

ekohl:
For what it's worth, I've now included the pytest runner @tomviner linked and it does exactly what I want it to.

pelme:
@tomviner @ekohl Yes, let's have a look at getting this merged at the sprint. :)

tomviner:
@ekohl see https://github.com/pytest-dev/pytest-django/pull/347

ekohl:
Given #347 is merged, I think this can be closed now.

pelme:
See #347 for information on why this is not merged

joeyespo:
This isn't as nice as the ability to enter *any* Django project and immediately knowing how to run the tests (and so moves the burden of domain knowledge onto maintainers of Django projects, or worse, to their contributors). It's a great property of Rails projects that the community is reluctant to diverge from.

But I'm sure the discussions at the Pytest Sprint (which I was unfortunately unable to attend) balanced that with other concerns. :+1: to consensus.

PR #227, opened 2015-04-07
Update database.rst

bartsanchez:
I think that you have a mistake in the documentation.


blueyed:
Thanks for reporting this issue, but it's correct as it is: it describes `--create-db`, which re-creates the DB when used with `--reuse-db` (as documented currently).

Please re-open in case I've missed something.

Issue #235, opened 2015-05-08
Make "py.test --help" more friendly if DSM is set, but PYTHONPATH is not


blueyed:
pytest-django might cause `py.test --help` to behave very unfriendly, when
DJANGO_SETTINGS_MODULE is set, but PYTHONPATH is not set correctly.  See the traceback below.

It would be nice if it would behave better, especially when `py.test --help` is used.

    % py.test --help
    ---------------------------------------------------------------------------
    ImportError                               Traceback (most recent call last)
    …/django17/django/conf/__init__.py in __init__(self, settings_module)
         93         try:
    ---> 94             mod = importlib.import_module(self.SETTINGS_MODULE)
         95         except ImportError as e:

    …/pyenv/project/lib/python3.4/importlib/__init__.py in import_module(name, package)
        108             level += 1
    --> 109     return _bootstrap._gcd_import(name[level:], package, level)
        110 

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _gcd_import(name, package, level)

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _find_and_load(name, import_)

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _find_and_load_unlocked(name, import_)

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _call_with_frames_removed(f, *args, **kwds)

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _gcd_import(name, package, level)

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _find_and_load(name, import_)

    …/pyenv/project/lib/python3.4/importlib/_bootstrap.py in _find_and_load_unlocked(name, import_)

    ImportError: No module named 'config'

    During handling of the above exception, another exception occurred:

    ImportError                               Traceback (most recent call last)
    …/pytest-django/pytest_django/plugin.py in _handle_import_error(extra_message)
         91     try:
    ---> 92         yield
         93     except ImportError as e:

    …/pytest-django/pytest_django/plugin.py in pytest_load_initial_conftests(early_config, parser, args)
        197         with _handle_import_error(_django_project_scan_outcome):
    --> 198             settings.DATABASES
        199 

    …/django17/django/conf/__init__.py in __getattr__(self, name)
         45         if self._wrapped is empty:
    ---> 46             self._setup(name)
         47         return getattr(self._wrapped, name)

[...]

PR #238, opened 2015-05-09
Skip Django setup and manage.py scanning with "pytest --help"

blueyed:
With `pytest --version` and `pytest --help` pytest-django should not
throw an error (ImportError) in case of invalid DSM settings etc.

This removes hooking into `pytest_configure` altogether, and moves the
late call do `_setup_django` into the session-scoped autoload fixture instead.

Fixes https://github.com/pytest-dev/pytest-django/issues/235


blueyed:
@pelme 
Please review and merge it if it's appropriate.

PR #242, opened 2015-06-02
Clean up urls cache

### LINKED ISSUE #183: pytest.mark.urls does not properly clear the url cache
In the `restore` function in the `_django_set_urlconf`, `clear_url_caches()` must be called to actually clear the url after `ROOT_URLCONF` is restored

birdsarah:
This fixes #183, but I didn't write any tests for this change as just needed a quick fix and not immediately sure how to test.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -314,7 +314,7 @@ def _django_set_urlconf(request):
     if marker:
         skip_if_no_django()
         import django.conf
-        from django.core.urlresolvers import clear_url_caches
+        from django.core.urlresolvers import clear_url_caches, set_urlconf
 
         validate_urls(marker)
         original_urlconf = django.conf.settings.ROOT_URLCONF
@@ -323,6 +323,10 @@ def _django_set_urlconf(request):
 
         def restore():
             django.conf.settings.ROOT_URLCONF = original_urlconf
+            # Copy the pattern from
+            # https://github.com/django/django/blob/master/django/test/signals.py#L152
+            clear_url_caches()
+            set_urlconf(None)
 
         request.addfinalizer(restore)
 
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -3,7 +3,6 @@ from django.conf import settings
 
 from pytest_django_test.compat import force_text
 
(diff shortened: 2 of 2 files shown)


blueyed:
Thanks!
Please squash that into a single commit.

Also, a test should be added. It should be not too hard.
We could look into it when merging.

pelme:
@birdsarah Thanks for the patch! Would you like to submit a test to this so that we can merge it? I plan on putting out a new release in a couple of days, it would be great to have this issue fixed!

birdsarah:
@pelme. Thanks for the message. I don't know how to test, and am afraid I don't have time to get my head into it in the next couple of days as am at europython.

pelme:
@birdsarah No problems! If you want to give it a go there is a pytest sprint going on tomorrow in the europython sprints. I was there today but am going home early tomorrow but there are a lot of people knowledgeable about pytest around if you need help :)

birdsarah:
@pelme - we did it :D - big thanks to @davidszotten at Europython.

pelme:
Thank you very much! Too bad I missed you at the sprints, but see you next year! :)

The tests fail on Django 1.3 (the import for `from django.conf.urls import patterns, url` should also try importing from `from django.conf.urls.defaults import patterns, url`): https://travis-ci.org/pytest-dev/pytest-django/jobs/72688782#L225

There is also a linter violation: https://travis-ci.org/pytest-dev/pytest-django/jobs/72688796#L136-L139

I can fix those when I merge this, but it might be nice to do the last bits too. Let me know either way and I will try to merge this soon!

birdsarah:
Will push a fix today.

blueyed:
:+1:

k4nar:
Nice work, I was just about to submit a similar PR :) .

k4nar:
Actually, I wonder if we shouldn't call `set_urlconf(None)` also when setting the temporary `ROOT_URLCONF`. This is what is done in `TestCase` here: https://github.com/django/django/blob/master/django/test/testcases.py#L239

pelme:
Thanks everyone, I have merged this and added a call to `set_urlconf(None)` as suggested by @k4nar .

Issue #244, opened 2015-06-09
Tests fail on master locally due to an error un subprocess


hoh:
Hello,

I cannot manage to run the tests for `pytest-django` successfully on my system: I get an error related to python's `subprocess` library every time. I am probably not the only person affected as I could reproduce it in a brand new vanilla Linux container on Ubuntu 14.04 LTS. 

# Steps to reproduce :
* setup a new computer/LXC running Ubuntu 14.04 or 15.05
* `apt-get install build-essential git virtualenv python2.7 command-not-found python-minimal`
* `git clone git@github.com:pytest-dev/pytest-django.git`
* `cd pytest-django`
* `make test`

# Result
1 test fails, with the exception `OSError: [Errno 2] No such file or directory`

# Full error log

```
okso@pytestdjango:/shared/contrib/pytest-django-master$ make test
virtualenv .
Running virtualenv with interpreter /usr/bin/python2
New python executable in ./bin/python2
Also creating executable in ./bin/python
Installing setuptools, pip...done.
bin/pip install -Ur requirements.txt
Obtaining file:///shared/contrib/pytest-django-master (from -r requirements.txt (line 1))
  Running setup.py (path:/shared/contrib/pytest-django-master/setup.py) egg_info for package from file:///shared/contrib/pytest-django-master
    
Downloading/unpacking django (from -r requirements.txt (line 2))
  Downloading Django-1.8.2-py2.py3-none-any.whl (6.2MB): 6.2MB downloaded
Downloading/unpacking django-configurations (from -r requirements.txt (line 3))
  Downloading django_configurations-0.8-py2.py3-none-any.whl
Downloading/unpacking pytest-xdist (from -r requirements.txt (line 4))
  Downloading pytest-xdist-1.12.tar.gz
  Running setup.py (path:/shared/contrib/pytest-django-master/build/pytest-xdist/setup.py) egg_info for package pytest-xdist
    
    no previously-included directories found matching '.hg'
Downloading/unpacking tox (from -r requirements.txt (line 5))
  Downloading tox-2.0.2-py2.py3-none-any.whl
Downloading/unpacking wheel (from -r requirements.txt (line 6))
  Downloading wheel-0.24.0-py2.py3-none-any.whl (63kB): 63kB downloaded
Downloading/unpacking twine (from -r requirements.txt (line 7))
  Downloading twine-1.5.0-py2.py3-none-any.whl
Downloading/unpacking south (from -r requirements.txt (line 8))
[...]

PR #245, opened 2015-06-12
Use builtin sqlite3 library instead of 'sqlite3' CLI

hoh:
This removes the dependency on the 'sqlite3' third-party program, which is not present by default on most distributions (eg: Ubuntu 14.04) in favor of the sqlite3 library built into Python.

Fixes https://github.com/pytest-dev/pytest-django/issues/244


blueyed:
Thanks!

PR #246, opened 2015-06-12
Fix case when User model is based on email and has no username

hoh:
This fixes usecases whith a custom User model that uses an email address as the username, but has no username field present. An example of such model is present in the [Django doc on customizing authentication](https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#a-full-example).

The function `def admin_client(...)` requires access to `django_username_field` as `admin_user.username` is not valid if the username field is absent.

This patch is based on current master and #241.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -241,21 +241,30 @@ def admin_user(db, django_user_model, django_username_field):
     try:
         user = UserModel._default_manager.get(**{username_field: 'admin'})
     except UserModel.DoesNotExist:
-        extra_fields = {}
-        if username_field != 'username':
-            extra_fields[username_field] = 'admin'
+        usermodel_fields = UserModel._meta.get_all_field_names()
+
+        extra_fields = {django_username_field: 'admin'}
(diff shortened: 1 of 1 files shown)


hoh:
Concerning tests, I looked at copying [test_custom_user_model](https://github.com/pytest-dev/pytest-django/blob/master/tests/test_fixtures.py#L219), but it looked quite excessive to duplicate/fork all the code of that function for this usecase.

pelme:
Thanks for the contribution. The patch looks good but I think we need a test for this.

We could introduce another fixture `django_testdir_custom_user` or something like that. We could that use that in all tests for custom user models without a lot of duplication.

It should be possible to pretty much move the `django_testdir.create_app_file(...)` calls in `test_custom_user_model` to that new fixture.

pelme:
Here is another PR which tackles the same problem: #241

aaugustin:
I hit the same problem and I think this pull request is correct.

blueyed:
Anybody up for writing a test to get this merged?

PR #251, opened 2015-07-27
Ensure the overridden URL conf is loaded

benemery:
If we override the `ROOT_URLCONF` without activating the new config, then performing an action which doesn't force django to load the new url conf will fail.

I've taken the liberty of cloning the quarantined url conf from https://github.com/pytest-dev/pytest-django/pull/242, perhaps once (/ if) they're both accepted this can be improved.


pelme:
This has been fixed in master and will be part of the next release, thanks! :)

PR #254, opened 2015-08-11
Enabled cascading truncation of tables when flushing them

thedrow:
This is not documented but it works and it is required many times when flushes are not successful.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -95,7 +95,8 @@ def _django_db_fixture_helper(transactional, request, _django_cursor_wrapper):
 
                 for db in connections:
                     call_command('flush', interactive=False, database=db,
-                                 verbosity=pytest.config.option.verbose)
+                                 verbosity=pytest.config.option.verbose,
+                                 allow_cascade=True)
                 for conn in connections.all():
                     conn.close()
             request.addfinalizer(flushdb)
--- a/pytest_django_test/app/models.py
+++ b/pytest_django_test/app/models.py
@@ -9,3 +9,6 @@ class Item(models.Model):
 
     def __str__(self):
         return self.name
+
+class ItemMetadata(models.Model):
+    owner = models.ForeignKey(Item)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -159,19 +159,32 @@ def django_testdir_initial(django_testdir):
                 def forwards(self, orm):
                     db.create_table(u'app_item', (
                         (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
-                        ('name', self.gf('django.db.models.fields.CharField')(max_length=100)),
+                        (u'name', self.gf('django.db.models.fields.CharField')(max_length=100)),
                     ))
                     db.send_create_signal(u'app', ['Item'])
+                    print("mark_south_migration_forwards")
+                    db.create_table(u'app_itemmetadata', (
+                        (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
+                        (u'owner', self.gf('django.db.models.fields.related.ForeignKey')(orm['app.Item'])),
+                    ))
+                    db.send_create_signal(u'app', ['ItemMetadata'])
                     print("mark_south_migration_forwards"),
+
 
                 def backwards(self, orm):
                     db.delete_table(u'app_item')
+                    db.delete_table(u'app_itemmetadata')
 
                 models = {
                     u'app.item': {
                         'Meta': {'object_name': 'Item'},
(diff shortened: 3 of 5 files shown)


pelme:
What does this do more specifically? In what situations is this necessary? Could we have a test for this somehow?

thedrow:
This is necessary to truncate tables that have foreign keys to other tables.
Otherwise you'll end up failing the entire test suite because you can't flush all the tables.
What kind of test do you need?

pelme:
I see, I was not aware of that! We would need a test that ensures that the flush is done by cascading (i.e. having a table with a foreign key to another table) and then ensures that everything works.

I'm thinking about something similar to https://github.com/pytest-dev/pytest-django/blob/b9eb210e592d15ef0e7f3fc50cb89c9c72fb44a4/tests/test_database.py#L167-L204 which creates a test file that is run to ensure that everything works. The models.py that is used from `django_testdir` (https://github.com/pytest-dev/pytest-django/blob/b9eb210e592d15ef0e7f3fc50cb89c9c72fb44a4/pytest_django_test/app/models.py) would need to be extended with another table with a FK to make it possible.

thedrow:
I'm unable to reproduce this with a test but able to reproduce it with my company's project.
Not sure what to do here.

thedrow:
In any case, nothing is broken by this change and the test I added at least proves that.
I hope you'll merge this anyway.

thedrow:
Something blew up tox and it wasn't me.

thedrow:
It's the same on master. I have no idea what's wrong.

pelme:
The tox issues is fixed in master if you want to update the PR.

I'd like the test to properly fail without this patch. There are a lot of edge cases with Django database setup, without tests that cover this we won't be able to maintain this project properly.

thedrow:
I can reproduce this bug with my own code which I cannot share.
I'll update the PR.
My tests prove that the change doesn't break current behaviour. Can it be good enough for now until I find a way to create a test case that reproduces the problem?
I'm really stuck because pytest-django doesn't flush cascadingly.

pelme:
Consistently being able to trigger the bug in your own test suite is a great start. You should start there and try to narrow it down. If you cannot make a test case for pytest-django, providing a minimal example or instructions that triggers the bug would help someone else to write the test. The clearer and better instructions you can give, the more likely it is that someone else will help out if you cannot fix it yourself.

Being realistic, tests written after the fix has been merged will never happen. I will not merge this before there is a test case that clearly triggers this bug.

*I* will be *really really* stuck maintaining pytest-django without a proper test suite. :-)

pytest_django/fixtures.py
```diff
@@ -92,7 +92,8 @@ def flushdb():
 
                 for db in connections:
                     call_command('flush', interactive=False, database=db,
-                                 verbosity=pytest.config.option.verbose)
+                                 verbosity=pytest.config.option.verbose,
+                                 allow_cascade=True)
```
blueyed: This will only handle the case for Django before 1.5..

blueyed:
@thedrow 
> I can reproduce this bug with my own code which I cannot share.

Are there migrations involved in your test?

I've found that you would have to migrate back to the end when testing them, e.g. 

    # Migrate to the end again.
    executor.loader.build_graph()  # reload.
    executor.migrate(executor.loader.graph.leaf_nodes(app))

blueyed:
@thedrow 
Can you rebase it on master for now, please?

thedrow:
Yes, migrations are involved in my tests.

pytest_django/fixtures.py
thedrow: Why, it's still available in Django versions above 1.5.

thedrow:
@blueyed Rebased.

maxnordlund:
I guess this is a goner, and I hit this today. This thread shows the issue, though it says it's fixed I still get this on 1.11.9:

https://code.djangoproject.com/ticket/6820

Basically, if you have a foreign key, in my case because of inheritance to django-invitation for a custom model, it explodes when it tries to truncate the tables.

To reproduce I think you need to modules, one which is considered third-party, and one in the app itself. Then setup single table inheritance and run the truncate command. But I'm not sure how to do that. This fix might also not work anymore, but from what I can see in the [source](https://github.com/django/django/blob/1.11.9/django/core/management/sql.py#L7), there's both an option for cascade and one for truncating everything.

How can this be fixed is the next question? Do you still want to get this merged @thedrow or should someone else (me) take another shot at it?

thedrow:
I no longer use Django so feel free to take this PR from here.

blueyed:
Closing in favor of #575.

Issue #243, opened 2015-06-08
Support database access for "session" level fixtures.


dan-passaro:
Our test suite uses a custom TestRunner which has a `run_suite` method like so:

    def run_suite(self, suite, **kwargs):
        call_command('loaddata', 'foo')
        call_command('loaddata', 'bar')
        return super(CustomTestRunner, self).run_suite(suite, **kwargs)

The fixtures it loads are indeed so common it makes sense just to put load them for every test case. For example we host multiple `Site`s off of our app and one of the fixtures loads a subset of those `Site` objects. Foreign keys to `Site` exist in almost all our apps so it makes sense to just load a few before the whole suite.

Although in theory I should be able to rename the fixtures to `initial_data.json` and have them load, that is for whatever strange reason not working. That is when I found there is no way to duct-tape over it in py.test, because the only time `db` access is allowed is at the function level.

Are there any plans to include a session-level fixture for db access?

blueyed:
Which Django version are you using?
IIRC `initial_data.json` is not being considered with apps that use migration in Django 1.7+.

There seems to be a workaround in a (possibly duplicate issue): https://github.com/pytest-dev/pytest-django/issues/105#issuecomment-61474664

You could also use a signal via `app/management/__init__.py` to work around this:

    # Handle initial data in Django 1.7+.

    from django.db.models import signals
    from django.core.management import call_command

    def load_initial_data(app_config, sender, **kwargs):
        my_label = __package__.split(".")[-2]
        if sender.label == my_label:
            call_command('loaddata', 'initial_data.yaml', app_label=my_label)

    signals.post_migrate.connect(load_initial_data, weak=False)

See also https://github.com/pytest-dev/pytest-django/pull/220#issuecomment-83310828

dan-passaro:
Aha! I had actually tried using `_django_db_setup` but got errors about not being allowed to access the Django DB without a marker. I suppose the `_django_cursor_wrapper` suppresses that issue as shown in #105.

It looks like this will allow me to use the DB in a session-level fixture, and will solve my problem. Is there any interest in making that accessible in a more user friendly way? Is this workaround unsupported and prone to break in the future?

pelme:
This is something which is worked on and that will be supported by a public API. I tagged this issue with "db-configuration" which is issues related to having more options when it comes to configuring the database. I will update this issue with more information once those changes gets into master!

HeyHugo:
[...]

Issue #105, opened 2014-06-11
Do common db setup between tests once


alvinchow86:
It would be useful if I could do some common database setup that's shared between tests (like loading some large fixture data or something), just once for the test run rather than for every test, for faster performance. The test would still be wrapped in a transaction, so any db changes in tests would be rolled back to the original state. 

However this doesn't seem possible in pytest_django, since the `db` fixture is only function scoped, whereas I might want to do this database setup on a module or session scope.

rbiersbach:
We also have a problem with this as we are using parameterize a lot for read-only access. 
We have a workaround but that just feels super awkward compared to the scopes which are intended for this purpose originally.

bluetech:
We are having some overlapping discussion on this in #514, I'll close this one in favor of that.

PR #258, opened 2015-09-04
Add `shared_db_wrapper` for creating long-lived db state

ktosiek:
This is my take on #105 and #243. This pull request adds a new fixture, `shared_db_wrapper`, that lets other fixtures access the database regardless of scope.

`shared_db_wrapper` is a contextmanager that creates a new savepoint/transaction, temporarily enables the cursor wrapper, and adds a finalizer through user fixture's request. This design, as opposed to making it a normal fixture, has 2 reasons:
- finalization needs to be tied to user fixture's scope, not `shared_db_wrapper`'s
- I don't want just having some shared state loaded to leave the cursor enabled

Is this approach sensible? If so, I'll be happy to write the documentation and make sure this PR is mergeable.


blueyed:
Looks good to me so far!
Could you please squash it already?

Would this help for keeping data from migrations with `transactional_db` (https://github.com/pytest-dev/pytest-django/pull/220#issuecomment-83310828)?
I was looking into https://github.com/fastmonkeys/stellar in this regard - which might be used if it becomes easily usable as a library (https://github.com/fastmonkeys/stellar/issues/53).
It basically allows to create a snapshot and rollback to it, which could be used in `transactional_db` (or a new variant).

@pelme 
What do you think?

ktosiek:
Sure, I'll squash it shortly.

This pretty much bails out in trasaction_db, so won't help there. But I think the API is restrictive enough on the user side that we can add some kind of snapshots later.

pelme:
This is something I personally have been thinking about and wanting for a long time. The API and implementation looks good and simple to use! Passing request explicitly to tie the scope to the user fixture is the sensible (and possibly only way currently to handle this). This will be a killer feature for pytest-django! :)

Is session/module/class fixtures that is used at the same time handled properly? I see no reason why they shouldn't be, but it would be nice with a test that uses all scopes at the same time.

Transactional db support with stellar or flushes would be a really cool future addition to shared_db_wrapper.

ktosiek:
I can see how I could safely test mixing class and function scopes, but how would I write a test for using it with session and module scopes?
Would a test that just calls the wrapper multiple times and checks the DB state around those calls good enough?

pelme:
@ktosiek It is possible to test that with a "pytester" test, that invokes a new pytest session. See for instance https://github.com/pytest-dev/pytest-django/blob/f1711f01c894682b69fed1924676916ecea53fb1/tests/test_unittest.py#L122-L153.

The trick is to use the `django_testdir` fixture which will create a new Django project for that particular test and then invoke `django_testdir.runpytest()` to run and collect tests.

ktosiek:
I'll push a pytester test shortly, but I also have some bad news: it looks like this won't work on Django 1.6 and 1.7, as they always close the DB connection in TransactionTestCase._post_teardown: https://github.com/django/django/blob/stable/1.7.x/django/test/testcases.py#L828

1.8 should be OK, as it only does that if some db backends don't support transactions.

pelme:
Hmm, couldn't we monkey patch _post_teardown in those cases and have it not close the connection? It would for sure be nice to support Django 1.6 and 1.7 if possible!

aaugustin:
Django 1.6 and 1.7 don't receive security updates anymore. I don't think you should refrain from merging this feature just because it doesn't work on these versions. I suggest to skip the tests on Django < 1.8.

Is there anything I can do to help completing this feature?

pelme:
@aaugustin  Agreed, it wouldn't be too bad to not support Django <1.7 with shared_db_wrapper, so let's avoid the complexities that would add.

With your expertise in Django transaction management, if you could have a quick look at the PR that would be very helpful. Also if you could try it out in a project where you need it that would be very helpful too.

There is nothing that blocks this from being merged. If I don't hear anything else I will merge it in a couple of days and push a new release. :)

aaugustin:
I'll try it in my current project and report back (hopefully today or tomorrow).

aaugustin:
I tried the patch and unfortunately I couldn't make it work.

I'm getting two kinds of errors:

- some tests fail, apparently because of isolation issues between tests -- which indicates the database state isn't reset between tests like I expect
- if the module where I declare a module-level `shared_db_wrapper` fixture also contains tests that don't touch the database and aren't marked with `@pytest.mark.django_db`, these tests fail with "Database access not allowed, use the "django_db" mark to enable it."

pytest_django/fixtures.py
```diff
@@ -192,6 +195,63 @@ def transactional_db(request, _django_db_setup, _django_cursor_wrapper):
...
+        if 'transactional_db' in request.funcargnames:
+            raise Exception(
+                'shared_db_wrapper cannot be used with `transactional_db`.')
+
+        with _django_cursor_wrapper:
+            if not connection.features.supports_transactions:
+                raise Exception(
+                    "shared_db_wrapper cannot be used when "
+                    "the database doesn't support transactions.")
+
```
aaugustin: It seems to me that the next 25 lines could we written as:

        try:
            _django_cursor_wrapper.enable()
            with transaction.atomic():
                yield
                transaction.set_rollback(True)
        finally:
            _django_cursor_wrapper.restore()

This raises the question of why this fixture needs to care about `_django_cursor_wrapper`; it's unclear to me why this is necessary.

docs/helpers.rst
```diff
@@ -191,6 +191,26 @@ transaction support.  This is only required for fixtures which need
...
+
+At the moment it does not work with ``transactional_db``,
+as the fixture itself depends on transactions.
+It also needs Django >= 1.6, as it's based on the ``atomic`` API.
+
+Example usage::
+
+  @pytest.fixture(scope='module')
+  def some_users(request, shared_db_wrapper):
+      with shared_db_wrapper(request):
```
aaugustin: This API adds some boilerplate. I'd love to write just:

```python
@pytest.fixture(scope='module')
def some_users(shared_db_wrapper):
    User.objects.create(username='foo')
```

aaugustin:
I'm not comfortable with pytest's internal APIs, but the general scheme here should be:

```python
    with transaction.atomic():
        # run shared_db_wrapper fixtures
        for test in in tests:
            with transaction.atomic():
                # run test
                transaction.set_rollback(True)
        transaction.set_rollback(True)
```

I hope this helps...

docs/helpers.rst
davidhalter: I would like that too, but there is an issue with this. When you specify a dependency to `some_users`, the `shared_db_wrapper` fixture will still be active. This is a problem, because you only want to use global state for a limited amount of data (e.g. creating the users and not modifying them).

I think that is why OP has called it a `shared_db_wrapper` and not `shared_db`. 

However, maybe there's a solution around this. Maybe it's possible to somehow create a special "destructor" before entering "local" scope. However this might be complicated given how the current architecture (I have no idea about pytest-django internals).

davidhalter:
If there's issues with this, I would be glad to help out. Let me know what's still missing.

docs/helpers.rst
davidhalter: But to at least simplify it a bit: Why do we need access to request?

ktosiek: We need `request` to register a finalizer that will run when the shared fixture (and not just the wrapper) is getting torn down.

Would it be desirable to put all that boilerplate in a decorator that would be used instead of (or maybe before?) pytest.fixture? Something like:
```
@shared_db_fixture(scope='module')
def some_users():
    User.objects.create(username='foo')

Comment (aaugustin) [inline docs/helpers.rst]: I have to admit that the interactions between fixtures of different scopes are totally over my head :-/

The end goal is to call `transaction.atomic` and `transaction.set_rollback` as follows -- but I don't know how to map this to pytest's hooks and to make it work within pytest-django's existing infrastructure.

```
with transaction.atomic():
    fixture_1()     # used in the entire test suite

    with transaction.atomic():
        fixture_2()     # used in one module

        with transaction.atomic():
            test_21()
            transaction.set_rollback(True)

        with transaction.atomic():
            test_22()
            transaction.set_rollback(True)

        with transaction.atomic():
            test_23()
            transaction.set_rollback(True)

        transaction.set_rollback(True)  # fixture 2

    with transaction.atomic():
        fixture_3()     # used in two modules

        with transaction.atomic():
            fixture_4()     # used in one module which also uses 3

            with transaction.atomic():
                test_41()
                transaction.set_rollback(True)

            with transaction.atomic():
                test_42()
                transaction.set_rollback(True)

            with transaction.atomic():
                test_43()
                transaction.set_rollback(True)

            transaction.set_rollback(True)  # fixture 4

        with transaction.atomic():
            fixture_5()     # used in one module which also uses 3

            with transaction.atomic():
                test_51()
                transaction.set_rollback(True)

            with transaction.atomic():
                test_52()
                transaction.set_rollback(True)

            with transaction.atomic():
                test_53()
                transaction.set_rollback(True)

                transaction.set_rollback(True)

            transaction.set_rollback(True)  # fixture 5

        transaction.set_rollback(True)  # fixture 3

    transaction.set_rollback(True)  # fixture 1

```

If multiple databases are being used, you need to call `transaction.atomic(using=...)` and `transaction.set_rollback(True, using=...)` for each of them.

Ping me if you have questions about Django's transaction management — I wrote it.

File: pytest_django/fixtures.py (hunk 1 above)
Comment (davidhalter) [inline pytest_django/fixtures.py]: @ktosiek What do you think about that?

**Comment by ktosiek** [comment]
This still has problems: there's no way that I see to force py.test to only run `db` after shared state fixtures.
This can cause the `db` atomic to be started first (it should be last, or at least after all shared fixtures), and then the whole test will get run in the last shared fixture's atomic (and will rollback the last fixture's DB changes).

It looks like this behavior depends on order of fixtures in first test function, maybe we can fix it by hooking into test collection?

**Comment by davidhalter** [comment]
> This still has problems: there's no way that I see to force py.test to only run db after shared state fixtures.

Would it help to use the decorator approach you proposed above? I think this would make sense even just for usability. I don't like how complicated it is, currently.

**Comment by ktosiek** [comment]
Not really, the decorator would just hide the `with shared_db_wrapper(...):` part, and won't change how it all works. But maybe a `pytest_runtest_setup` hook could be used to order test's fixtures by scope.

File: docs/helpers.rst (hunk 2 above)
Comment (ktosiek) [inline docs/helpers.rst]: I can't use `with atomic():` because I have to register a callback on request :-/.

Thank you for pointing out transaction.set_rollback(True) - this will make my code cleaner (no more dummy exceptions!).

For the `using=...` part I might even be able to use Django's test case enter/exit atomics, or at least get some inspiration from there.

**Comment by davidhalter** [comment]
But with a decorator you could control the fixture order, right? You could just place it in the first spot. (Note that I don't have a lot of knowledge about py.test fixtures)

**Comment by ktosiek** [comment]
Nope, what I need is to affect relative ordering between fixtures on the actual test. It's not about fixture's dependencies. Using a decorator won't help there.

I kind of have an idea how to do that (which is a bit hacky - using pytest_runtest_setup and pre-loading some fixtures with item._request.getargvalue), but I'll have to try it (hopefully later today).

**Comment by davidhalter** [comment]
Hmm, I see! There's probably solutions for that as well, but they are quite hacky. Wouldn't this work?:

```
import inspect                                                                  
from textwrap import dedent                                                     
from functools import wraps                                                     
                                                                                
                                                                                
def shared_db_fixture(scope='module'):                                          
    def func_wrapper(func):                                                     
        arg_list = list(inspect.getfullargspec(func))                           
        old_args = inspect.formatargspec(*arg_list)                             
        arg_list[0].insert(0, 'request')                                        
        arg_list[0].insert(0, '_shared_db_wrapper')                             
        new_args = inspect.formatargspec(*arg_list)                             
                                                                                
        code = dedent("""                                                       
        @wraps(func)                                                            
        def wrapper%s:                                                          
            with _shared_db_wrapper(request):                                   
                return func%s                                                   
        """ % (new_args, old_args))                                             
                                                                                
        exec(code, globals(), locals())                                         
        print(code, locals())                                                   
        return locals()['wrapper']  # Don't ask me why it doesn't work directly.
                                                                                
    return func_wrapper                                                         
                                                                                
                                                                                
@shared_db_fixture()                                                            
def test_hello(my_fixture):                                                     
    assert myfixture.foobar                                                     
```

Clearly this is hacky, but it's actually very clear what it does. Hacking around with pytest_runtest_setup is probably not cleaner (and if so, please show me :)). Also, pytest itself has very hacky origins, since it uses function names as an actual feature (which is strange, when you first start with pytest, but really cool once you get it). This leads to the fact that there's no way of modiying the function input anymore, like args/kwargs in a decorator, except for `exec`.

Tell me what you think!

**Comment by ktosiek** [comment]
as a nicer looking API? Probably. But it won't help with the fixture instantiation order (you'd need to wrap each test to move db-based fixtures after shared db fixtures)

**Comment by davidhalter** [comment]
True. I've just realized how complicated this actually is.

**Comment by davidhalter** [comment]
My intermediate fix BTW is to use this:

```
@pytest.fixture(scope="session", autouse=True)
def setup_django_db_hack(_django_db_setup, request, _django_cursor_wrapper):
    _django_db_fixture_helper(False, request, _django_cursor_wrapper)
```

This initializes the database in the beginning and with that all fixtures can write to it. What do you guys think about this?

**Comment by larsrinn** [comment]
Even though this is quite an old PR, I would be very much interested in the functionality it provides (without reading through the whole thread though). Why hasn't it been merged? Just because of the merge conflicts?

**Comment by davidhalter** [comment]
IMO it's not just a simple problem and this PR doesn't solve everything, really. There are issues in the underlying pytest infrastructure/architecture that make it really hard to do this right.

My personal approach has been altered a bit, I use this for a kind of large Django project:

```
@pytest.fixture(scope="session", autouse=True)
def actually_setup_db(django_db_setup):
    """
    This is an autouse fixture to guarantee calling django_db_setup. If this is
    not done, Django allows accessing the sqlite.db file and modifying it. This
    is obviously NOT what we want.
    """


def session_db(actually_setup_db, django_db_blocker):
    """
    Unlocks the database session-wide. The default of django is to block the
    database.
    """
        yield
```

It's not perfect. There are some major issues that you might run into. Don't ask me why I didn't make this an autouse fixture... It's just bug-prone and you have to play with it. The fixtures also might need to be defined in the right places, etc. It's just not what you really want. They need to improve some pytest things first, IMO.

I also think that this issue might be a blocker and is highly relevant: https://github.com/pytest-dev/pytest/issues/2405

ktosiek:
IIRC this PR was blocked by problems with forcing fixture ordering (see https://github.com/pytest-dev/pytest-django/pull/258#issuecomment-203130289), I haven't checked if there are any new options for working around them.

yozlet:
@ktosiek Looks like fixtures are now loaded in scope order, thanks to https://github.com/pytest-dev/pytest/pull/3306 (which went into pytest 3.5.0). Does this help things?

ktosiek:
@yozlet thank you for pointing this out, it might help. But I won't have time to work on this PR this month.

pdonorio:
The feature of keeping db data across multiple tests would be very important in some scenarios for our tests, at the very least for performance reasons.

May I help you in any way @ktosiek to get this released?

bluetech:
FYI, we are having some discussion on alternatives to this PR in #514.

PR #261, opened 2015-09-14
Use Django's `--keepdb` with Django 1.8+

blueyed:
Django 1.8 supports `--keepdb` [1], and it makes sense to make use of
it.  This should probably handle migrations in the context of
`transactional_db` better, i.e. by re-applying migrations with a flushed
DB.

1: https://docs.djangoproject.com/en/1.8/ref/django-admin/#django-admin-option---keepdb

### DIFF
--- a/pytest_django/db_reuse.py
+++ b/pytest_django/db_reuse.py
@@ -95,6 +95,8 @@ def create_test_db_with_reuse(self, verbosity=1, autoclobber=False,
     This method is a monkey patched version of create_test_db that
     will not actually create a new database, but just reuse the
     existing.
+
+    This is only used with Django < 1.8.
     """
     test_database_name = self._get_test_db_name()
     self.connection.settings_dict['NAME'] = test_database_name
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -43,15 +43,18 @@ def _django_db_setup(request,
     if request.config.getvalue('nomigrations'):
         _disable_native_migrations()
 
+    db_args = {}
     with _django_cursor_wrapper:
-        # Monkey patch Django's setup code to support database re-use
-        if request.config.getvalue('reuse_db'):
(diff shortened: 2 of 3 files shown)


blueyed:
Please note that this will change the errors you will see when the test database cannot be created on PostgreSQL, because Django will ignore the error from `CREATE DATABASE` with `--keepdb`.

Reported for Django at https://code.djangoproject.com/ticket/25406.

pelme:
This is great! Is this fully backward compatible with our old monkey patching? (I guess it is since all tests passes)

I think we should deprecate `--reuse-db` and just stick with `--keepdb` and follow the Django conventions. In some distant future we can just drop our own monkey patches altogether and use Django's builtin support. :)

blueyed:
> I think we should deprecate `--reuse-db` and just stick with `--keepdb`

:+1: 

Will you do it?

pelme:
I'd like to get a new release out and will focus on that for now, but I can give it a shot after that if you or no-one else gets there before me :)

pelme:
(I have no problem merging this before btw, we can do the deprecation in a separate PR)

blueyed:
> Is this fully backward compatible with our old monkey patching?

The errors are different, because it triggers a bug in Django (https://github.com/pytest-dev/pytest-django/pull/261#issuecomment-140378359).

Maybe it is better to not mess with it for the next release then?

agamdua:
This would be a really cool thing to have - I would like to drop pytest in a project I'm on now, but its dependent on the `--keepdb` flag.

I guess this is conditional on a solution to https://code.djangoproject.com/ticket/25406.. Can I help in any way?

pelme:
I have merged this now but not deprecated `--reuse-db` yet.

@agamdua You can already get the same functionality with `--reuse-db` (and it has been possible for quite some time). This PR is about using the Django implementation for database reuse rather than our own. :)

Issue #264, opened 2015-09-16
Create models with managed = False


ojake:
I have a project that works with 3rd party database, all the models for that database have managed = False in Meta because I don't want them to be created in my db.

I would still like to run some tests with fake data. So my question is: is there any way to create unmanaged models only for testing?

ojake:
I tried to find something that would let me test unmanaged models, and only found solutions for default django testing, nothing that would work with pytest-django.
So I hacked a marker with signature pytest.mark.django_use_model(model) that creates model (or models) for the scope of marker ignoring the managed = False.
@pelme do you think there is a chance that such gizmo belongs to pytest-django? I already have tests for this, but if it doesn't belong here I won't bother with updating documentation but will create it as standalone plugin instead.
If it can be merged to pytest-django, I think it would make sense to make it that test units marked with django_use_model would not require django_db, I'm not sure how to accomplish that.

It's here: https://github.com/ojake/pytest-django/tree/unmanaged_model

Cheers.

blueyed:
I think it would be useful.

But it makes sense to require the DB marker, since it will hit/use the database.
Or are you referring to skipping the setup part of it?

For reference: https://docs.djangoproject.com/en/1.8/ref/models/options/#django.db.models.Options.managed.

Anyway, it would be possible to automatically include / require another fixture/marker with yours.

ojake:
Hey.
Currently it has to be:
```python
@pytest.mark.django_db
@pytest.mark.django_use_model(MyUnamanged)
def test_myunmanaged_model():
    ...
```
My thinking was that it's kind of imperative that if test uses model, it will most likely hit db, so django_db could be omitted, but maybe it's better to keep it explicit.

blueyed:
While I like the explicit approach (which could be used as a marker), it can be included (see `_live_server_helper`, https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/fixtures.py#L350-L351):

    request.getfuncargvalue('db')

But then it might request/use `db`, when `transactional_db` is requested explicitly - not sure?!
A test for this would be good.

gorlins:
old thread but figured I'd post my workaround, which is simple and not pytest specific.

I have different settings modules for staging, prod, unittests etc but same result could be achieved using env variables or command line opts

In some models.py:

    MANAGED = settings.AM_I_TESTING  # Any logic/hard code that designates we're testing
    
    class SomeModel(Model):
        class Meta:
            managed = MANAGED

[...]

PR #270, opened 2015-09-24
Add support for testing unmanaged models

ojake:
As per https://github.com/pytest-dev/pytest-django/issues/264.
This is marker pytest.mark.django_use_model that allows testing of models that have `Meta.managed = False`

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -94,6 +94,32 @@ on what marks are and for notes on using_ them.
          client('some-url-with-invalid-template-vars')
 
 
+``pytest.mark.django_use_model`` - force model creation for unmanaged models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


ojake:
Ok, there is something weird going on. Looks like tests marked with ``django_use_model`` are not even executed. I discovered it when trying to use it with coverage.
I'm going to try to figure this out, but I would appreciate some help, because I clearly don't understand how pytest-django works.

Here's an example of "passing" test
https://github.com/ojake/pytest-django/blob/bug-no-test/tests/test_database.py#L190

blueyed:
Let's keep the PR open..
I currently don't have time to look into it, but we'll come to it.

ojake:
Thanks!

timb07:
Hi. I don't understand py.test well enough to work out why
`@pytest.mark.django_use_model(Unmanaged)`
causes the test not to be collected, but I have found a way that does work.

Mark with:
`@pytest.mark.django_use_model(model=Unmanaged)`

And then in _django_use_model(), fetch the model(s) like this:
`model = marker.kwargs['model']`
instead of using request.getfuncargvalue().

blueyed:
Closing this in favor of https://github.com/pytest-dev/pytest-django/pull/310.

PR #275, opened 2015-09-27
Django 1.9 support

pelme:
Work-in-progress support for Django 1.9

### DIFF
--- a/generate_configurations.py
+++ b/generate_configurations.py
@@ -28,7 +28,7 @@ PYTHON_MAIN_VERSIONS = ['python2.7', 'python3.4']
 PYTHON_VERSIONS = ['python2.6', 'python2.7', 'python3.2', 'python3.3',
                    'python3.4', 'python3.5', 'pypy', 'pypy3']
 PYTEST_VERSIONS = ['2.7.3', '2.8.0']
-DJANGO_VERSIONS = ['1.4', '1.5', '1.6', '1.7', '1.8', '1.9', 'master']
+DJANGO_VERSIONS = ['1.4', '1.5', '1.6', '1.7', '1.8', '1.9',]
 SETTINGS = ['sqlite', 'sqlite_file', 'mysql_myisam', 'mysql_innodb',
             'postgres']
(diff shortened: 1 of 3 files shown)


generate_configurations.py
```diff
@@ -28,7 +28,7 @@ def is_pypy(self):
 PYTHON_VERSIONS = ['python2.6', 'python2.7', 'python3.2', 'python3.3',
                    'python3.4', 'python3.5', 'pypy', 'pypy3']
 PYTEST_VERSIONS = ['2.7.3', '2.8.0']
-DJANGO_VERSIONS = ['1.4', '1.5', '1.6', '1.7', '1.8', '1.9', 'master']
+DJANGO_VERSIONS = ['1.4', '1.5', '1.6', '1.7', '1.8', '1.9',]
```
blueyed: @pelme 
It might work to move `master` to the beginning of the list to still have it in there.
IIRC the list is used from the end when generating variants.

pelme: Good point! This was just a throw-away branch to have all tests run on 1.9, but I will see if that works. It's bad not to have hard failures for anything else than sqlite :/

blueyed: > It's bad not to have hard failures for anything else than sqlite

You mean that everything != sqlite should fail?  Even Django master?

pelme: Currently (https://travis-ci.org/pytest-dev/pytest-django/builds/82408028) the only tests that runs against Postgres/MySQL is allowed failures (that also runs against master).

We need to have Postgres/MySQL running regularly with 1.4-1.9 and to have those build fail the entire build if there are failures.

blueyed: I see.  We could use only sqlite with Django master then maybe?

blueyed: We could only use `sqlite_file` for Django master in the Travis config probably.

pelme: Yes that would probably be good enough. I'm trying to fix up the Django master support now, a couple of things moved seems to have been deprecated.

PR #276, opened 2015-09-29
Use pytest 2.8.1

blueyed:


blueyed:
`test_ds_in_pytest_configure` seems to fail with 2.8.1: `testdir.runpytest_subprocess()` returns 5 instead of 0.

The skipif refers to https://github.com/pytest-dev/pytest/issues/1073, but that should be fixed.

pelme:
That is because pytest's new status codes, that tests has no test cases and the status is "yellow" instead of "green". We should just add a dummy test there.

blueyed:
Went into master already (https://github.com/pytest-dev/pytest-django/commit/6bccc6cb6bc5e7bb012aca62de7a46f0de14e0e7 etc).

pelme:
Sorry, did not see this quick enough. But good that it is solved, hope to get a new release out soonish. :)

PR #277, opened 2015-10-02
Fix for multiple inheritance of Django test cases

### LINKED ISSUE #265: TestCase inheritance breaks with setUpTestData (Django 1.8)
Hi,

We have a large test base with test cases which inherit from each other. I am in the process of converting them to Django 1.8 and tried to use the new `setUpTestData` class method for setting up test data.

This fails in a very strange way for tests using inheritance: `setUpTestData` is only invoked for the base class and not for the derived class. I think this may be because of [this commit](https://github.com/pytest-dev/pytest-django/commit/8a9aacbd835a49064ff8999137fa6ede34f3c520), which changes and overrides `setUpClass` and `tearDownClass` in a way which I honestly don't quite understand...

I have created a [failing test](https://github.com/pytest-dev/pytest-django/commit/107ba5b51aabc087d7c69078043c8c3816278ae1) which shows the problem; the lines which cause errors are annotated with `TODO` comments.

Any ideas on how to fix or circumvent this problem are greatly appreciated! :-)

pelme:
This PR hopefully fixes #265

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,7 +8,13 @@ Bug fixes
 ^^^^^^^^^
 * Ensure urlconf is properly reset when using @pytest.mark.urls. Thanks to
   Sarah Bird, David Szotten, Daniel Hahler and Yannick PÉROUX for patch and
-  discussions.
+  discussions. Fixes `issue #183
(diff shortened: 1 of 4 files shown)


pelme:
Closed in 96f5fb1817285fbf9ef9b0de89c5190d134f845a.

PR #278, opened 2015-10-04
Keep the existing PYTHONPATH in tox

ktosiek:
Without this the tests could not be run on systems that already use PYTHONPATH (like NixOS).

### DIFF
--- a/generate_configurations.py
+++ b/generate_configurations.py
@@ -49,7 +49,7 @@ TOX_TESTENV_TEMPLATE = dedent("""
     deps =
     %(deps)s
     setenv =
-         PYTHONPATH = {toxinidir}
+         PYTHONPATH = {toxinidir}:{env:PYTHONPATH:}
          UID = %(uid)s
     """)
 
--- a/tox.ini
+++ b/tox.ini
@@ -97,7 +97,7 @@ deps =
     django-configurations==0.8
     south==1.0.2
 setenv =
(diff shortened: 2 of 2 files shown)


blueyed:
Thanks.

Would using `passenv PYTHONPATH` instead of `setenv PYTHONPATH` also work?
(ref: https://testrun.org/tox/latest/config.html#confval-passenv=SPACE-SEPARATED-GLOBNAMES).

I am not sure if the `PYTHONPATH = {toxinidir}` is required after all.

blueyed:
@ktosiek 
Can you try if #279 works with NixOS?

ktosiek:
I think it should not be set at all if it's not needed (and if someone has to pass `PYTHONPATH` in because of one's environment, there's always `TOX_TESTENV_PASSENV`) - this commit is only about not overriding what's already there.

blueyed:
@ktosiek 
Yes, I see.  But not overriding anything seems to be better achieved in #279 then.
Do you agree?

ktosiek:
#279 would be nice, but it doesn't seem to work (neither for me locally, nor on Travis)

blueyed:
Yes, you're right.

Therefore: :+1: for this one.

pelme:
Ok, I merged to master in d3e03b90fcfb2be4e0a55bef122291c67a2c22e8 :)

PR #279, opened 2015-10-04
Do not set PYTHONPATH explicitly

blueyed:
tox will prepend the cwd automatically [1].

1: https://bitbucket.org/hpk42/tox/src/3ed5dc353a99acf57859a2dd265b5c2e973480e3/tox/_pytestplugin.py?at=default&fileviewer=file-view-default#_pytestplugin.py-192:193

Ref: https://github.com/pytest-dev/pytest-django/pull/278#issuecomment-145352363

To fix #278, maybe `passenv PYTHONPATH` would be required.

### DIFF
--- a/generate_configurations.py
+++ b/generate_configurations.py
@@ -49,7 +49,6 @@ TOX_TESTENV_TEMPLATE = dedent("""
     deps =
     %(deps)s
     setenv =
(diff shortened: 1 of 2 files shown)


blueyed:
Does not work.

Issue #280, opened 2015-10-06
Problem with setUpClass


MikeVL:
My test code is:
```
class MarketFixtures(object):
    fixtures = [
        'tests/fixtures/common_rubricator.json',
    ]

class MarketOfferPictureTestCase(MarketFixtures, TestCase):
    ...
```
Raised with error
```
/usr/local/lib/python2.7/dist-packages/pytest_django/plugin.py:301: in pytest_runtest_setup
    _disable_class_methods(cls)
/usr/local/lib/python2.7/dist-packages/pytest_django/plugin.py:266: in _disable_class_methods
    _method_is_defined_at_leaf(cls, 'setUpClass'),
/usr/local/lib/python2.7/dist-packages/pytest_django/plugin.py:254: in _method_is_defined_at_leaf
    return getattr(cls.__base__, method_name).__func__ is not getattr(cls, method_name).__func__
E   AttributeError: type object 'MarketFixtures' has no attribute 'setUpClass'
```

blueyed:
Also a problem with tests from django-autocomplete-light: https://travis-ci.org/yourlabs/django-autocomplete-light/jobs/83851177

pelme:
Seems easy to reproduce, will try to push a fix and 2.9.1 release soon!

pelme:
This is now fixed in master, and released in 2.9.1 on PyPI!

PR #281, opened 2015-10-06
Fix for #280.

pelme:
The previous fix did not properly check the full MRO chain when looking
for the super method, which then failed when used with mixins that does
not define setUpClass/tearDownClass.

This commit fixes issue #280.


pelme:
Merged in 1bf4920dd202ebd0a954ef455f19857b2e626d50.

PR #310, opened 2016-01-29
Add support for testing unmanaged models (updated)

timb07:
As noted in ojake's PR #270 (which was created to address #264), the issue with the implementation of the pytest.mark.django_use_model mark that caused tests not to be collected can be fixed relatively easily by supplying the model argument as a kwarg. (I'm not very familiar with pytest, so there may be a better way of fixing this issue.)

This PR includes:
- ojake's django_use_model implementation from #270, modified with…
- the necessary changes for the fix noted above;
- removal of the contextlib.closing() wrapper around connection.schema_editor() (which isn't necessary and doesn't work anyway);
- addition of a check for the existence of django.db.connection.schema_editor before the relevant tests in test_database.py, so the tests are skipped for Django < 1.7.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -288,7 +288,7 @@ def live_server(request):
     --liveserver command line option or if this is not provided from
     the DJANGO_LIVE_TEST_SERVER_ADDRESS environment variable.  If
     neither is provided ``localhost:8081,8100-8200`` is used.  See the
-    Django documentation for it's full syntax.
+    Django documentation for its full syntax.
 
     NOTE: If the live server needs database access to handle a request
           your test will have to request database access.  Furthermore


pytest_django/plugin.py
```diff
@@ -375,6 +379,48 @@ def _django_db_marker(request):
...
+
+    if isinstance(model, (list, tuple)):
+        models = model
+    else:
+        models = (model,)
+
+    with connection.schema_editor() as schema:
+        schema.deferred_sql = []
+        for model_class in models:
+            if not hasattr(model, '_meta'):
```
cleder: should this be model_class ?

halfnibble:
What needs to be done to make this a reality? Does anyone want help on this?

blueyed:
@halfnibble 
I guess you could pick up the branch behind this PR and finish it in a new PR.

timb07:
The new PR has problems with mysql; can anyone help? @halfnibble?

halfnibble:
Yes. I think I will take a look at this. We use MySQL in production, so supporting it is important!

blueyed:
@halfnibble 
Please take a look at the MySQL related failures in #600 then.

PR #314, opened 2016-02-02
tests: update versions, drop support for Python 3.2

blueyed:
Python 3.2 is not supported by pip 8+ anymore, which is used by tox via
Python 3.5's `virtualenv_support`.
Failing build: https://travis-ci.org/pytest-dev/pytest-django/jobs/106463859.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,43 +4,42 @@ language: python
 python:
   - "3.5"
 env:
-  - TESTENV=pypy-2.8.1-master-sqlite_file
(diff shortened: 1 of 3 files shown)


blueyed:
@pelme 
Ping.  Would be nice to have master passing on Travis again.

pelme:
Very nice! :)

blueyed:
btw: using `virtualenv<14` might be an alternative way to fix this (see https://github.com/yyuu/pyenv-virtualenv/commit/1f761275cde2f84440c95bc1dbfbd16e41446a18).

PR #321, opened 2016-03-31
Testcase for #85

thedrow:
This demonstrates the failure I'm experiencing when using the URL marker twice. See #85 for discussion. 
I'm not sure how to fix this but it's failing my test suite as well.

### DIFF
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -63,3 +63,45 @@ def test_urls_cache_is_cleared(testdir):
 
     result = testdir.runpytest_subprocess()
     assert result.ret == 0
+
+def test_urls_cache_is_cleared_and_new_urls_can_be_assigned(testdir):
+    testdir.makepyfile(myurls="""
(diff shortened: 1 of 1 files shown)


tests/test_urls.py
```diff
@@ -63,3 +63,45 @@ def test_something_else():
...
+
+        @pytest.mark.urls('myurls')
+        def test_something():
+            reverse('first')
+
+
+        @pytest.mark.urls('myurls2')
+        def test_something_else():
+            with pytest.raises(NoReverseMatch):
+                reverse('first')
```
blueyed: @thedrow 
I am not sure if this test is really what you want to test: the `urlpatterns` from `myurls2` uses `name='first'`, so I would expect that it does not throw `NoReverseMatch` here actually?!

blueyed:
I've created https://github.com/pytest-dev/pytest-django/pull/355 as a followup and a fixup commit according to my comment above.

PR #323, opened 2016-04-01
added autouse function scoped fixture that cleares django.contrib.sites.models.SITE_CACHE

### LINKED ISSUE #322: Tests with django sites framework onetoonefield causes unexpected behavior
Assume you have a model:

```
class Customer(models.Model):
    site = models.OneToOneField('sites.Site')
```

And when using the sites middleware, without setting SITE_ID, the site is looked up and cached based on the requests host information: https://github.com/django/django/blob/master/django/contrib/sites/models.py#L12

This causes unexpected behavior if testing a multi tenant site, as the request.site object will be the one from the SITE_CACHE, that might have an already populated _request.site.customer_ from previous execution.

I will submit a proposal for fixing this, as it can cause plenty of pain when debugging :)

peterlauri:
Closes #322

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,6 +9,9 @@ Features
 * Added new function scoped fixture ``mailoutbox`` that gives access to
   djangos ``mail.outbox``. The will clean/empty the ``mail.outbox`` to
   assure that no old mails are still in the outbox.
+* If ``django.contrib.sites`` is in your INSTALLED_APPS, Site cache will
+  be cleared for each test to avoid hitting the cache and cause wrong Site
+  object to be returned by ``Site.objects.get_current()``.
 
 Compatibility
 ^^^^^^^^^^^^^
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -239,3 +239,18 @@ Example
         assert m.from_email == 'from@example.com'
         assert list(m.to) == ['to@example.com']
 
+
+Environment autouse fixtures
+----------------------------
+
+pytest-django provides some pytest fixtures that are of autouse
+nature. They provide functionality to assure a clean environment
+during tests.
+
+
+Clearing of site cache
+~~~~~~~~~~~~~~~~~~~~~~
+
+If ``django.contrib.sites`` is in your INSTALLED_APPS, Site cache will
+be cleared for each test to avoid hitting the cache and cause wrong Site
+object to be returned by ``Site.objects.get_current()``.
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -539,6 +539,20 @@ def _template_string_if_invalid_marker(request):
             else:
                 dj_settings.TEMPLATE_STRING_IF_INVALID.fail = False
 
+
+@pytest.fixture(autouse=True, scope='function')
+def _django_clear_site_cache():
+    """Clears ``django.contrib.sites.models.SITE_CACHE`` to avoid
+    unexpected behavior with cached site objects.
+    """
+
+    if django_settings_is_configured():
+        from django.conf import settings as dj_settings
+
+        if 'django.contrib.sites' in dj_settings.INSTALLED_APPS:
+            from django.contrib.sites.models import Site
+            Site.objects.clear_cache()
+
 # ############### Helper Functions ################
 
 
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -3,9 +3,12 @@ from __future__ import with_statement
 import os
 
 import pytest
+from django.contrib.sites.models import Site
+from django.contrib.sites import models as site_models
 from django.core import mail
 from django.db import connection
 from django.test import TestCase
+from pytest_django.lazy_django import get_django_version
 
 from pytest_django_test.app.models import Item
 
@@ -215,3 +218,26 @@ class TestrunnerVerbosity:
             "*PASSED*"])
         assert ("*Destroying test database for alias 'default' ('*')...*"
                 not in result.stdout.str())
+
+
+@pytest.mark.skipif(
+    get_django_version() < (1, 8),
+    reason='Django 1.7 requires settings.SITE_ID to be set, so this test is invalid'
+)
+@pytest.mark.django_db
+@pytest.mark.parametrize('site_name', ['site1', 'site2'])
+def test_clear_site_cache(site_name, rf, monkeypatch):
+    request = rf.get('/')
+    monkeypatch.setattr(request, 'get_host', lambda: 'foo.com')
+    Site.objects.create(domain='foo.com', name=site_name)
+    assert Site.objects.get_current(request=request).name == site_name
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('site_name', ['site1', 'site2'])
+def test_clear_site_cache_check_site_cache_size(site_name, settings):
+    assert len(site_models.SITE_CACHE) == 0
+    site = Site.objects.create(domain='foo.com', name=site_name)
+    settings.SITE_ID = site.id
+    assert Site.objects.get_current() == site
+    assert len(site_models.SITE_CACHE) == 1


pelme:
Thanks for the PR!

There are some failing tests, and we also need an internal test to make sure that this fix works.

Apart from that, this PR looks good to me!

peterlauri:
@pelme I fixed up tests, added one test for site cache clear, should be ok to go. Please note the comment in the test, that it needs refactoring if we choose to use xdist -n option...

peterlauri:
Anything missing in this PR to get merged?

pytest_django/plugin.py
```diff
@@ -514,6 +514,20 @@ def _template_string_if_invalid_marker(request):
             else:
                 dj_settings.TEMPLATE_STRING_IF_INVALID.fail = False
 
+
+@pytest.fixture(autouse=True, scope='function')
+def django_clear_site_cache():
+    """Clears ``django.contrib.sites.models.SITE_CACHE`` to avoid
+    unexpected behavior when cached site objects.
```
blueyed: s/when/with/

pytest_django/plugin.py
```diff
@@ -514,6 +514,20 @@ def _template_string_if_invalid_marker(request):
...
+    """
+
+    if django_settings_is_configured():
+        from django.conf import settings as dj_settings
+
+        if 'django.contrib.sites' in dj_settings.INSTALLED_APPS:
+            from django.contrib.sites import models
+            models.SITE_CACHE.clear()
```
blueyed: I think that is quite some overhead for every test, especially when not using sites at all etc.

Can we create / request this fixture dynamically only if the condition is met once?

What about providing the fixture, but not automatically use it?
Even when you require it, it's not necessary for all your tests.

peterlauri: I think the benefit of this autouse fixture overcomes the overhead of a list lookup, in comparison with all setup/teardown that is done with databases etc. 

We could cache the list lookup, but is that really worth it? Can we guarantee that INSTALLED_APPS are not modified by tests?

The core reason for doing this, is that the SITE_CACHE is evil :), and people that write tests for their projects where sites are used, they will tear their hair off when site objects suddenly starts to mismatch with referenced objects. The SITE_CACHE is nice for production, but not for testing...

I was against having the mailbox cleaned with an autouse fixture, as we talked about in #410. But this is a different thing, this is a cache that sits there in its evil darkness and just makes people that loves testing wanna kill them self when debugging strange errors that the SITE_CACHE causes...

pelme: I think this PR is reasonable. People that are not using the site framework (not having it in INSTALLED_APPS) won't notice any difference. 

There is a helper to clear the cache `Site.objects.clear_cache()` that I think should be used instead of directly clearing the SITE_CACHE which from what I understand is not a public API:
https://docs.djangoproject.com/en/1.10/ref/contrib/sites/#caching-the-current-site-object

pytest_django/plugin.py
```diff
@@ -514,6 +514,20 @@ def _template_string_if_invalid_marker(request):
             else:
                 dj_settings.TEMPLATE_STRING_IF_INVALID.fail = False
 
+
+@pytest.fixture(autouse=True, scope='function')
+def django_clear_site_cache():
```
pelme: We should prefix this fixture with a `_` to make it clear that it is an internal fixture that is not meant to be requested directly by tests.

tests/test_fixtures.py
```diff
@@ -332,3 +332,15 @@ def test_unblock_manually(self, django_db_blocker):
...
+@pytest.mark.parametrize('run_no', [(1,), (2,)])
+def test_clear_site_cache(run_no):
+    """
+    Run this two times, would fail on run #2 if SITE_CAHCE would not be cleared.
+
+    Note that this test might start to fail if we choose to use pytest-xdist -n.
+    """
+    from django.contrib.sites.models import SITE_CACHE
+    assert len(SITE_CACHE) == 0
+    SITE_CACHE[run_no] = 'run_no {}'.format(run_no)
```
pelme: From what I understand (see the comment about the `SITE_CACHE.clear()` too): SITE_CACHE is not part of a public Django API. This test should be rewritten to make sure the cache is cleared by checking `Site.objects.get_current()` which seems to be the public API to interact with the site framework.

Something like this will probably do (I have not actually run this code and probably missed some detail):
```py
@pytest.mark.django_db
@pytest.mark.parametrize('name', ['site1', 'site2'])
def test_clear_site_cache(settings, site_name):
    settings.SITE_ID = 1
    Site.objects.create(id=1, domain='foo.com', name=site_name)
    assert Site.objects.get_current().name == site_name
```

peterlauri:
Btw, where do we document internal fixtures? I mean, we should describe what they do, right?

blueyed:
> Btw, where do we document internal fixtures? I mean, we should describe what they do, right?

Use a docstring?!

peterlauri:
Should we really assume that the user reads the code? My philosophy is that we should assume that all functionality is described in the public docs (readthedocs). Even that the code is documented, the normal user (read newbie), it is not sufficient. Or are we generating docs from docstrings to the public docs?

tests/test_fixtures.py
peterlauri: The test is not really correct, but I get your point :) SITE_CACHE is never used if settings.SITE_ID is set, it is looked up based on hostname... but I use similar approach, unless my "unit test" that checks SITE_CACHE is enough. But it is not really enough, as in future django versions SITE_CACHE might change name or use some other way of caching...

pelme:
Good point, we should document that we clean the sites cache but I don't think we should document "_django_clear_site_cache" explicitly in the user documentation. For that purpose a docstring is good.

We could add a section in helpers that describe additional things we do for each Django test?

peterlauri:
Perfect, I add a section there if magic stuff that happens in
pytest-django. Review the test, fails without the fixture, passes with it.

pelme:
Excellent! 👍 

Just prefix the fixture with `_` and add docs. :)

peterlauri:
@pelme added docs, changed _ and also put some notes in changelog. Done for final review.

blueyed:
Fails for TOXENV=py27-pytest30-django1.7-postgres (https://travis-ci.org/pytest-dev/pytest-django/jobs/175482486).

docs/changelog.rst
```diff
@@ -9,6 +9,9 @@ Features
 * Added new function scoped fixture ``mailoutbox`` that gives access to
   djangos ``mail.outbox``. The will clean/empty the ``mail.outbox`` to
   assure that no old mails are still in the outbox.
+* If ``django.contrib.sites`` is in your INSTALLED_APPS, Site cache will
+  be cleared for each tests to avoid hitting cache and cause wrong site
```
blueyed: s/tests/test/

blueyed: s/hitting/hitting the/

blueyed: s/site/Site/

docs/helpers.rst
```diff
@@ -239,3 +239,18 @@ Example
         assert m.from_email == 'from@example.com'
         assert list(m.to) == ['to@example.com']
 
+
+Environment autouse fixtures
+----------------------------
+
+pytest-django provides some pytest fixtures that are of autouse
+nature. They provide functionality to assure a clean environment
+for pytest tests.
```
blueyed: s/for pytest tests/during tests/ ?!

docs/helpers.rst
```diff
@@ -239,3 +239,18 @@ Example
...
+
+
+Clearing of site cache
+~~~~~~~~~~~~~~~~~~~~~~
+
+If ``django.contrib.sites`` is in your INSTALLED_APPS, Site cache will
+be cleared for each tests to avoid hitting cache and cause wrong site
+objects to be returned by ``Site.objects.get_current()``.
```
blueyed: Same typos as above.

peterlauri:
The problem is that Django==1.7 does require SITE_ID to be set if using the sites framework, so we cannot test it for the case where the request object is used to map to correct Site object.

I have added a unit test that covers Django==1.7 as well, that checks that SITE_CACHE is emptied...

docs/changelog.rst
peterlauri: done

docs/helpers.rst
peterlauri: done

docs/helpers.rst
peterlauri: done

pytest_django/plugin.py
peterlauri: done

tests/test_environment.py
```diff
@@ -4,9 +4,11 @@
 
 import pytest
 from django.contrib.sites.models import Site
+from django.contrib.sites.shortcuts import get_current_site
```
blueyed: Unused import?

peterlauri: that was fixed in e085c1b

peterlauri:
squashed...

pelme:
Thanks a lot! :)

peterlauri:
Nice to contribute :)

PR #326, opened 2016-04-06
Make help for --create-db more clear

bedwards:
I thought it was a typo, but I now believe that --create-db does rely on --reuse-db. I don't understand why that would be the case, so maybe this isn't the correct fix.


pelme:
Thanks for the suggestion! I reworded the text a bit more to hopefully make it clearer. :)

bedwards:
Ah! Yes, that makes sense.

PR #331, opened 2016-04-28
Add support for serialized rollback. Fix #329.

### LINKED ISSUE #329: Tests using live_server fixture removing data from data migrations
I've created a simple test case to reproduce this behavior https://github.com/ekiro/case_pytest/blob/master/app/tests.py which fails after second test using live_server fixture. 
MyModel objects are created in migration, using RunPython. It looks like after any test with live_server, every row from the database is truncated.

aaugustin:
WORK IN PROGRESS - UNTESTED

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -3,6 +3,15 @@ Changelog
 
 NEXT
 ----
+
+Features
+^^^^^^^^
+* Add support for serialized rollback in transactional tests.
+  Thanks to Piotr Karkut for `the bug report
+  <https://github.com/pytest-dev/pytest-django/issues/329>`_.
+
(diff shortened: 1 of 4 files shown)


blueyed:
I've picked this up and rebased it in https://github.com/pytest-dev/pytest-django/pull/353.

aaugustin:
Great, thanks! I'm closing this PR in favor of #353, then.

If memory serves, the patch didn't work properly, which you've probably noticed by now...

blueyed:
Well, there's no test for it yet, but it had a wrong name for some kwarg.
Before it gets merged a test would be needed of course.
Let's move the discussion over to #353.

PR #339, opened 2016-05-26
Fix get_django_version() asserts for django_user_model / django_username_field

adamchainz:
Just spotted this. It doesn't fail because functions can be compared with types fine on Python 2, and no one is using Django < 1.5 on Python 3. Still worth fixing.


pelme:
Thank you! 👍 😄

PR #347, opened 2016-06-22
Add faq entry for a manage.py test integration

tomviner:
As discussed at Pytest Sprint, alternative approach for https://github.com/pytest-dev/pytest-django/pull/225

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -63,6 +63,54 @@ transactions are tested or not.
 
 .. _faq-getting-help:
 
+How can I use ``manage.py test`` with pytest-django?
+----------------------------------------------------
+
+pytest-django is designed to work with the ``py.test`` command, but if you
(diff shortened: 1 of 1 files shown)


ekohl:
I think this is a fine solution and it would work for me, but I'm wondering what the argument is against including this in the package itself.

tomviner:
@ekohl the arguments against adding this to the code base, are:
- this is not the recommended invocation of pytest-django
- adding it and documenting it confuses users
- the options `--ds` and `--dc` won't work as Django selects the settings file before pytest-django is loaded
- splitting the arguments with `--` is slightly confusing
- the test runner is simple to add to ones own codebase if really required

ekohl:
Fair enough. I'll be happy with this given we have a docker-compose setup which uses django-admin as the entrypoint and it's documented now.

PR #349, opened 2016-06-24
generate_configurations: remove UID / db creation from tox configs

blueyed:
This is only required with Django versions before 1.7, and is meant to
make the diff of config updates simpler.


blueyed:
Requires #350 to be merged first.

PR #356, opened 2016-06-24
Improve the handling of sqlite_file tests

pelme:


tests/test_unittest.py
```diff
@@ -71,7 +71,7 @@ class TestFoo(TestCase):
             def test_foo(self):
                 # Make sure we are actually using the test database
                 db_name = settings.DATABASES['default']['NAME']
-                assert db_name.startswith('test_') or db_name == ':memory:'
+                assert 'test_' in db_name or db_name == ':memory:'
```
blueyed: `.startswith('test_')` instead of `in`?

blueyed: But then you also need the `os.path.split` probably?! - so it's fine like that.

PR #359, opened 2016-06-25
Block Django database access as early as possible.

pelme:
This should fix various places where the production database could accidentally been accessed.

/cc @blueyed


pelme:
I went ahead and merged this, I am working on refactoring the database setup fixtures and needed it. Let me know if you have any comments @blueyed :)

PR #362, opened 2016-07-02
Database fixture refactor

pelme:
This PR brings refactoring of the internal pytest-django fixtures to make them easily overridable.

This PR is intended to solve most issues tagged with the db-configuration label:
https://github.com/pytest-dev/pytest-django/issues?q=is%3Aopen+is%3Aissue+label%3Adb-configuration.

It also removes the previously `_get_test_db_name` hack with a proper way to achieve the same outcome.

The changes in fixtures.py and plugin.py is mostly done without breaking backward compatibility.

Todo:
* [x] Document the new fixtures
* [x] Go through issues and add documentation examples on how to solve the particular cases and/or modify the code


pelme:
@blueyed It would be helpful if you would take a quick look especially at https://github.com/pytest-dev/pytest-django/blob/eb167359d5a5e8da7c7378f5f1f484d8aef6b9be/pytest_django/fixtures.py#L21-L97 and provide some early feedback if the those fixtures makes sense to you. I have not yet written the docs for them, but some early feedback would be nice. :)

pelme:
(There are docs for the `django_db_blocker` fixture, I intend to write similar docs for the others too.)

pytest_django/fixtures.py
```diff
@@ -7,71 +7,106 @@
...
 
 
-# ############### Internal Fixtures ################
-
 @pytest.fixture(scope='session')
-def _django_db_setup(request,
-                     _django_test_environment,
-                     _django_cursor_wrapper):
-    """Session-wide database setup, internal to pytest-django"""
+def django_db_modify_db_settings(request):
```
blueyed: This only does the xdist suffix handling, right?
Would be good to factor `django_db_test_name` out of it (where it would only build the test name), and then have this do the xdist related change only - which could also be its own fixture though.
In the end it would be good if you could override a fixture that gets a test_name and return your modified version.

pytest_django/plugin.py
```diff
@@ -14,17 +14,24 @@
...
+from .fixtures import admin_client  # noqa
+from .fixtures import admin_user  # noqa
+from .fixtures import client  # noqa
+from .fixtures import db  # noqa
+from .fixtures import django_user_model  # noqa
+from .fixtures import django_username_field  # noqa
+from .fixtures import live_server  # noqa
+from .fixtures import rf  # noqa
+from .fixtures import settings  # noqa
+from .fixtures import transactional_db  # noqa
```
blueyed: I think this should use the previous style, and then you would only need one `# noqa`, too.

pytest_django/plugin.py
```diff
@@ -231,7 +238,7 @@ def pytest_load_initial_conftests(early_config, parser, args):
 
         # Forcefully load django settings, throws ImportError or
         # ImproperlyConfigured if settings cannot be loaded.
-        from django.conf import settings
+        from django.conf import settings  # noqa
```
blueyed: Why is that `# noqa` needed?

blueyed: Overwrites a previous import?

pytest_django/plugin.py
```diff
@@ -324,20 +331,26 @@ def _django_test_environment(request):
     """
     if django_settings_is_configured():
         _setup_django()
-        from django.conf import settings
+        from django.conf import settings  # noqa
```
blueyed: Same question as above.

pytest_django/plugin.py
```diff
@@ -324,20 +331,26 @@ def _django_test_environment(request):
...
-    """Wrapper around Django's database access, internal to pytest-django.
+@pytest.fixture(scope='session')
+def django_db_blocker():
+    """Wrapper around Django's database access.
+
+    This object can be used to re-enable database access.  This fixture is used
+    internally in pytest-django to build the other fixtures and can be used for
+    special database handling.
+
+    The object is a context manager and returned provides the methods
```
blueyed: s/returned //

pytest_django/plugin.py
```diff
@@ -324,20 +331,26 @@ def _django_test_environment(request):
...
+    .enable_database_access()/.disable_database_access() and .restore() the
+    temporarily enable database access.
```
blueyed: s/the/to/

blueyed: `restore` has been renamed to `restore_previous_access`.

docs/database.rst
```diff
@@ -128,3 +128,51 @@ Using ``--nomigrations`` will disable Django migrations and create the database
...
+You can also manage the access manually via these methods:
+
+.. py:method:: django_db_blocker.enable_database_access()
+
+  Enable database access. Must be followed by a call to
+  :func:`~django_db_blocker.restore_previous_access`.
+
+.. py:method:: django_db_blocker.disable_database_access()
+
+  Enable database access. Must be followed by a call to
```
blueyed: s/Enable/Disable/

docs/database.rst
```diff
@@ -128,3 +128,51 @@ Using ``--nomigrations`` will disable Django migrations and create the database
...
+  :func:`~django_db_blocker.restore_previous_access`.
```
blueyed: s/restore/enable/

blueyed: Although it must not really be followed..

blueyed: Both methods should probably say that they should be followed by a call to `restore_previous_access`?

pytest_django/plugin.py
pelme: It overrides the settings fixture which is named the same.

pytest_django/plugin.py
pelme: I couldn't get flake8 to shut up with the previous style (previously we also had a useless tuple). I tried `#noqa`, `#flake8: noqa`, on the first line, last line and all lines but no success.

pytest_django/plugin.py
blueyed: Hmm, should we use `django_settings` for the fixture then instead?  Makes it more explicit anyway.
The old name would still be available, but deprecated.

pytest_django/fixtures.py
pelme: I figured it gives more flexibility to just change the settings dict however you like. Then you can influence not only test db name, but also hostname etc.

What do you think about a fixture that is called `django_db_modify_db_settings_xdist_suffix`, and then have `django_db_modify_db_settings` be implemented as:
```py
def django_db_modify_db_settings(django_db_modify_db_settings_xdist_suffix):
    pass
```
?

Then it would be trivial to override with or without xdist-naming or anything else.

pytest_django/plugin.py
pelme: That would be good. 👍

pytest_django/plugin.py
blueyed: I've thought that the following would work, but apparently does not:
```python
from foo import (foo, bar,
                 baz)  # noqa
```

I'd rather use the useless tuple then as a workaround for now though.
An alternative might be to define `__all__`?

pelme: With `__all__` we also have to add stuff in plugins.py. The tuple is annoying, then you have to change stuff at two places.

pytest_django/plugin.py
blueyed: Do you want to do it in this PR also?
I could give it a try, but then we certainly get some conflicts - but it should be fine.

pelme: Please go ahead :) It's fine as its own PR, I can fixup the conflicts when I rebase this branch.

docs/database.rst
```diff
@@ -132,21 +132,101 @@ migrations in case ``--nomigrations`` is used, e.g. in ``setup.cfg``.
...
+
+This fixture is by default requested from :fixture:`django_db_setup`.
+
+django_db_use_migrations
+""""""""""""""""""""""""
+
+.. fixture:: django_db_use_migrations
+
+Returns True/False whether or not to use migrations to create the test
+databases.
```
blueyed: Move the "True/False" to the end in brackets?!  Or remove it completely?

docs/database.rst
```diff
@@ -132,21 +132,101 @@ migrations in case ``--nomigrations`` is used, e.g. in ``setup.cfg``.
...
+
+This fixture is by default requested from :fixture:`django_db_setup`.
+
+django_db_keepdb
+""""""""""""""""
+
+.. fixture:: django_db_keepdb
+
+Returns True/False wheter or not to keep the re-use an existing database and to
+keep it after the test run.
```
blueyed: Same as above regarding True/False.
s/wheter/whether/

blueyed: s/keep the//

pelme:
@blueyed I think I've documented the different use cases I have found in the issue tracker (and I've built small Django projects and verified that they actually works). Can you come up with more ideas that should be part of the examples section? I think it is a pretty good start to showcase how to customize stuff anyways.

Is there anything that you would like to change API-wise, or anything else before this gets merged?

PR #363, opened 2016-07-04
minor: formatting / line-length fixes

blueyed:


pelme:
Could we enforce this with flake8?

I usually prefer 100 or 120 as max line length since i think 80 leads to too many line breaks pretty much everywhere.

I'm fine having either but it would be nice with consistency from the linter. :)

blueyed:
We can make flake8 check it, but it would fail a lot currently.
I think we should aim for the limit recommended by PEP8 (79).  You can always use `# noqa` with longer lines, but usually it results in code that is more readable.
Therefore I'd say to not enforce it for now, but keep it in mind, and add it later as a check.

blueyed:
Thanks!  Also for squashing the commit yourself.
However, in that case (when using the GitHub UI) you have to manually trim the commit message.

pelme:
Good point. I was a bit too quick on the merge buttons. I'll fix the commit message in future merges :)

blueyed:
Awesome, thanks!
Do you plan to release version 3 soon?

pelme:
Hopefully today.

I want to get support for current Django master, have a second look at the db-fixture stuff and test the release with my own test suite.

PR #364, opened 2016-07-05
Add type=bool for django_find_project/INVALID_TEMPLATE_VARS_ENV

blueyed:
This refactors the _parse_django_find_project_ini method into
_get_boolean_value, and uses it for INVALID_TEMPLATE_VARS_ENV, too.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -74,13 +74,13 @@ def pytest_addoption(parser):
     parser.addini('django_find_project',
                   'Automatically find and add a Django project to the '
                   'Python path.',
-                  default=True)
+                  type='bool', default=True)
     group._addoption('--fail-on-template-vars',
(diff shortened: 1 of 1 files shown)


pelme:
Hmm, did travis run?

blueyed:
Now it does - pushed the same commit with a new hash.

blueyed:
Not sure about since when `type='bool'` is supported by pytest though.

blueyed:
`type='bool'` is supported since 2.9.0 only.

blueyed:
We could require 2.9 or do something like https://github.com/s0undt3ch/pytest-instafail/blob/733c301eb848ff38c16613670f4797e88c081478/pytest_instafail.py#L24-L40.

pelme:
Thanks! pytest >=2.9 is fine :)

PR #382, opened 2016-08-22
README - link to changelog

adamchainz:
I visit PyPI pages when upgrading packages to try find the projects' changelogs. This would help with such hunting, making it more obvious :)


blueyed:
Thanks! :+1:

PR #383, opened 2016-08-22
Changelog - fix some code highlighting

adamchainz:
Markdown code highlighting uses one tick but RST uses two, I suspect these were just a mixup from that confusion.


blueyed:
Thanks, LGTM!

PR #390, opened 2016-09-08
tox + travis-ci changes

steenzout:
- cache directories where virtualenvs reside to speed up process
- trigger different builds per Python version
- massive shrink of `tox.ini`
- minor cleanup of `.travis.yml`

checked carefully and the number of virtualenvs tested on travis-ci is the same.

before changes [total elapsed time = 20m](https://travis-ci.org/pytest-dev/pytest-django/builds/156458481).

[currently it's around 15m](https://travis-ci.org/steenzout/python-pytest-django/builds/158359954).


steenzout:
anyone?

.travis.yml
```diff
@@ -2,31 +2,29 @@
...
   allow_failures:
-    - env: TESTENV=python2.7-3.0.0-master-postgres
-    - env: TESTENV=python3.5-3.0.0-master-postgres
+    - env: TOXENV=py27-pytest300-djangomaster-postgres
+    - env: TOXENV=py35-pytest300-djangomaster-postgres
+
+cache:
+  directories:
+    - "${HOME}/virtualenv"
+    - "${TRAVIS_BUILD_DIR}/.tox"
```
blueyed: What about caching just `~/.cache/pip` instead?
After all the cache size is also an issue, since it will be fetched before the builds.

What is `${HOME}/virtualenv` used for?  Isn't that Travis' location for the Python?!

tox.ini
```diff
@@ -1,1978 +1,58 @@
...
+envlist =
+    - py35-pytest300-django{master,1.10,1.9,1.8}-postgres
+    - py34-pytest300-django1.10-postgres
+    - py33-pytest300-django1.8-postgres
+    - py27-pytest300-django1.10-{mysql_innodb,mysql_myisam,postgres}
+    - py27-pytest300-django{master,1.9,1.8,1.7}-postgres
+    - pypy3-pytest300-django1.8-{sqlite,sqlite_file}
+    - pypy3-pytest292-django1.8-sqlite_file
+    - pypy-pytest300-django1.10-sqlite_file
+    - py{35,34,33,27,py3,py}-checkqa
```
blueyed: This (checkqa) was done for py27 and py35 only before.

.travis.yml
steenzout: I think the packages on the `install:` section get installed on the `virtualenv`:

```
install:
  (...)
  - pip install tox==2.3.1
  - pip install tox-travis
```

so I typically add these but I can also remove this if you like.

steenzout: I can try and test just caching ``/.cache/pip`.
I'll need to do it tonight though (I'm on US/Mountain Time).

if you want to merge this one and a new PR for the cache change just let me know.

steenzout:
sorry, I didn't see `generate_configurations.py`.
it crossed my mind you might have been using a script to maintain this big `tox.ini` though :)
I've removed it now.

for me it has been easy to maintain the envs between `tox.ini` and `.travis.yml`.

regarding the `--skip-missing-interpreters` I think the docs say it will return success if the interpreter is missing.
do you want me to add this and remove the `allow_failures`?

blueyed:
I think `allow_failures` is different.
It seems like on Travis with Python 3.5 you have all older Python versions, too (but not the other way around).  But this PR adds https://pypi.python.org/pypi/tox-travis, which seems to handle this better.

It's :+1: from me, although I somehow dislike that there is now only a build per Python version, which makes it hard to see at first glance that e.g. all django 1.10 builds failed (because you need to look at each job's output to figure that out).

@pelme 
What do you think?

steenzout:
I can make it run a single complete test per test environment in parallel.
it's pretty simple.

let me know.

blueyed:
> I can make it run a single complete test per test environment in parallel.

Sounds good.. let's see how it compares in term of duration.

steenzout:
FYI caching `.pip` is not good enough.
some files seem to be updated in the installation process therefore you'll always extra overhead on top of each build.

.travis.yml
steenzout: FYI caching `.pip` is not good enough.
some files seem to be updated in the installation process therefore you'll always extra overhead on top of each build since the cache will always be changed and an upload will need to be made.

steenzout:
I've enabled the parallel builds, removed `tox-travis` due to that and changed the directories being cached (improves build performance and shortens cache size).

steenzout:
ping.

blueyed:
Seems to be much slower now in total ( 51 min 48 sec) vs initially (39 min 10 sec).
Elapsed time is 13 min 18/15 min 6 sec vs 13 min 33 sec now.

I am not really convinced that it's better than the previous approach through `generate-configurations.py`.

steenzout:
trigger a rebuild and see if the cache helps.

[mine](https://travis-ci.org/steenzout/python-pytest-django) was faster.

pelme:
Thanks a lot! This looks really good. I'm very happy to move to this. When `generate_configurations.py` was conceived, tox parametrization was not a thing. :)

pelme:
@steenzout I updated to pytest 3.0.3 and merged this. Thanks a lot @steenzout ! 👍 😃

PR #391, opened 2016-09-09
Add is_blocked to django_db_blocker

mdentremont:
- In some cases it is nice to be able to detect whether the DB is setup

### DIFF
--- a/tests/test_fixtures.py
+++ b/tests/test_fixtures.py
@@ -361,3 +361,19 @@ class Test_django_db_blocker_is_blocked:
 
         django_db_blocker.block()
         assert django_db_blocker.is_blocked
+
+    def test_with_unblock_context_manager(self, django_db_blocker):
+        assert django_db_blocker.is_blocked
+
+        with django_db_blocker.unblock():
+            assert not django_db_blocker.is_blocked
+
+        assert django_db_blocker.is_blocked
+
+    def test_with_block_context_manager(self, django_db_blocker, db):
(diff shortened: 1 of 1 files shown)


mdentremont:
@blueyed Thoughts on this one?

mdentremont:
@blueyed Little delayed 😛

pelme:
Thanks for the PR!

Where would this be useful? Can you give an example where one wants to check this rather than just enabling/disabling?

mdentremont:
@pelme We have fixtures which might just instantiate a Django model vs actually persisting to the DB; this allows us to minimize how many tests need to use the DB.

An example of one case where we would use this would be the user model; unfortunately I have since switched companies, so I am not able to look for specific examples 😛

pelme:
So you really want to know whether or not the db fixture is active or not? The db blocker is just a utility to block the database, I'm not sure this is the correct location to figure out whether or not database access was requested in a fixture.

You could also achieve the same by checking `'db' in request._fixture_values`. It is a bit of a hack but conditionally changing fixture behavior is a bit of a hack. :)

I'd like to have a clear example where this would be useful and see if it can already be done with the current implementation.

rygwdn:
Like @mdentremont said, the intention is to determine whether or not the db is available (i.e. unblocked) so we can determine whether or not to `.save()` a model in a fixture.  We considered using something like `'db' in request._fixture_values` but I don't think that worked when we used `transactional_db` or `@pytest.mark.django_db`

PR #393, opened 2016-09-12
Avoid deprecation warnings on pytest 3.0

mdentremont:
- Use getfixturevalue instead of getfuncargvalue if it exists

### DIFF
--- a/pytest_django/compat.py
+++ b/pytest_django/compat.py
@@ -22,4 +22,3 @@ def getfixturevalue(request, value):
         return request.getfixturevalue(value)
 
     return request.getfuncargvalue(value)
-
--- a/pytest_django/compat.py
+++ b/pytest_django/compat.py
@@ -18,8 +18,8 @@ except ImportError:
 
 
 def getfixturevalue(request, value):
-    try:
+    if hasattr(request, 'getfixturevalue'):
         return request.getfixturevalue(value)
-    except AttributeError:
-        return request.getfuncargvalue(value)
+
+    return request.getfuncargvalue(value)
 
--- a/pytest_django/compat.py
+++ b/pytest_django/compat.py
@@ -1,3 +1,5 @@
+# This file cannot be imported from until Django sets up
+
 try:
     # Django 1.11
     from django.test.utils import setup_databases, teardown_databases  # noqa
@@ -15,10 +17,3 @@ try:
 except ImportError:
     # Django 1.7.
     from django.db.backends import BaseDatabaseWrapper  # noqa
-
-
-def getfixturevalue(request, value):
-    if hasattr(request, 'getfixturevalue'):
-        return request.getfixturevalue(value)
-
-    return request.getfuncargvalue(value)
--- a/pytest_django/compat.py
+++ b/pytest_django/compat.py
@@ -15,3 +15,11 @@ try:
 except ImportError:
     # Django 1.7.
     from django.db.backends import BaseDatabaseWrapper  # noqa
+
+
+def getfixturevalue(request, value):
+    try:
+        return request.getfixturevalue(value)
+    except AttributeError:
+        return request.getfuncargvalue(value)
+
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -9,6 +9,7 @@ import pytest
 from . import live_server_helper
 
 from .django_compat import is_django_unittest
+from .pytest_compat import getfixturevalue
 
 from .lazy_django import get_django_version, skip_if_no_django
 
@@ -154,7 +155,6 @@ def db(request, django_db_setup, django_db_blocker):
     """
     if 'transactional_db' in request.funcargnames \
             or 'live_server' in request.funcargnames:
-        from .compat import getfixturevalue
         getfixturevalue(request, 'transactional_db')
     else:
         _django_db_fixture_helper(False, request, django_db_blocker)
@@ -323,5 +323,4 @@ def _live_server_helper(request):
     function-scoped.
     """
     if 'live_server' in request.funcargnames:
-        from .compat import getfixturevalue
         getfixturevalue(request, 'transactional_db')
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -8,6 +8,7 @@ import pytest
 
 from . import live_server_helper
 
+from .compat import getfixturevalue
 from .django_compat import is_django_unittest
 
 from .lazy_django import get_django_version, skip_if_no_django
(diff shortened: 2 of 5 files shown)


mdentremont:
Unfortunately I've addressed the warnings by using duplicated try/except blocks - it might be nice to extract a helper function to avoid the duplication, similar to pytest-bdd https://github.com/pytest-dev/pytest-bdd/blob/30076d478c34afc67c41460a2956b430f15e7f90/pytest_bdd/utils.py#L26

I didn't see any other extracted functions in the codebase, so didn't want to break any patterns.

Let me know if you have any suggestions!

rpkilby:
A compat module already exists - it seems like would be appropriate there.

```py
# compat.py
def getfixturevalue(request, value):
    try:
        return request.getfixturevalue(value)
    except AttributeError:
        return request.getfuncargvalue(value)

# fixtures.py
from compat import getfixturevalue
...

getfixturevalue('transactional_db')
```

mdentremont:
@rpkilby Thanks, done!

pip-selfcheck.json
```diff
@@ -0,0 +1 @@
+{"last_check":"2016-09-26T19:36:56Z","pypi_version":"8.1.2"}
```
blueyed: That should not be in here.

blueyed:
> What warning were you seeing?

Sorry, apparently I only saw the last commit(s) when reviewing.

mdentremont:
@blueyed Any ideas why the builds are failing on me?

pip-selfcheck.json
mdentremont: Oops - deleted

mdentremont: I added it to the .gitignore as it seems to be generated when I try to run the tests

mdentremont:
.. Going to try reverting everything to see if there build will pass

blueyed:
@mdentremont 
You should not import the .compat module at module level (because that also does Django imports).

```diff
 .../deoplete/sources/deoplete_jedi/cache.py        | 37 ++++++++++------------
 .../deoplete/sources/deoplete_jedi/worker.py       |  3 +-
 2 files changed, 18 insertions(+), 22 deletions(-)

diff --git a/rplugin/python3/deoplete/sources/deoplete_jedi/cache.py b/rplugin/python3/deoplete/sources/deoplete_jedi/cache.py
index 2c84fd5..20d2422 100644
--- a/rplugin/python3/deoplete/sources/deoplete_jedi/cache.py
+++ b/rplugin/python3/deoplete/sources/deoplete_jedi/cache.py
@@ -145,29 +145,25 @@ def reap_cache(max_age=300):
     Module level completions are exempt from reaping.  It is assumed that
     module level completions will have a key length of 1.
     """
-    with _cache_lock:
-        now = time.time()
-        cur_len = len(_cache)
-        for cached in list(_cache.values()):
-            if cached.key[-1] not in ('package', 'local', 'boilerplate~',
-                                      'import~') \
-                    and now - cached._touched > max_age:
-                _cache.pop(cached.key)
-        return len(_cache), cur_len
+    while True:
+        time.sleep(300)
+
+        with _cache_lock:
+            now = time.time()
+            cur_len = len(_cache)
+            for cached in list(_cache.values()):
+                if cached.key[-1] not in ('package', 'local', 'boilerplate~',
+                                          'import~') \
+                        and now - cached._touched > max_age:
+                    _cache.pop(cached.key)
+
+            if cur_len - len(_cache) > 0:
+                log.debug('Removed %d of %d cache items', len(_cache), cur_len)
 
 
 def cache_processor_thread(compl_queue):
-    last_clear = time.time()
     errors = 0
     while True:
-        now = time.time()
-        if now - last_clear >= 30:
-            after, before = reap_cache()
-            reaped = before - after
-            if reaped > 0:
-                log.debug('Removed %d of %d cache items', reaped, before)
-            last_clear = time.time()
-
         try:
             compl = compl_queue.get()
             cache_key = compl.get('cache_key')
@@ -176,8 +172,6 @@ def cache_processor_thread(compl_queue):
                 cached = store(cache_key, compl)
                 log.debug('Processed: %r', cache_key)
             errors = 0
-        except queue.Empty:
-            pass
         except Exception as e:
             errors += 1
             if errors > 3:
@@ -190,6 +184,9 @@ def start_background(compl_queue):
     t = threading.Thread(target=cache_processor_thread, args=(compl_queue,))
     t.daemon = True
     t.start()
+    t = threading.Thread(target=reap_cache)
+    t.daemon = True
+    t.start()
 
 
 # balanced() taken from:
diff --git a/rplugin/python3/deoplete/sources/deoplete_jedi/worker.py b/rplugin/python3/deoplete/sources/deoplete_jedi/worker.py
index 92e63ef..f66d55b 100644
--- a/rplugin/python3/deoplete/sources/deoplete_jedi/worker.py
+++ b/rplugin/python3/deoplete/sources/deoplete_jedi/worker.py
@@ -64,11 +64,10 @@ def run(self):
             try:
                 work = self.in_queue.get()
                 self.log.debug('Got work')
-                self.out_queue.put(self.completion_work(*work), timeout=0.5)
+                self.out_queue.put(self.completion_work(*work))
                 self.log.debug('Completed work')
             except Exception:
                 self.log.debug('Worker error', exc_info=True)
-                time.sleep(0.05)
 
 
 def start(count, desc_len=0, short_types=False, show_docstring=False,
```

pytest_django/compat.py
```diff
@@ -15,3 +15,11 @@ def teardown_databases(db_cfg, verbosity):
 except ImportError:
     # Django 1.7.
     from django.db.backends import BaseDatabaseWrapper  # noqa
+
+
+def getfixturevalue(request, value):
+    try:
+        return request.getfixturevalue(value)
+    except AttributeError:
+        return request.getfuncargvalue(value)
```
blueyed: Oh, there might hide some AttributeError in getting the fixture.

Better do:
```python
try:
    f = request.getfixturevalue
except AttributeError:
    return request.getfuncargvalue(value)
return f(value)
```

mdentremont: Ah good catch!

mdentremont: If we're going down this route - might as well just use a hasattr?

mdentremont:
@blueyed - Oh man, finally realizing why this isn't working: we're importing from compat in some cases - before we try pulling in the DB fixture ---- the DB fixtures are what trigger Django setup (or at least just before it gets triggered), no?

mdentremont:
@blueyed It took awhile - but I finally got this thing to build 😛

pelme:
Thanks for the fix, reviews and the persistence to get this working! :)

PR #395, opened 2016-09-15
Fix deprecation warning

rpkilby:


pelme:
Hmm, is there any chance we could have our own test suite catch any pytest warnings and fail? We should never ship a version that gives pytest-warnings.

I have not seen this warning myself in pytest-django itself or in my own project, where does it show up?

rpkilby:
I found this while fixing deprecation warnings for DRF. Running their test suite w/ the `-Wonce` flag generates this warning as well as the one from #393.

rpkilby:
> Hmm, is there any chance we could have our own test suite catch any pytest warnings and fail? We should never ship a version that gives pytest-warnings.

I was able to view errors w/ `python -Wonce -m pytest`. You could try to force errors w/ `-Werror`, but I was getting errors from other dependencies.

blueyed:
See also https://github.com/fschulze/pytest-warnings.

pelme:
Thanks for this PR! I've opened a new issue to track catching warnings in our CI setup: #415

Issue #76, opened 2014-03-20
Support cleanup between tests with multiple databases


ftobia:
pytest-django doesn't clean up between tests when using Django with multiple databases. The problem is related to this StackOverflow question: http://stackoverflow.com/questions/10121485/django-testcase-not-using-transactions-on-secondary-database

The `db` fixture uses Django's TestCase under the covers.

Here is the code from the original `db` fixture:

```
case = TestCase(methodName='__init__')
case._pre_setup()
request.addfinalizer(case._post_teardown)
request.addfinalizer(_django_cursor_wrapper.disable)
```

Here is how I patched it as a work-around in my code:

```
case = TestCase(methodName='__init__')
case.multi_db = True
case._pre_setup()
request.addfinalizer(case._post_teardown)
request.addfinalizer(_django_cursor_wrapper.disable)
```

Obviously that multi_db flag exists and defaults to False for a reason. I'm not sure the right way to incorporate multi_db support into pytest_django, and I'm not sure the right way to test such a change.

If you have a suggestion I can work on a pull request.

pelme:
@slafs There has not been much discussions/requests for multi database support apart from this.

It would be pretty simple to pass on `multi_db=True`, but we also needs some kind of tests for this, so we also need a secondary database set up. If anyone is willing to work on this, it would be great :)

robdennis:
I've been using the work-around @ftobia mentions in the opening post, but after upgrading to Django 1.8.4 from 1.7 it stopped working.

There was Django [change](https://github.com/django/django/commit/da9fe5c717c179a9e881a40efd3bfe36a21bf4a6) to make the code handling setup/tearDown look for a ``multi_db`` class attribute instead of an instance attribute.

Changing @ftobia's work-around to the following worked for me:
```
case = TestCase(methodName='__init__')
case.__class__.multi_db = True
case._pre_setup()
request.addfinalizer(case._post_teardown)
request.addfinalizer(_django_cursor_wrapper.disable)
```

adrianandreias:
+1 :+1: 

This is stopping us from switching from Django test runner to pytest.

eremeevdev:
Why no use `django_case.multi_db = True` here https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/fixtures.py#L107 by default?

knabben:
+1

guilhermemaba:
+1

pelme:
[...]

PR #397, opened 2016-09-23
Added support for multi_db

pmourlanne:
This adds support a marker and a fixture, that make tests behave like Django TestCase's multi_db class attribute (as defined in https://github.com/pytest-dev/pytest-django/issues/76).

I added empty tests to test the added fixture and marker. I'm not sure how to implement them. I'm willing to try if you help me with a few pointers :>

(I have added the same kind of multi_db support for pytest-django 2.9.1, I can do another PR if you are interested)

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -12,8 +12,8 @@ from .django_compat import is_django_unittest
 
 from .lazy_django import get_django_version, skip_if_no_django
 
-__all__ = ['django_db_setup', 'db', 'transactional_db', 'admin_user',
-           'django_user_model', 'django_username_field',
+__all__ = ['django_db_setup', 'db', 'transactional_db', 'multi_db',
+           'admin_user', 'django_user_model', 'django_username_field',
            'client', 'admin_client', 'rf', 'settings', 'live_server',
            '_live_server_helper']
 
@@ -107,7 +107,8 @@ def django_db_setup(
         request.addfinalizer(teardown_database)
 
 
-def _django_db_fixture_helper(transactional, request, django_db_blocker):
+def _django_db_fixture_helper(transactional, request,
+                              django_db_blocker, multi_db=False):
     if is_django_unittest(request):
         return
 
@@ -124,6 +125,7 @@ def _django_db_fixture_helper(transactional, request, django_db_blocker):
         from django.test import TestCase as django_case
 
     test_case = django_case(methodName='__init__')
+    test_case.multi_db = multi_db
     test_case._pre_setup()
     request.addfinalizer(test_case._post_teardown)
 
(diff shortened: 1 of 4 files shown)


pytest_django/fixtures.py
```diff
@@ -124,6 +125,7 @@ def _django_db_fixture_helper(transactional, request, django_db_blocker):
         from django.test import TestCase as django_case
 
     test_case = django_case(methodName='__init__')
+    test_case.multi_db = multi_db
```
blueyed: This needs to be set on the class also, not the instance.

pytest_django/fixtures.py
```diff
@@ -124,6 +125,7 @@ def _django_db_fixture_helper(transactional, request, django_db_blocker):
...
     test_case._pre_setup()
     request.addfinalizer(test_case._post_teardown)
```
blueyed: I think a more universal approach would be to get access to the Django TestCase, where you then could set `multi_db` on.  I will create a PR for that.
/cc @pelme

blueyed:
Heh, since `multi_db` is a class attribute the following (in your conftest.py) is enough after all:

```python
from django.test import TestCase
TestCase.multi_db = True
```

blueyed:
FWIW, I am the following now as a workaround until something like this and/or #431 gets in:

```python
@pytest.fixture
def multi_db():
    pass


@pytest.fixture
def db(request):
    if 'multi_db' in request.fixturenames:
        TestCase.multi_db = True
    request.getfixturevalue('db')


@pytest.fixture
def transactional_db(request):
    if 'multi_db' in request.fixturenames:
        TransactionTestCase.multi_db = True
    request.getfixturevalue('transactional_db')
```

(It could/should hook into / use `django_db_modify_db_settings` probably instead to be a bit cleaner)

pytest_django/fixtures.py
pmourlanne: What difference does it make if it's set on the instance or the class?
(It's genuine curiosity, I made the change :>)

pmourlanne:
@blueyed I made the changes you requested (I think :o)

I still don't know how to write the tests for the new marker and fixture :/

blueyed:
Don't worry.  Let's wait for some feeback from others.
With e.g. https://github.com/pytest-dev/pytest-django/pull/431 this might not be necessary after all.

pmourlanne:
Hello, I just updated this pull request :)

I had difficulties making the marker work. The solution used here, to check for the marker in `_django_db_fixture_helper` is not ideal, other solutions are welcome :)
Internal tests should be working with sqlite, I still need to work on other backends.

bluetech:
I like the approach here, unfortunately `multi_db` is now deprecated, replaced by `databases`, so we should change to use that. I'll close this PR, as it's probably better to do that in a new PR, maybe basing on this one.

Thanks @pmourlanne!

Issue #405, opened 2016-10-24
--pdb breaks django TestCase db flush/rollback


asfaltboy:
Added a failing test case in this branch:
https://github.com/asfaltboy/pytest-django/blob/feature/pdb-failure/tests/test_unittest.py#L249

blueyed:
Please submit it as a PR, so we can discuss it here.

asfaltboy:
Latest changes in pytest 3.0.2, cause this, discussion in PR #406

PR #406, opened 2016-10-24
Add a failing unittest case when using `--pdb` option

asfaltboy:
Here's a failing test for #405, tests on travis-ci and tox pass. This helped to isolate it to occurring on latest `pytest==3.0.3`, probably has something to do with the `pdb` changes in the last minor releases. Specifically, [changelog for 3.0.2](https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#302) includes merge of PR pytest-dev/pytest#1890:

> Do not call tearDown and cleanups when running tests from unittest.TestCase subclasses with --pdb enabled. This allows proper post mortem debugging for all applications which have significant logic in their tearDown machinery (#1890). Thanks @mbyt for the PR.

In essence, it's not a bad change; but, when combined with `--reuse-db` or when `setUp` create unique Django objects this fails hard.

I guess one way, would be to re-enable this cleanup for `pytest-django` only.

But, to be honest, not sure how many more devs may spend hours looking for their missing tearDown code. On the other hand, I myself have suffered from ["early cleanUp when using pdb"](https://github.com/pytest-dev/pytest-mock/pull/58#issuecomment-253556301). Hmmm, maybe there is a better way.

Other users have noted this can cause issues at pytest-dev/pytest#1932

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -387,6 +387,22 @@ def _django_setup_unittest(request, django_db_blocker):
 
         cls = request.node.cls
 
+        # implement missing (as of 1.10) debug() method for django's TestCase
+        # see pytest-dev/pytest-django#406
+        def _cleaning_debug(self):
+            testMethod = getattr(self, self._testMethodName)
+            skipped = (
+                getattr(self.__class__, "__unittest_skip__", False) or
+                getattr(testMethod, "__unittest_skip__", False))
+
+            if not skipped:
+                self._pre_setup()
+            super(cls, self).debug()
+            if not skipped:
+                self._post_teardown()
+
+        cls.debug = _cleaning_debug
+
         _restore_class_methods(cls)
         cls.setUpClass()
         _disable_class_methods(cls)
--- a/tests/test_unittest.py
+++ b/tests/test_unittest.py
@@ -235,3 +235,46 @@ class TestCaseWithTrDbFixture(TestCase):
     def test_simple(self):
         # We only want to check setup/teardown does not conflict
         assert 1
+
+
+def test_pdb_enabled(django_testdir):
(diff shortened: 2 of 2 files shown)


asfaltboy:
Tests on travis-ci and tox pass, so I managed to isolate it to occurring on latest `pytest==3.0.3`, probably has something to do with the `pdb` changes in the last minor releases.

blueyed:
Can you merge/cherry-pick https://github.com/pytest-dev/pytest-django/pull/400 into this, so it fails here, too, please?

asfaltboy:
Added a quick workaround in latest commit d5696a2 ; this is basically copy-pasta of `django.test.testcases.TestCase.__call__` sans try/except wrappers.

pytest_django/plugin.py
```diff
@@ -386,6 +386,20 @@ def _django_setup_unittest(request, django_db_blocker):
...
+                getattr(testMethod, "__unittest_skip__", False))
+
+            if not skipped:
+                self._pre_setup()
+            super(cls, self).debug()
+            if not skipped:
+                self._post_teardown()
+
+        cls.debug = _cleaning_debug
+
```
blueyed: Needs a comment/doc!

It was mentioned to report it for Django itself also.. has that been done?

asfaltboy:
Django ticketed opened, for back ref: https://code.djangoproject.com/ticket/27391

adamchainz:
@blueyed I didn't get around to opening the Django ticket as I said I would in pytest-dev/pytest#1977 ... thanks @asfaltboy !

pelme:
Thanks for all debugging! I've merged this and will soon put out a 3.1.1 release which contains this fix!

PR #407, opened 2016-10-25
_django_clear_outbox should clear mail.outbox instead of creating new list

peterlauri:
The _django_clear_outbox fixture creates a new list, not clearing the existing mailbox. This can cause problems as fixture execution order cannot be guaranteed, and there might be fixtures (in my case an mailoutbox fixture) that just contains a reference to mail.outbox to be able to check what emails was sent out by the system... however, tests fails randomly as sometimes my own fixture actually gets reference to an old mailbox, not the newly created one...


pelme:
Thanks! :)

PR #414, opened 2016-11-05
Run checks at the start of tests

### LINKED ISSUE #337: Automatically run checks at the start of tests
I've been trying to get Django's test runner to run `manage.py check` at the start of the test suite in django/django#6294. It actually used to do it in 1.7 and then no one noticed it disappearing in 1.8. There is already a solution in that PR but it requires a lot of secondary effort to fix all the check failures that appear in Django's own test suite.

The reason to run check is to avoid wasting time on tests when there is a setup related issue that shoudl be fixed first. Since tests don't run checks by default, TDD-only workflows (i.e. no `runserver` etc.) can miss check errors until e.g. CI runs them explicitly, which can waste effort. Also they take a trivial amount of time to run so don't slow tests down really.

I think it would probably be easier to add it to `pytest-django` before I finish my Django core PR. Also we're using `pytest-django` at work now, so it would fix it for me for the time being.

adamchainz:
Fixes #337. Checks are run at the start of tests, like Django does in 1.11+ after django/django#6294 was merged.


adamchainz:
A few things need discussing/doing:

1. Don't merge this until django/django#6294 is merged 😄 
2. This currently breaks output. Pytest-django diverges from Django test runner by implementing the database as a pytest fixture that requires a database-touching test to set it up, rather than always setting it up. By running checks at this point, we get broken output:

``
tpkg/test_the_test.py::TestLiveServer::test_a System check identified no issues (0 silenced).
PASSED
```

The output is required to see the failing checks.

At YPlan our implementation has similar code to this PR in a fixture like:

```
@pytest.fixture(scope='session', autouse=True)
def django_check(request, django_db_setup, django_db_blocker):
```

`pytest-django` could make the DB `autouse=True` and then output from checks could appear before the first test run, but this would also slow down test runs that are purely non-database.

3. I have yet to add a config option to disable it, a changelog note, or tests with xdist(?).

pelme:
Very interesting feature! I often forget to run the Django checks myself...

pytest-django is designed to be usable where not the full test suite depends on Django, and being able to run a subset of tests and construct Django dependencies (DB) when needed. autouse=True for db would not really be feasible without breaking that and I think we should try to find another way around to make this usable. :)

We could look at pytest-pep8 or pytest-flakes for inspiration. We could use a command line option like `--django-checks` that would run checks produce a django check section and warn or fail the test run.

adamchainz:
Updated now I've updated my Django PR. I still haven't added a config option to disable it or a changelog note, but it's working the same way Django 1.11 will.

blueyed:
@adamchainz 
There are CI failures.. do you have an update?

adamchainz:
No, rebased on latest master and pushed because Travis had lost the logs. Will look at this agian, thanks for nudge

MRigal:
Hi @adamchainz would be great to have this in!

blueyed:
It should likely use some more specific pytest hook, e.g. `pytest_configure`.

In general I think that running this separately in CI is better than with (possibly) every pytest invocation.

adamchainz:
> In general I think that running this separately in CI is better than with (possibly) every pytest invocation.

Checks take milliseconds and can save a lot of painful debugging because they have clearer errors around misconfiguration. If they aren't run in tests, they can be missed in TDD workflows, since most users would only see them on runserver and might work on a change without actually running the server. I found this repeatedly in my job at the time and that's why I pushed to make it officially done nat start of tests in Django 1.11 (IIRC it was done accidentally in Django up until 1.9 but then a refactor dropped it in 1.10).

blueyed:
Ok.  I'm ok with adding it, but there should be an option for it, defaulting to off in my opinion - but I would be happy enough if I could disable it.

Like you said already it probably needs to integrate differently with pytest.

MRigal:
I still find it a very useful feature, defaulting to on in my opinion.

@adamchainz have you tried to make it a discoverable test? It could just run always as the first of the tests and add to the failures and/or stop when used with `-x` or `--maxfail`, a bit like https://pytest-ordering.readthedocs.io/en/develop/ works.
However I'm not sure how to have it also run when combining with `-k` or only on a selected file...

blueyed:
I think `pytest_collection_modifyitems` could always add an item.
But I'm not sure it should be run when `-k` or args in general are used then.

PR #426, opened 2016-11-21
in django >= 1.9 one can set module value to None in settings.MIGRATION_MODULES to not run

### LINKED ISSUE #417: Tests are failing for Django master
I'd like to make a release but we should fix this first.

Output from the failing test:
https://travis-ci.org/pytest-dev/pytest-django/jobs/173636497

peterlauri:
Closes #417

### DIFF
--- a/pytest_django/migrations.py
+++ b/pytest_django/migrations.py
@@ -1,8 +1,17 @@
 # code snippet copied from https://gist.github.com/NotSqrt/5f3c76cd15e40ef62d09
+from pytest_django.lazy_django import get_django_version
+
+
 class DisableMigrations(object):
+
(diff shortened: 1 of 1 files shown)


blueyed:
Awesome!
Is this documented?

peterlauri:
Do you mean documented in Django? If so, yes:

https://github.com/django/django/blob/master/django/db/migrations/questioner.py#L41

pelme:
Thanks! This looks good to me. The migration skipping has always been a hack and this looks less so to me. :)

peterlauri:
With exception to Django 1.8 and below that stays the same. Thanks for merging it!

PR #427, opened 2016-11-21
Make sure mail.outbox == foo and mail.outbox != bar raises exceptions.

pelme:
Refs #410.

@peterlauri please have a look!


pelme:
I use assert `mail.outbox == []` and it was not picked up by the error message, this should fix it. :)

peterlauri:
Looks good...

PR #431, opened 2016-11-21
Expose fixtures to change Django's {Transaction,}TestCase

blueyed:
This adds `django_db_testcase` and `django_transactional_db_testcase`,
which allows to override them to e.g. enable the `multi_db` feature:

```
@pytest.fixture
def django_db_testcase(django_db_testcase):
    django_db_testcase.multi_db = True
    return django_db_testcase
```

Ref: https://github.com/pytest-dev/pytest-django/pull/397

TODO:
  - [ ] doc
- [ ] test(s) - not sure what should be tested here?  Maybe the `multi_db` feature (which requires to setup multiple DBs though)

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -14,6 +14,7 @@ from .pytest_compat import getfixturevalue
 from .lazy_django import get_django_version, skip_if_no_django
 
 __all__ = ['django_db_setup', 'db', 'transactional_db', 'admin_user',
+           'django_db_testcase', 'django_transactional_db_testcase', 
            'django_user_model', 'django_username_field',
            'client', 'admin_client', 'rf', 'settings', 'live_server',
            '_live_server_helper']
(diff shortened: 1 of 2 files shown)


blueyed:
Not really required for `multi_db` after all, in case you want to change it globally (https://github.com/pytest-dev/pytest-django/pull/397#issuecomment-261987751).

This PR makes still sense, and is needed in case you want to use some specialized/derived class:

```python
from django.test import TestCase

class MultiDbTestCase(TestCase):
    multi_db = True

@pytest.fixture
def django_db_testcase():
    return MultiDbTestCase
```

Alphare:
Is there any way I could help with furthering the support for multiple databases?
I have a multi-tenant system with a connection for each tenant and I need to get testing quite soon.

sanjioh:
+1
Having this merged would be great!
I'm finding myself dealing with django and multiple DBs...

chris-erickson:
We'd like to support read replicas on Amazon Aurora/Postgres and this is one impediment to doing it confidently.  We have a vast test suite that would be a lot of work to rework with the Django TestCase.  I've tried various versions of ideas suggested here and none seem to solve the multi-db problem.  

However, if I write a router with 
```python
    def db_for_read(self, model, **hints):
        return "default"

    def db_for_write(self, model, **hints):
        return "default"
```
it works as one would expect.  Of course this doesn't really test multiple databases in the way one would want, but I think that's more of a system dynamics problem that I assume is accounted for (replica lag and it's ramifications).

Any clarity on the "best" way to hack this in right now would also be appreciated, I see a few different examples but none seem to do it, that I've tried.

bluetech:
In my opinion this is a bit of a cop-out, I'd like us to try and provide proper interfaces as much as we can. The major one ATM it seems is multi-db support (`database` attribute), let's work on supporting that (there are many issues tracking that). So I'll close this for now.

PR #434, opened 2016-11-23
re-introduced auto clearing of mail.outbox

### LINKED ISSUE #433: 3.1.0 not backward compatible with 3.0.0
Does this project use semver? It looks like there's a backward-incompatible change between 3.0.0 and 3.1.0 that ended up failing any tests that relied on `mail.outbox`. What versions should we be tracking to avoid backward-incompatible changes?

peterlauri:
Closes #433

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+3.1.2
+-----
+
+Bug fixes
+^^^^^^^^^
+
+* Auto clearing of ``mail.outbox`` has been re-introduced to not break
+  functionality in 3.x.x release. This means that Compatibility issues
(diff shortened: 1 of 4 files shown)


pelme:
Thank you!

Wouldn't replacing `mail.outbox` like this be a regression from the fix in #407?

peterlauri:
@pelme so true, fixed that and squashed...

pelme:
Sorry to not noticed the first time around, but is the monkeypatch in mailoutbox really necessary now? Wouldn't just `return mail.outbox` be enough?

peterlauri:
@pelme so true again, fix and squash... But now there is no code to maintain anymore... :(

pelme:
Excellent :D

PR #440, opened 2016-12-13
Fixed admin_user fixture for custom user model

gradam:
When custom user model had 'email' field set as username filed the fixture was resulting in Error `TypeError: create_superuser() got multiple values for argument 'email'`. Additional checking if `username_field` is not equal to 'email' fixed it.


blueyed:
Thanks!

But I am duplicating it for https://github.com/pytest-dev/pytest-django/pull/458/files - which has a request for a missing test already.
Maybe you can help out with one?

PR #445, opened 2017-01-09
[WIP] Travis: upgrade tox and pypy3

blueyed:


blueyed:
With pypy3.3-5.2-alpha1 it still fails, and `pip` reports it as `3.2` (and not 3.3)?!
https://travis-ci.org/pytest-dev/pytest-django/jobs/190338976

blueyed:
There would be a solution in https://github.com/travis-ci/travis-ci/issues/6727#issuecomment-279763549.

Issue #67, opened 2014-02-16
Doesn't work out of the box


ustun:
The out of the box experience with this package is very frustrating. I think I have spent the last hour trying to get this to work to no avail. This package doesn't work out of the box. 

I installed pytest-django, then running py.test, I get lots of weird errors.

```
export PYTHONPATH=`pwd`
export DJANGO_SETTINGS_MODULE=myproject.settings

py.test

ImportError: Could not import settings 'myproject.settings' (Is it on sys.path?): No module named settings
```

Tried setup.py trick to install into virtualenv. No luck. 

Let's try to run at least tests in a single app. 

```
tree mobile_notifications/tests
mobile_notifications/tests
├── __init__.py
└── tests.py

0 directories, 2 files

py.test mobile_notifications/tests
========================================================= test session starts ==========================================================
platform darwin -- Python 2.7.3 -- py-1.4.20 -- pytest-2.5.2
collected 0 items 

===========================================================  in 0.00 seconds ===========================================================
```

It didn't collect any items.

Any ideas?

bittner:
@ustun, you can fix the tests (file) collection issue by adding the following to the `pytest.ini` or `tox.ini` file of your project:
``` ini
[pytest]
python_files = tests.py test_*.py
```
See the [related py.test docs](http://docs.pytest.org/en/latest/example/pythoncollection.html#changing-naming-conventions).

`py.test` not collecting `tests.py` is still an issue today, with Django 1.10, pytest 3.0.5 and pytest-django 3.1.2. I'll place a PR to fix this in [our documentation](http://pytest-django.readthedocs.io/en/latest/tutorial.html#step-2-point-pytest-to-your-django-settings).

We should do better
---------------------
Though, it would certainly be better to fix this in a more automagical fashion. Since this project calls itself pytest-_django_ it's perfectly okay for users to expect that the test layout provided by `django-admin.py startapp` is honored by default. Hence:

- Either we change the [default app layout](https://github.com/django/django/tree/master/django/conf/app_template/) in the Django project itself,
- or we make pytest [pick up `tests.py` too](https://github.com/pytest-dev/pytest/blob/master/_pytest/python.py#L73-L75) by default. 

Both options don't seem to be appealing, because they are outside the scope of this project.

- Can we override the default value of pytest's `python_files` option via the plugin mechanism?

There must be a way to make `tests.py` being picked up without instructing users how to configure pytest.

bittner:
[...]

PR #446, opened 2017-01-11
[Docs] Add test discovery config instruction to tutorial chapter

bittner:
- Add test discovery config instruction to tutorial chapter
- Use code-block directive for code blocks

Addresses issue #67.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -18,8 +18,9 @@ How can I make sure that all my tests run with a specific locale?
 
 Create a `pytest fixture <http://pytest.org/latest/fixture.html>`_ that is
 automatically run before each test case. To run all tests with the english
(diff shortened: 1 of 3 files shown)


bittner:
Can we merge these changes to the docs even though the tests don't all pass? Only the tests for pypy3 fail, and they fail for an unrelated reason.

blueyed:
@bittner 
I'd like someone else to review/approve (and then merge it) though.
The more eyes the better.

I was looking at the pypy3 build issues already.. it's because setuptools dropped support for 3.2 and Travis uses an old pypy3 version (by default).  https://github.com/pytest-dev/pytest-django/pull/445.

pelme:
Very nice, thanks!

PR #447, opened 2017-01-17
Pimp up the docs! (use same theme as pytest)

bittner:
The current docs use the default theme, which is a bit old-fashioned. This PR makes our docs use the same Sphinx theme [as the pytest project](https://github.com/pytest-dev/pytest/blob/master/doc/en/conf.py#L108-L113).

- Use Flask theme** for docs (same as pytest project)
- Use anonymous links in changelog to avoid link name clashes

** Flask is an [Alabaster theme](https://alabaster.readthedocs.io/en/latest/) clone.

Note
-----

- The Flask theme **requires** installing [Flask-Sphinx-Themes](https://github.com/pallets/flask-sphinx-themes) via `pip install Flask-Sphinx-Themes` (I've found no place for that in the `Makefile` or `../requirements.txt`) -- _should be added!_
- `pygments_style` is overridden [despite related instruction](https://github.com/pallets/flask-sphinx-themes#pygments-style) in the theme docs to match the code block coloring of the pytest docs.
- `make html` needs to run twice in order to resolve references across the documents.
- A **logo** would be nice! pytest has it, we don't.  :cry: :disappointed:

### DIFF
--- a/docs/_templates/sidebarintro.html
+++ b/docs/_templates/sidebarintro.html
@@ -1,0 +1,5 @@
+<h3>About pytest-django</h3>
+<p>
+  pytest-django is a plugin for <a href="http://pytest.org/">pytest</a>.
+  Better testing for your <a href="https://www.djangoproject.com/">Django</a> project!
+</p>
--- a/docs/_templates/sidebarintro.html
+++ b/docs/_templates/sidebarintro.html
@@ -1,4 +1,4 @@
-<h3>About pytest-django</h3>
+<h3>About</h3>
 <p>
   pytest-django is a plugin for <a href="http://pytest.org/">pytest</a>.
   Better testing for your <a href="https://www.djangoproject.com/">Django</a> project!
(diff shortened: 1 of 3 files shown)


pelme:
Very nice, this is a very welcome PR that will surely make the docs much more attractive! :)

blueyed:
Thanks a lot!

bittner:
The [last build of the docs](https://readthedocs.org/projects/pytest-django/) on ReadTheDocs has failed, unfortunately. I don't know how to look at the build details (probably due to lacking permissions). Maybe the missing template is the culprit.

As mentioned in the "Notes" above, the `Flask-Sphinx-Themes` package needs to be installed using `pip`, and maybe the build even needs to be triggered twice.

blueyed:
https://readthedocs.org/projects/pytest-django/builds/4914835/
It is missing the flask theme.

Sorry, I have missed that from the notes.

blueyed:
I have found this for pytest: https://github.com/pytest-dev/pytest/commit/e14e966d

PR #456, opened 2017-02-01
Travis: re-add pypy3 to builds

blueyed:


blueyed:
There we go, still failing though.

PR #458, opened 2017-02-05
Username field can be email

pashinin:
Username field can be changed to another field. And if it is `email` then:
```
TypeError: create_superuser() got multiple values for argument 'email'
```


pelme:
Thanks for the PR! Could you please add a test for this change too?

pashinin:
Not sure where to put all this. A custom user should be defined and specified in settings.py (and user filed name changed). Then any test using `admin_client` will do the job.

blueyed:
@pashinin 
There was previously no test failing without this patch, so there needs to be a new one covering this also.
Try putting a `pdb.set_trace()` just before and see what tests are reaching this already, and then go from there.

blueyed:
Followed up by #484?!

PR #231, opened 2015-04-10
Added a new option  to avoid forcing of DEBUG setting to False

bubenkoff:
closes #228

Issue #228, opened 2015-04-09
why DEBUG is hardcoded to False?


bubenkoff:
Hi

https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/plugin.py#L239

this looks not too flexible
I tried a lot of things before i found this hardcode - i needed to understand why my liveserver fails, and it returned just standard 500 instead of debug page, and debug is set to True in my test settings
So i think this hardcode should be removed to respect test settings

blueyed:
IIRC the setup with Django's testrunner is also False, to reflect what would be used in production, but I am not certain.

:+1: for a way to configure/override this.

bubenkoff:
ok i'll prepare PR

birdsarah:
Is this fixed #231 was reverted. This option would be really useful for me.

blueyed:
@birdsarah 
Thanks for bringing this up.
You could pickup #231 and adjust it according to the feedback (double negation IIRC) and submit is as a new PR, since @bubenkoff appears to have forgotten about it.

bubenkoff:
@blueyed i didn't forget, but don't have time and honestly motivation ATM :)

blueyed:
@bubenkoff 
Fair enough and thanks for letting us now. That's what I've thought.. :)
Given your existing contribution it should be easy for somebody else to pick it up from there.

@birdsarah 
Do you think you can do that?

birdsarah:
yes, but i've got other contributions that i'm behind on on other projects, so it has to go to the bottom of the stack.

PR #463, opened 2017-02-20
Implement the --django-debug plugin option with tests

dulacp:
Follows the #228 issue, and the un-merged #231 PR.

I've taken into account @blueyed [comment](https://github.com/pytest-dev/pytest-django/pull/231#issuecomment-93381011), and included a comprehensive list of tests.

Cheers

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -5,6 +5,7 @@ test database and provides some useful text fixtures.
 """
 
 import contextlib
+from distutils.util import strtobool
 import inspect
 from functools import reduce
 import os
@@ -105,6 +106,13 @@ def pytest_addoption(parser):
         dest="nomigrations",
         default=False,
         help="Enable Django migrations on test setup",
+    )
+    group._addoption(
+        "--django-debug",
+        action="store",
+        dest="djangodebug",
+        default="None",
+        help="Configure the DEBUG setting. Defaults to False"
     )
     parser.addini(
         CONFIGURATION_ENV, "django-configurations class to use by pytest-django."
@@ -466,7 +474,9 @@ def django_test_environment(request):
         from django.conf import settings as dj_settings
         from django.test.utils import setup_test_environment, teardown_test_environment
 
-        dj_settings.DEBUG = False
+        if request.config.getvalue('djangodebug') != 'None':
+            dj_settings.DEBUG = bool(strtobool(request.config.getvalue('djangodebug')))
+
         setup_test_environment()
         request.addfinalizer(teardown_test_environment)
 
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -110,9 +110,9 @@ def pytest_addoption(parser):
     group._addoption(
         "--django-debug-mode",
         action="store",
-        dest="djangodebugmode",
+        dest="django_debug_mode",
         default="None",
-        help="Configure the DEBUG setting. Defaults to None."
+        help="Configure Django's DEBUG setting."
     )
     parser.addini(
         CONFIGURATION_ENV, "django-configurations class to use by pytest-django."
@@ -471,20 +471,21 @@ def django_test_environment(request):
     """
     if django_settings_is_configured():
         _setup_django()
-        import django
-        from django.conf import settings as dj_settings
         from django.test.utils import setup_test_environment, teardown_test_environment
 
-        if request.config.getvalue("djangodebugmode") != "None":
-            django_debug_mode = bool(strtobool(request.config.getvalue("djangodebugmode")))
+        setup_test_environment_kwargs = {}
+
+        django_debug_mode = request.config.getvalue("django_debug_mode")
+        if django_debug_mode != "None":
+            import django
+            debug = bool(strtobool(django_debug_mode))
             if django.VERSION >= (1, 11):
-                setup_test_environment(debug=django_debug_mode)
+                setup_test_environment_kwargs["debug"] = debug
             else:
-                dj_settings.DEBUG = django_debug_mode
-                setup_test_environment()
-        else:
-            # default setup
-            setup_test_environment()
+                from django.conf import settings as dj_settings
+                dj_settings.DEBUG = debug
+
+        setup_test_environment(**setup_test_environment_kwargs)
         request.addfinalizer(teardown_test_environment)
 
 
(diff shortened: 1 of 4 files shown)


karyon:
fwiw, starting from django 1.11 (to be released in april), django's test runner will have that option: https://docs.djangoproject.com/en/dev/ref/django-admin/#cmdoption-test-debug-mode

blueyed:
@karyon 
Nice find.

It gets passed through to `setup_test_environment` (https://github.com/django/django/blob/5a6f70b4281817656db2f36c5919036d38fcce7f/django/test/runner.py#L461), which we should do then, too (in https://github.com/pytest-dev/pytest-django/blob/b3d7fd66416250f1c333dce49685e96bb89eabc4/pytest_django/plugin.py#L340).

@dulaccc 
Can you change this PR accordingly?
It should pass it in for Django 1.11+, but change the setting (as done currently) for older versions.

dulaccc:
Of course I will update that in the afternoon!

auvipy:
is this going out of radar?

blueyed:
@dulaccc 
Can you update this please?  (we might have a release soon)

dulacp:
I fixed the PR with the Django 1.11 new `setup_test_environment` API.
And I used the naming convention `--django-debug-mode`.

Is that okay for you?

Cheers

karyon:
fyi, the tests failed :)

dulacp:
@karyon are you sure the tests are stable between builds? because I've only 3 commits above the #4145f73 commit (which is marked as green on master). And the failed tests on travis are totally non-related to my additions.

I'll investigate,
Cheers

karyon:
sorry, i'm not a maintainer of pytest-django, i just saw that the tests failed and thought i let you know because i myself often don't notice that :)

dulacp:
Oh, alright. Thanks for the heads-up @karyon! :)

dulacp:
Hey @blueyed, 

All the jobs targeting Python 2.7 are failing on Travis.
I suspect that part of the Travis stack changed during the past 15 days (because 15 days ago the commit 4145f73 was green).

Do you have other feedbacks on such an issue?

blueyed:
Tests are failing with pytest 3.3 (and the other jobs are using an older version (cache likely?!))
See #550.

blueyed:
Please rebase, since our tests in master were failing with pytest 3.3.

dulacp:
@blueyed nice catch for the pytest version and the caching issue!
I guess everything is green now 🍻

dulacp:
@blueyed can you merge this before master diverge?
Cheers

pytest_django/plugin.py
```diff
@@ -66,6 +67,9 @@ def pytest_addoption(parser):
     group._addoption('--migrations',
                      action='store_false', dest='nomigrations', default=False,
                      help='Enable Django migrations on test setup')
+    group._addoption('--django-debug-mode',
+                     action='store', dest='djangodebugmode', default='False',
+                     help='Configure the DEBUG setting. Defaults to False')
```
blueyed: Shouldn't it default to `None`?
Also punctuation at the end?!

pytest_django/plugin.py
```diff
@@ -332,11 +336,21 @@ def django_test_environment(request):
...
+        from distutils.version import StrictVersion
+        import django
         from django.conf import settings as dj_settings
         from django.test.utils import (setup_test_environment,
                                        teardown_test_environment)
-        dj_settings.DEBUG = False
-        setup_test_environment()
+        if request.config.getvalue('djangodebugmode') != 'None':
+            django_debug_mode = bool(strtobool(request.config.getvalue('djangodebugmode')))
+            if StrictVersion(django.get_version()) >= StrictVersion('1.11'):
```
blueyed: I think this should use the same mechanism for comparing versions like we do elsewhere, i.e. just via `django.VERSION`.

samueldotj:
@dulacp  This issue took a solid 45 minutes to track it down - I guess a lot of other developers would have faced the same issue. It would be great if you could get this committed.

richardARPANET:
Just wasted 30 minutes of my life due to this. Would be great to get this merged.

blueyed:
@samueldotj @richardARPANET 
What exactly took you time to track it down?  IIRC this is about adding a new option to control the DEBUG setting only, isn't it?

Also it clearly cannot be "just merged" given 4 conflicting files.

richardARPANET:
@blueyed My application was configured by env vars. One for the DEBUG setting.

```DEBUG = os.environ['DEBUG'].lower() == 'true'```

It worked at run-time, but not at test run-time. Which let to me trying/checking all sorts of things, I didn't expect pytest-django to be overriding this setting at all.

**Comment by blueyed** [comment]
I see.
FWIW this can also be controlled through settings, i.e.

```python
@pytest.fixture(scope="session", autouse=True)
def dj_debug(settings):
    settings.DEBUG = True
```

(or per test)

Django itself has `--debug-mode`, used by `setup_test_environment` (where we should/could pass this then also).

blueyed:
@dulacp 
I'll pick it up later I guess - please let me know if you want / can do it instead.

@auvipy 
These kind of (obvious) "reviews" are not useful.


With regard to debugging you might be interested in https://github.com/pytest-dev/pytest-django/pull/725 - I'd appreciate a review of it in return.

pytest_django/plugin.py
dulacp: Done.

pytest_django/plugin.py
dulacp: Done.

blueyed:
@dulacp 
Thanks!

I've pushed a fixup commit, let me know if this looks good to you, and we can finally (squash-)merge this then.

dulacp:
Thanks for the fixup commit! It looks great :)

blueyed:
Great.  Just wondered about the name - it mentions `--django-debug` initially, but is `django-debug-option`.  I'd prefer the shorter one, what do you think?
(sorry if this was changed / discussed before, just a quick thought)

blueyed:
I think we should use `--django-debug` and `--django-no-debug` flags here.

dulacp:
I've made the changes to use `django-debug` now.

Cheers

Andrew-Chen-Wang:
@blueyed Should the master branch be merged to incorporate some new tests and pytest 5.4.0 or is this good to go?

bluetech:
Oops, I was not aware of this PR and added a similar solution in #888. Only I used an ini option and called it `django_debug_mode`. Since that one is already merged, I'll close this one. Sorry for missing it, but let me know if you have any comments.

PR #470, opened 2017-03-17
Improve --fail-on-template-vars: use origin if available

blueyed:
This will be there with Django 1.9+ always, and in case of
`settings.TEMPLATE_DEBUG` before.  It also not go up to the Template,
but uses the nearest location (which is required with extending
templates).

Using `django.template.base.Origin` also gives the benefit of having the
full/absolute path.

btw: I suggested to not throw it with the `default` filter in Django in https://github.com/django/django/pull/8200 - we might add it in pytest-django already / instead (if it gets not accepted in Django).


blueyed:
Thanks for the review, squashed into two commits now.

PR #476, opened 2017-04-25
Do not tear down database when run with --reuse-db. Fixes #411

### LINKED ISSUE #411: --reuse-db and --create-db not working together
I noticed the below issue when using the workflow presented here: https://pytest-django.readthedocs.io/en/latest/database.html#example-work-flow-with-reuse-db-and-create-db. It seems like the behaviour changed in a recent version although it's possible I never noticed this occurred.

When I run the tests passing the `--create-db` flag, the database is recreated and the migrations are run (notice the time). The database, however, is dropped despite the `--reuse-db` flag being passed via `addopts`.

```
$ pytest tests/app --create-db
...............................................................................................................................
127 passed in 8.23 seconds
```

When I next run the tests without the `--create-db` flag the database is again recreated and the migrations are run because it was previously dropped.

```
$ pytest tests/app
...............................................................................................................................
127 passed in 7.01 seconds
```

If I run the tests a third time exactly as above, the database is properly reused:

```
$ pytest tests/app
...............................................................................................................................
127 passed in 1.12 seconds
```

dekkers:
This will only teardown the database when --reuse-db is not used. It restores the behaviour from before the database fixture refactor and will make it possible to reuse a database with initial test data.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -106,7 +106,7 @@ def django_db_setup(
                 verbosity=pytest.config.option.verbose,
             )
 
-    if not django_db_keepdb:
+    if not request.config.getvalue('reuse_db'):
         request.addfinalizer(teardown_database)
 
 


blueyed:
@dekkers 
Can you add a test for this, please?
Let us know if you need help with that.

blueyed:
Closing in favor of #560.

PR #484, opened 2017-05-28
Allow to use email as USERNAME_FIELD. Tested.

gradam:
Allow you to use email as username_field.
Also fixed a bug in an admin_client fixture that was caused by using username=admin_user.username instead of admin_user.get_username().


JakubSemikSTX:
When do you expect to merge it with the master? As I am facing the same problem and for now I have to just use a forked version from @gradam.

gradam:
Do you need anything else for this to be merged?

PR #487, opened 2017-05-30
Fix broken links

azalea:
Fix broken links of pytest documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -43,11 +43,11 @@ Why would I use this instead of Django's `manage.py test` command?
 Running your test suite with pytest-django allows you to tap into the features
 that are already present in pytest. Here are some advantages:
 
-* `Manage test dependencies with pytest fixtures. <http://pytest.org/latest/fixture.html>`_
+* `Manage test dependencies with pytest fixtures. <http://pytest.org/en/latest/fixture.html>`_
 * Less boilerplate tests: no need to import unittest, create a subclass with methods. Write tests as regular functions.
 * Database re-use: no need to re-create the test database for every test run.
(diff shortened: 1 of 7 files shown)


README.rst
```diff
@@ -43,11 +43,11 @@ Why would I use this instead of Django's `manage.py test` command?
 Running your test suite with pytest-django allows you to tap into the features
 that are already present in pytest. Here are some advantages:
 
-* `Manage test dependencies with pytest fixtures. <http://pytest.org/latest/fixture.html>`_
+* `Manage test dependencies with pytest fixtures. <http://pytest.org/en/latest/fixture.html>`_
 * Less boilerplate tests: no need to import unittest, create a subclass with methods. Write tests as regular functions.
```
blueyed: Woah!
Looks more like an issue with the redirection from readthedocs (if we're using this?!): it used to select the user's language etc..  using `https` (where it redirects to) does not help though.

blueyed: Anyway, the final link would be https://docs.pytest.org/en/latest/fixture.html - so it should be used here when going with this workaround probably.

blueyed:
Seems like the redirects have been fixed.

http://pytest.org/latest/fixture.html redirects to https://pytest.org/en/latest/fixture.html now.

I think the documentation should get updated to use https directly though.
And having the "en" redirect in there already seems to be fine, too.

blueyed:
Thank you!

pinguin999:
Hi

sad story but the link: 

> https://docs.pytest.org/en/latest/xdist.html

 stopped working.

Kolja

PR #491, opened 2017-06-14
Improve documentation on database fixtures

mpasternak:
"How to populate the database" is a bit hot topic with pytest-django, especially if you look at some StackOverflow questions. My update is an attempt to explain, what's the current state of things with pytest-django and to give straightforward code examples on how to initialise the database. What was especially unclear to me (and what I wanted to fix) was that live_server fixture uses transactional_db and having this in mind, the data need to be populated as function-scope and not as module-scope as the docs currently suggest. 

I also added flask-theme for sphinx-docs and an auto builder, to make writing docs even easier. 

Thanks for taking time to review this PR.


docs/database.rst
```diff
@@ -340,27 +340,66 @@ Put this into ``conftest.py``::
...
 
-Put this in ``conftest.py``::
+Populate the test database if you use transactions or live_server.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In case you want to use transactions in your tests (you use
+`pytest.mark.django_db(transaction=True)` marker or
+:fixture:`transactional_db` fixture), you need to populate your database
+every single time a function starts, because - by the time of this writing -
+pytest-django uses ``TRUNCATE`` SQL command for transaction-enabled tests
```
pelme: Maybe remove "by the time of this writing"? If this fact ever changes, the documentation should be updated too! :)

docs/database.rst
```diff
@@ -340,27 +340,66 @@ Put this into ``conftest.py``::
...
+If you are using the :func:`pytest.mark.django_db` marker or :fixture:`db`
+fixture, you probably don't want to explictly handle transactions in your
+tests. In this case, it is sufficient to populate your database only
+once. You can put code like this in ``conftest.py``::
+
+    import pytest
+
+    from django.core.management import call_command
+
+    @pytest.fixture(scope='module')
```
pelme: Is there any reason this uses '`scope='module'`? The old example used session which I think is the proper scope for this kind of fixture.

docs/database.rst
```diff
@@ -340,27 +340,66 @@ Put this into ``conftest.py``::
...
+`pytest.mark.django_db(transaction=True)` marker or
+:fixture:`transactional_db` fixture), you need to populate your database
+every single time a function starts, because - by the time of this writing -
+pytest-django uses ``TRUNCATE`` SQL command for transaction-enabled tests
+to remove data from the database between tests.
+
+In case you use :fixture:`live_server` fixture, it enforces using
+:fixture:`transactional_db`, so you also need to populate the test database
+this way. Put this code into ``conftest.py``. Please note, that even if it
+is very similar to previous one, the scope is changed from ``module`` to
```
pelme: module->scope, see the above comment on scope.

docs/database.rst
```diff
@@ -340,27 +340,66 @@ Put this into ``conftest.py``::
...
 pytest-django fixture is used to create the test database. When
 ``call_command`` is invoked, the test database is already prepared and
 configured.
 
-Put this in ``conftest.py``::
+Populate the test database if you use transactions or live_server.
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In case you want to use transactions in your tests (you use
+`pytest.mark.django_db(transaction=True)` marker or
```
pelme: This should use the rst markup: ```:py:func:`pytest.mark.django_db` ```

docs/database.rst
mpasternak: No idea. The old example used ``module`` IIRC and yes, ``session`` seems more proper. I'll update that.

mpasternak:
Updated, please re-review.

docs/helpers.rst
```diff
@@ -175,23 +175,28 @@ The user model used by Django. This handles different versions of Django.
...
-
 This fixture will ensure the Django database is set up.  This only
 required for fixtures which want to use the database themselves.  A
-test function should normally use the :py:func:`~pytest.mark.django_db`
+test function should normally use the :py:func:`pytest.mark.django_db`
 mark to signal it needs the database.
 
+.. fixture:: transactional_db
+
 ``transactional_db``
```
blueyed: Ah, good point of moving the `.. fixture:` anchor above.
Should be done also for the other existing ones I've added inbetween, and then maybe add it for all fixtures (I only did it for new ones / when editing others).

Makefile
```diff
@@ -18,11 +18,15 @@ $(VENV)/bin/pytest: $(VENV)/bin/python requirements.txt
 	touch $@
 
 $(VENV)/bin/sphinx-build: $(VENV)/bin/pip
-	$(VENV)/bin/pip install sphinx
+	$(VENV)/bin/pip install sphinx sphinx-autobuild
+	$(VENV)/bin/pip install Flask-Sphinx-Themes
```
blueyed: This has been moved to the "docs" tox env.
I think a new env "docs-dev" would be good then, where sphinx-autobuild gets used then (sounds interesting, have to check it out myself).

blueyed: Or "live-docs" - according to the Makefile target you have added, and which could still be in here, but would use tox then also.

docs/helpers.rst
mpasternak: Hi, 

thank you for taking time to re-review. 

I don't understand "LGTM" acronym. 

Also, per your comment "good point of moving the .. fixture: anchor" and the rest of it, it is unclear to me if you need me to perform any actions on the file or not.

Makefile
mpasternak: That's another part of your comment that I don't seem to understand, do you need me to perform any updates on this code?

docs/helpers.rst
blueyed: LGTM = Looks goot to me.

You could also move the `.. fixture …` for other fixtures, but you do not have to.  Would be good though, of course.

Makefile
blueyed: You will run into a conflict when rebasing.
Handling docs has moved from the Makefile to tox.ini (and the Makefile just calls it).
Therefore the `sphinx-autobuild` dep needs to be installed there, and should become a new tox env probably (`live-docs`).
Feel free to leave this out though, of course.

Makefile
```diff
@@ -20,6 +20,9 @@ $(VENV)/bin/pytest: $(VENV)/bin/python requirements.txt
 docs:
 	tox -e docs
 
+live-docs: $(VENV)/bin/sphinx-build
+	SPHINXBUILD=../$(VENV)/bin/sphinx-build sphinx-autobuild docs docs/_build/html
+
```
blueyed: What I've meant that this should become something like:
```
live-docs:
	tox -e live-docs
￼```
But just leave it out there for now.
I might pick it up the next time working on docs (and hopefully remembering it).

File: docs/helpers.rst [hunk 8]
```diff
@@ -239,6 +248,7 @@ This fixture provides the same transactional database access as
 sequences (if your database supports it). This is only required for
 fixtures which need database access themselves. A test function should
 normally use the ``pytest.mark.django_db`` mark with ``transaction=True`` and ``reset_sequences=True``.
+>>>>>>> a8c14aafc832f2b926c848eff00795637a527d34
```
blueyed: Looks like an unresolved conflict.

Makefile
mpasternak: sphinx-autobuild starts a long-running web server process, that serves rendered documentation and automatically reloads it on changes. 

I see no advantage running it under tox, as it would probably block "clean" tox run and cause a timeout.

docs/helpers.rst
mpasternak: Updated

Makefile
mpasternak: I added it as a convenience. 

I don't see any point of delaying documentation changes (which is the whole idea of this PR) in order to discuss ``live-docs`` and sphinx-autobuild.

Please ignore changes and other non-essential stuff.

bluetech:
Thanks @mpasternak. Due to the time passed I took the liberty of adjusting and committing this myself, see 192b63493efba2e199abdd6d76afa30277949207.

PR #500, opened 2017-08-01
Allow setting the live server port.

dtomas:
As the possibility to define a fixed live server port has been added back in Django 1.11.2 (https://code.djangoproject.com/ticket/28212), I'd like to have it available in pytest-django as well.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -302,10 +302,17 @@ def live_server(request):
     addr = (request.config.getvalue('liveserver') or
             os.getenv('DJANGO_LIVE_TEST_SERVER_ADDRESS'))
 
-    if addr and django.VERSION >= (1, 11) and ':' in addr:
-        request.config.warn('D001', 'Specifying a live server port is not supported '
-                            'in Django 1.11. This will be an error in a future '
-                            'pytest-django release.')
+    if addr and ':' in addr:
+        if django.VERSION >= (1, 11):
+            ports = addr.split(':')[1]
+            if '-' in ports or ',' in ports:
+                request.config.warn('D001',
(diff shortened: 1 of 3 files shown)


blueyed:
Thanks!

Can you cover your patch with tests, please?
See https://codecov.io/gh/pytest-dev/pytest-django/compare/781daae764e4e083eee6be111c2f0ea138aa8cf3...a725d1e9e1e817a8f63419ae7e6f6b31befc91b1/diff.

dtomas:
My changes in fixtures.py should already be covered by test_fixtures.py. I'm not sure why they are shown as not covered, as most of the test code (except the one for versions 1.11.0 up to 1.11.2) seems to have been run.
I'm surprised that fixtures.py:314 onwards has not been executed while at the same time test_specified_port_range_error_message_django_111() seems to have been run and passed.
I might be misreading the codecov output though, not having worked with it before.

blueyed:
You are right, there seems to be an issue with the coverage reporting.

Please rebase it on master and force-push - let's see if that fixes it.

blueyed:
I think https://github.com/pytest-dev/pytest-django/pull/537 should have fixed this.
Re-triggering a build to get codecov report.

blueyed:
Coverage report seems to be correct now, but it shows that we do not test the non-current Django 1.11 (< 1.11.2 IIRC).
Maybe we should not support this after all, and remove the code regarding it? /cc @pelme

avacore1337:
This patch solves our problems with testing our django REST framework application with selenium. Our Angular frontend doesn't really know what port to connect to otherwise. So thank you ^^

blueyed:
Thank you!

PR #502, opened 2017-08-09
Implement --debug-mode option to set DEBUG to True prior to run tests

adrysn:
Follows https://github.com/pytest-dev/pytest-django/pull/463 since it has been quite a while after the last discussion in the PR.

I just slightly changed @dulaccc's PR. Changes are:

1. option name `--django-debug` is changed to `--debug-mode` to follow the naming in Django 1.11 (https://docs.djangoproject.com/en/dev/ref/django-admin/#cmdoption-test-debug-mode).
2. `--debug-mode` is boolean-basis, not string-basis. For example,
    - `pytest` runs tests in non-debug mode (`DEBUG=False`)
    - `pytest --debug-mode` runs tests in debug mode (`DEBUG=True`)
3. as a result of 2., tests other than testing override DEBUG to True is removed.


blueyed:
I would prefer a `--django` prefix with the option, since it comes from the pytest-django plugin.

pytest_django/plugin.py
```diff
@@ -66,9 +66,9 @@ def pytest_addoption(parser):
     group._addoption('--migrations',
                      action='store_false', dest='nomigrations', default=False,
                      help='Enable Django migrations on test setup')
-    group._addoption('--debug-mode',
-                     help='Set DEBUG to true prior to run tests')
+    group._addoption('--django-debug',
                      action='store_true', dest='debug_mode', default=False,
+                     help='Set DEBUG to true prior to run tests')
```
blueyed: nitpick (no need to change it already): s/true/True ?!

adrysn: I'll change it. Thanks.

adrysn:
It seems that original pull-request is active again (https://github.com/pytest-dev/pytest-django/pull/463), so it would be appropriate that I close this pull-request.

PR #504, opened 2017-08-21
django_client_class fixture to use custom client class

sliverc:
Rest Framework [0] and other Django extensions use their own custom django client.
Instead of overwriting all client fixtures it is desirable to be able to overwrite client class only.

This is analogue to Django SimpleTestCase [1] where client_class can be overridden as well.

[0] http://www.django-rest-framework.org/api-guide/testing/#apiclient
[1] https://docs.djangoproject.com/en/1.11/_modules/django/test/testcases/#SimpleTestCase

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -15,7 +15,7 @@ from .django_compat import is_django_unittest
 from .lazy_django import skip_if_no_django
 
 __all__ = ['django_db_setup', 'db', 'transactional_db', 'admin_user',
-           'django_user_model', 'django_username_field',
+           'django_user_model', 'django_username_field', 'django_client_class',
            'client', 'admin_client', 'rf', 'settings', 'live_server',
            '_live_server_helper', 'django_assert_num_queries']
 
@@ -175,13 +175,19 @@ def transactional_db(request, django_db_setup, django_db_blocker):
 
 
 @pytest.fixture()
-def client():
-    """A Django test client instance."""
+def django_client_class():
+    """Django test client class."""
     skip_if_no_django()
 
     from django.test.client import Client
 
-    return Client()
+    return Client
+
+
+@pytest.fixture()
+def client(django_client_class):
+    """A Django test client instance."""
+    return django_client_class()
 
 
 @pytest.fixture()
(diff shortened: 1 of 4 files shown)


pytest_django/fixtures.py
```diff
@@ -178,13 +178,19 @@ def transactional_db(request, django_db_setup, django_db_blocker):
 
 
 @pytest.fixture()
-def client():
-    """A Django test client instance."""
+def client_class():
```
blueyed: Can we name it `django_client_class`?
New fixtures should have the `django_` prefix (from/via pytest-django, not necessarily from Django).

sliverc: Renamed it.

pelme:
Why not just override the client fixture in your local project conftest.py?

```py

import pytest
from django.test.client import Client

class CustomClient(Client):
    pass

@pytest.fixture
def client():
    return CustomClient()
```

fixtures are meant to be able to be replaced on a project level. This may not be obvious and could be documented, but I think that approach is cleaner/better. Thoughts? :)

sliverc:
After this change it will still be possible to overwrite `client` fixture if desired. Besides `client` fixture there is also `admin_client`. This means if a user uses both fixtures two fixtures need to be overwritten. And in case of `admin_client` it is not just a one liner.

I could also see more use cases for client fixtures e.g. `auth_client` where a client is authenticated but is not admin. Such a fixture could then be based on `django_client_class` as well.

sliverc:
I guess it is:

```python
@pytest.fixture
def django_client_class():
  return CustomClient
```

vs.

```python
@pytest.fixture
def client():
    return CustomClient()

@pytest.fixture
def admin_client(db, admin_user):
    client = CustomClient()
    client.login(username=admin_user.username, password='password')
    return client
```

It is a small advantage in terms of lines. But if you feel like it is not worth the new fixture feel free to close the PR.

pytest_django/fixtures.py
```diff
@@ -175,13 +175,19 @@ def transactional_db(request, django_db_setup, django_db_blocker):
 
 
 @pytest.fixture()
-def client():
-    """A Django test client instance."""
+def django_client_class():
```
blueyed: Could this be session-scoped then (by default)?

sliverc:
Makes total sense. Set scope to session.

sliverc:
Closing this PR as the gain is minimal and there doesn't seem to be an interest in it.

PR #505, opened 2017-08-21
tests/Travis: test against latest pytest 3

blueyed:


pelme:
Thanks! Looks good to me! :)

PR #506, opened 2017-08-21
tests: test_invalid_template_variable: use include

blueyed:
This shows issue https://github.com/pytest-dev/pytest/issues/580, which
is fixed in pytest 3.2.1.


blueyed:
The real fix is to upgrade pytest (https://github.com/pytest-dev/pytest-django/pull/505).
Not sure if we should have a workaround for older pytest versions? (and if it is easily possible after all)

Issue #473, opened 2017-04-06
New release?


sloria:
Do the maintainers have plans to cut the next release? A few projects at my job depend on some of the new features such as `django_assert_num_queries`.

If there are any blockers to release, I'd be glad to help out.

auvipy:
1.11 based new release should be released @blueyed

blueyed:
@pelme 
What do you say?
I am not even sure if I have permissions to do a release.

ashwoods:
any news with this? Also glad to help out.

p-himik:
Another problem is that Python 3.6 is specified as supported only in the unreleased version.

pelme:
We need an updated change log. If anyone wants to prepare it, that would be very helpful and I will then do a release as soon as possible.

I've added @blueyed as a maintainer on PyPI, so that he can upload packages too!

To have more frequent releases, here are a couple of ways you can help that would be very very appreciated ❤️:
* The release process is now slightly manual (tagging, pushing tag to Github, building sdist+wheel, uploading to PyPI). This could be scriptet. I.e. just specifying those steps in a simple script that would take a version number as an argument. This would also need to be documented.

* The changelog management is slightly painful and prevents frequent releases.  Something like towncrier integrated with the above script would be very helpful.

* Help out doing releases. You would already need to be a pytest-django github maintainer or have some kind of track record in the pytest community. Let me know you PyPI username and I will add you as a maintainer. :)

@p-himik That is annoying, but just to be clear: the current release works fine with Python 3.6 (I use it myself daily).

blueyed:
Thanks, @pelme.

Good ideas, I think using [towncrier](https://github.com/hawkowl/towncrier) might make sense, but am not sure if it is really better than skimming the git log manually once per release - pytest itself is using it, but there is more going on there in general.

pelme:
It does not take too long to skim the changelog, but (speaking for myself), removing the small steps and automating releases would be nice and probably make it so that I would release more often. Ideally it would be nice to put out bugfix releases with pretty much each merged bugfix PR. It sucks to wait for bugfix releases. :)

ashwoods:
I'll work on a tox config for managing releases as soon as I am back from vacations.

MRigal:
@pelme what about a new release now? :)

jedie:
> @pelme what about a new release now? :)

Yes, the PyPi package is very old and down't support django 1.11 :(

[...]

PR #511, opened 2017-09-04
Add tox packaging helpers

ashwoods:
As mentioned on #473, this pull requests add some tox helpers for release management. 
As the setup.py uses versioning from scm, just create a tag for correct versioning.

This is for manual signed releases. An alternative would be to use travis deploy.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -14,6 +14,7 @@ envlist =
 deps =
     django-configurations==2.0
     pytest-xdist==1.15
+    {env:_PYTESTDJANGO_TOX_EXTRA_DEPS:}
 
     checkqa: flake8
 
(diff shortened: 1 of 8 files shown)


tox.ini
```diff
@@ -46,3 +46,39 @@ commands =
...
+basepython = python3.6
+skip_install = true
+deps =
+    sphinx
+    doc8
+commands =
+    doc8 docs/
+
+[testenv:readme]
+basepython = python3.5
```
blueyed: Why only 3.5 here (and below)?
Could also be 3.6?!

ashwoods: Twine 1.9.1 doesn't officially support 3.6, as well as some of the other tools. I set all the max versions with which I have worked with.

ashwoods:
@blueyed will do tonight!

blueyed:
Thanks!

PR #517, opened 2017-09-20
Admin client login

### LINKED ISSUE #513: admin_client is not checking for login success
`client.login` inside `admin_client` can return `False` in the case when there's an existing admin user with a password set to something other than `'password'`. Perhaps, `admin_client` should use `force_login` instead?

p-himik:
Fixes #513

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -227,7 +227,7 @@ def admin_client(db, admin_user):
     from django.test.client import Client
 
     client = Client()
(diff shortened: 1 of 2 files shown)


blueyed:
LGTM, but does not work with Django 1.8 apparently.

p-himik:
@blueyed Do you have any suggestions on how to better implement it to support all required versions of Django?

bluetech:
Replaced with #881. Thanks @p-himik!

PR #518, opened 2017-09-23
Django 2.0 ImportError on django.core.urlresolvers

jpic:


pytest_django/plugin.py
```diff
@@ -441,7 +441,10 @@ def _django_set_urlconf(request):
     if marker:
         skip_if_no_django()
         import django.conf
-        from django.core.urlresolvers import clear_url_caches, set_urlconf
+        try:
+            from django.core.urlresolvers import clear_url_caches, set_urlconf
+        except ImportError:
+            from django.urls import clear_url_caches, set_urlconf
```
blueyed: Shouldn't it prefer the new version?

blueyed:
Also a comment would make sense here.

blueyed:
Thanks @jpic!

blueyed:
btw: I would like to get coverage reporting in (https://github.com/pytest-dev/pytest-django/pull/520) and then make it ready for Django 2.0 in general.  Can you take a look at #520, please?

PR #521, opened 2017-09-29
Support Django 2.0

blueyed:
Includes https://github.com/pytest-dev/pytest-django/pull/526.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,13 +1,10 @@
 [tox]
 envlist =
-    - py{35,36}-pytest3-dj{master,111,110,19,18}-postgres
-    - py34-pytest3-dj{111,110}-postgres
+    - py{35,36}-pytest3-dj{20,111,110,19,18}-postgres
+    - py34-pytest3-dj{20,111,110}-postgres
     - py33-pytest3-dj18-postgres
     - py27-pytest3-dj{111,110}-{mysql_innodb,mysql_myisam,postgres}
     - py27-pytest3-dj{111,110,19,18,17}-postgres
-    - pypy3-pytest3-d18-{sqlite,sqlite_file}
-    - pypy3-pytest2-d18-sqlite_file
-    - pypy-pytest3-dj110-sqlite_file
     - py{36,py27}-checkqa
 
 [testenv]
--- a/tox.ini
+++ b/tox.ini
@@ -1,14 +1,14 @@
 [tox]
 envlist =
-    - py{35,36}-pytest3-django{master,1.11,1.10,1.9,1.8}-postgres
-    - py34-pytest3-django{1.11,1.10}-postgres
-    - py33-pytest3-django1.8-postgres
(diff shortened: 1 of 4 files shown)


blueyed:
@pelme 
Would need some input/help with the private attributes you have used in https://travis-ci.org/pytest-dev/pytest-django/jobs/284386224#L564.

blueyed:
@auvipy 
Thanks for the "review", but it was clearly not testing dj20, and is currently broken in ~3 places.
Maybe you have some experience/insight there?

pelme:
@blueyed Thanks for working on this. Looking forward to Django 2.0!

https://github.com/django/django/commit/fba4f831bc75369c975a95c9b7774e9e89f8a2f9#diff-e0a696d19bf764562a439d3080a30fda seems to be the cause here. We should look for `apps.loading` instead of `apps._lock.locked()` which is nicer. :)

This would probably do the trick (disclaimer: I have not actually run this code):
```py
def apps_is_loading():
    if hasattr(apps, 'loading'):
        return apps.loading
    return apps._lock.locked()
```

tox.ini
```diff
@@ -18,12 +15,13 @@ deps =
...
-    django1.9: Django>=1.9,<1.10
-    django1.8: Django>=1.8,<1.9
-    django1.7: Django>=1.7,<1.8
+    djmaster: https://github.com/django/django/archive/master.tar.gz
+    dj20: Django>=2.0a1,<2.1
+    dj111: Django>=1.11,<1.12
+    dj110: Django>=1.10,<1.11
+    dj19: Django>=1.9,<1.10
+    dj18: Django>=1.8,<1.9
+    dj17: Django>=1.7,<1.8
```
auvipy: django 1.7 should be removed

auvipy:
@blueyed you could allow failures for 2.0 version in travis matrix and fix the 2.x errors in a separate pr. I saw old url patterns being used in one place

michael-k:
I fixed the middleware issues: blueyed/pytest-django#2

tox.ini
michael-k: I opened a separate PR for that: pytest-dev/pytest-django#524

blueyed:
@pelme 
Thanks.  `apps.loading` will not become `False` after loading finished however - adjusted the test.

blueyed:
@auvipy 
Sure, but I want to keep this out of this PR to not mix too many things.

blueyed:
@auvipy @pelme 
Good to go?

adamchainz:
Yo can we get a PyPI release with this? I test all my projects with `pytest-django` so can't make them Django 2 ready until this is out 😁 😁

blueyed:
@adamchainz 
Sorry, I guess you have to use the master tarball then for now.
As for us using the latest pytest release seems to work with testing against dj20 however - so what is the problem you are seeing exactly?

adamchainz:
My bad, I misread a few stack traces - the problems were really in my library code :)

PR #524, opened 2017-10-09
Drop support for Django 1.7

michael-k:
As suggested in https://github.com/pytest-dev/pytest-django/pull/521#discussion_r143324799

### DIFF
--- a/pytest_django/compat.py
+++ b/pytest_django/compat.py
@@ -1,4 +1,5 @@
 # This file cannot be imported from until Django sets up
+from django.db.backends.base.base import BaseDatabaseWrapper
 
 try:
     # Django 1.11
@@ -11,9 +12,3 @@ except ImportError:
         (_DiscoverRunner(verbosity=verbosity,
                          interactive=False)
          .teardown_databases(db_cfg))
-
-try:
-    from django.db.backends.base.base import BaseDatabaseWrapper  # noqa
-except ImportError:
-    # Django 1.7.
-    from django.db.backends import BaseDatabaseWrapper  # noqa
--- a/pytest_django/compat.py
+++ b/pytest_django/compat.py
@@ -1,6 +1,4 @@
 # This file cannot be imported from until Django sets up
-from django.db.backends.base.base import BaseDatabaseWrapper
-
 try:
     # Django 1.11
     from django.test.utils import setup_databases, teardown_databases  # noqa
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,6 @@ matrix:
     - python: 2.7
       env: TOXENV=py27-pytest3-dj18-postgres
     - python: 2.7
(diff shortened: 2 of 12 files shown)


pytest_django/compat.py
```diff
@@ -1,4 +1,5 @@
 # This file cannot be imported from until Django sets up
+from django.db.backends.base.base import BaseDatabaseWrapper
```
michael-k: Is this (`pytest_django.compat`) a public API?  If not, we could remove it and adjust the import in `plugin.py`.

blueyed:
Thanks for tackling this - the diff should be covered 100% though in this case I think.

michael-k:
> the diff should be covered 100% though in this case I think.

I'll look into it this weekend or so.

blueyed:
Looking forward to it.
Let's wait for https://github.com/pytest-dev/pytest-django/pull/521 to be merged first.

blueyed:
#521 is merged, so please rebase.

michael-k:
> please rebase

done

blueyed:
close-reopen again for codecov to report back hopefully.

blueyed:
Do you feel like covering the diff for this?
https://codecov.io/gh/pytest-dev/pytest-django/compare/16097801330500399d210d423689d71e18514c78...1a8bfe20feaab4d7e5f6553c035ce0262df7fd0b/diff

michael-k:
Give me a few hours to look over it.  I'll report back as soon as I have an idea of how much effort this requires.

michael-k:
- .
    ```
    # fixtures.py
    def django_db_setup(…):
        …
        if django_db_keepdb:
            if get_django_version() >= (1, 8):  # removed in this PR
                setup_databases_args['keepdb'] = True
            else:  # removed in this PR
                …  # removed in this PR
        …
    ```

    `django_db_keepdb` is `True` if `request.config.getvalue('reuse_db') and …`, but `reuse_db`/`reuse-db` is only set in `test_db_setup.py`. I do not know why this does not cover the above code.

- The other thing that changed without coverage ist `if get_django_version() >= (1, 8) and dj_settings.TEMPLATES:` to `if dj_settings.TEMPLATES:` (2x) in `plugin.py`.  `def _fail_for_invalid_template_variable(request)` is not covered at all and `def _template_string_if_invalid_marker(request)` has only one fully covered line.  I have no idea how to cover the missing lines.

*Conclusion:*
The lines were not tested before.  The changes (effectively replacing `get_django_version() >= (1, 8)` with `True` in conditions) don't affect the codebase at all for Django 1.8+. Diff coverage is missing because tests are missing, not because this PR made anything worse.
I do not regard the missing tests as an “easy pick”, so I'd say that they are out of scope of this PR. :(

blueyed:
Thanks!

PR #526, opened 2017-10-10
tox: use shorter names with Django factors

blueyed:
Taken out of https://github.com/pytest-dev/pytest-django/pull/521.


blueyed:
Will keep it with #521.

PR #412, opened 2016-10-29
Create --querycount parameter

renanivo:
It displays a list of top N tests with most queries.

I was inspired by the `--duration` parameter of pytest and tried to make work as similar as a could.

Note that I used the report sections to store the executed queries. Even though I'm not sure it is the best way to do that, it had a nice side effect: The executed queries are displayed when a test fails.

TODO:
 - [ ] capture all connections; if there are multiple this should be reported separately (but for the single nodeid, similar to setup/call/teardown (see 4. below))
 - [ ] revisit tests to check if it is using the test DB always really
 - [x] counts appear to be wrong: I get the same with "teardown" as with "call" always (fixed in diff/patch)
 - [ ] combine counts per nodeid, but only list counts for setup, call, and teardown in parentheses with it.
 - [ ] currently you will get the captured queries in case of failures (and `-s` I assume?).  There could also be a `django_db_log_queries` mark and fixture (context manager) that would do the same, but then could be used on a single test for debugging.  This should get refactored to provide this easily.
 - [ ] rebase on master for and refactor with `_assert_num_queries`

IDEAS:
 - [ ] default to 10 for `--querycount`, i.e. do not require an argument?

PR #535, opened 2017-10-27
PR #412 with eager pytest-cov

blueyed:
See / via https://github.com/pytest-dev/pytest-django/pull/412#issuecomment-340077539.


blueyed:
Let's keep it in #412.

PR #536, opened 2017-10-27
Revisit/clean Travis/tox config

blueyed:


blueyed:
Less jobs, more coverage.  I like.

Issue #531, opened 2017-10-27
Multiple django.setup() calls break other plugins


pakal:
_setup_django() is called multiple times, in conftest setup and as a a trylast, and it leads to multiple resets of logging config (driven by django LOGGING setting).

Inbetween these calls, other plugins like pytest-capturelog try to setup their own logging handlers, but these get reset. I've monkey patched like below, but I guess _setup_django() could use a flag to prevent its multiple execution.

```
def pytest_configure():
    # we have to prevent bug where pytest calls django.setup() multiple
    # times, breaking the logging conf (and pytest-capturelog plugin)
    import django.apps
    if django.apps.apps.ready:
        django.setup = lambda: None
```

blueyed:
Thanks for the report.

It would be nice to have a failing test case for pytest-django itself as a first step.

pakal:
Here is a testcase to show the problem : https://github.com/pytest-dev/pytest-django/pull/538

I tested it against python2.7 and Django==1.11.6, sorry I couldn't use the whole test suite or tox locally since there are lots of breakages on windows (something to do with file locking on DB removal).

Basically django.setup() should be idempotent, but in the meantime a workaround can be added here (I talk with django devs about this evolution).

pakal:
For reference, the ticket on Django side : https://code.djangoproject.com/ticket/28752#ticket

blueyed:
@pakal

> Would it be OK to make django.setup() idempotent, or even expose a "is_ready" flag for easier introspection ?

There is `django.conf.settings.configured` for this I think?!

pakal:
Is it the same thing ?

It looks like "settings.configured" is to say that a source of settings has been specified, not that django.setup() has been called, but it's fuzzy for me :  https://docs.djangoproject.com/en/1.11/topics/settings/#either-configure-or-django-settings-module-is-required

blueyed:
I am not sure, but it came to my mind - IIRC pytest-django has checks for this at least..  haven't looked at the code, nor do I remember it apparently good enough, sorry.

/cc @pelme

blueyed:
Please test/review https://github.com/pytest-dev/pytest-django/pull/629.

PR #538, opened 2017-10-30
Add testcase to show erroneous multiple runs of django.setup()

pakal:
Testcase to advance on  https://github.com/pytest-dev/pytest-django/issues/531

### DIFF
--- a/tests/test_pytest_django_initialization.py
+++ b/tests/test_pytest_django_initialization.py
@@ -1,0 +1,49 @@
+import django
+import pytest
+
+
+BARE_SETTINGS = '''
+# At least one database must be configured
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.sqlite3',
+        'NAME': ':memory:'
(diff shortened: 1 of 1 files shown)


blueyed:
Rebased against master.
I think it could be fixed for Django 2.0 at least given https://code.djangoproject.com/ticket/27176 (https://github.com/django/django/pull/7682).

blueyed:
Will include this into https://github.com/pytest-dev/pytest-django/pull/629.

blueyed:
I think the test added to https://github.com/pytest-dev/pytest-django/pull/629/files#diff-b8f370066f4a8dcc417f6e09e1689a84 is enough to test this, do you agree?

blueyed:
Backed out the changes in #629, basically only fixing this issue now.

pakal:
Yes the test added above seems to make the same as mine, ensuring django.setup() is only called once

pakal:
Thanks !

PR #547, opened 2017-12-02
Add fixture for asserting maximum number of database queries

anrie:
Often I don't want to test against the exact number of queries performed by a particular piece of code but just be sure that changes to the codebase don't end up in an unnoticed excessive increase of queries.

[django-test-plus](https://github.com/revsys/django-test-plus#assertnumquerieslessthannumber---context) has a really nice helper for this kind of scenario: https://github.com/revsys/django-test-plus#assertnumquerieslessthannumber---context

This PR tries to port the same functionality to `pytest-django`.

Open issues:
* No tests so far
* Does renaming `assert_num_queries` to `assert_exact_num_queries` make sense?

Cheers,
Andreas

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -249,8 +249,13 @@ Example
 ``django_assert_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. fixture:: django_assert_num_queries
+
 This fixture allows to check for an expected number of DB queries.
-It currently only supports the default database.
+
+It wraps `django.test.utils.CaptureQueriesContext`.  A non-default DB
+connection can be passed in using the `connection` keyword argument, and it
+will yield the wrapped CaptureQueriesContext instance.
 
 
 Example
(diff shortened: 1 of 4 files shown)


blueyed:
Please rebase, since our tests in master were failing with pytest 3.3.

anrie:
@blueyed I did a rebase and the tests are now passing.

blueyed:
@lukaszb 
What do you think?

lukaszb:
I'd leave previous name - it might be less strict but it's how Django named their test method and I'm for consistency here.

Other than that I don't see anything wrong with `max` alternative.

blueyed:
I agree with @lukaszb.

@anrie 
Please change the existing fixture to keep the existing name.

anrie:
@blueyed There you go!

blueyed:
Rebased and fixed a warning reported by Sphinx.

blueyed:
Extended it a bit to return the wrapped context manager, and added support for `connection`.
Does it look fine?

anrie:
Looks very good to me!

blueyed:
Thank you, merged!

PR #548, opened 2017-12-03
Update supported versions

hugovk:
More info about `python_requires`: [setuptools](https://setuptools.readthedocs.io/en/latest/setuptools.html?highlight=python_requires#new-and-changed-setup-keywords) and [PEP 345](https://www.python.org/dev/peps/pep-0345/).

Note: Python 3.3 was dropped in https://github.com/pytest-dev/pytest-django/pull/536.


blueyed:
Pushed a fixup for `python_requires`.

blueyed:
Thank you!

PR #554, opened 2017-12-06
Document use of authenticated `client` and `django_user_model`

### LINKED ISSUE #553: Document use of authenticated `client` and `django_user_model`
From my first reading of the docs, it appeared that pytest-django provided an authenticatable admin user (`admin_client`) or an anonymous standard user (`client`) but no way to authenticate a normal user. Only after reading the source code did I discover that `client.login()` was possible (which makes sense now that I grok that it wraps Django's test client, but wasn't clear at first). 

So I started to document that, then realized the docs could use additional general cleanup in places, so am making a PR for a general helpers doc tweak.

shacker:
- Includes misc documentation cleanup
- Resolves #553

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -19,7 +19,7 @@ on what marks are and for notes on using_ them.
 .. :py:function:: pytest.mark.django_db:
 
 This is used to mark a test function as requiring the database. It
-will ensure the database is setup correctly for the test. Each test
+will ensure the database is set up correctly for the test. Each test
 will run in its own transaction which will be rolled back at the end
 of the test. This behavior is the same as Django's standard
 `django.test.TestCase`_ class.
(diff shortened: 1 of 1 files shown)


docs/helpers.rst
```diff
@@ -261,65 +285,23 @@ Example
         assert list(m.to) == ['to@example.com']
 
 
-Environment autouse fixtures
-----------------------------
+Automatic cleanup
```
shacker: I changed this title because they appear to the end developer not as fixtures but as general features of pytest-django.

docs/helpers.rst
```diff
@@ -261,65 +285,23 @@ Example
...
-``mail.outbox`` will be cleared for each pytest, to give tests a empty
-mailbox. It is however more pytestic to use the ``mailoutbox`` fixture
-to access ``mail.outbox``.
-
-
-Examples
---------
-
-
-Example with ``rf``, ``admin_user``, fixture and class-based views
```
shacker: I removed this example because it was not at all clear how it demonstrates the automatic cleanup discussed in this section. If an example is needed, hopefully we can come up with something more succinct and demonstrative.

blueyed:
Thank you!

Issue #284, opened 2015-10-10
Request factory with a preset admin user


Glueon:
Wouldn't it be reasonable to have a request factory, which sets `user` field by default on every created request? Like `admin_client` does. For 

     @fixture
     def rf_admin(admin_user):
         from django.test.client import RequestFactory

         class AdminRequestFactory(RequestFactory):
             def request(self, **request):
                r = super().request(**request)
                r.user = admin_user
                return r

         return AdminRequestFactory()

Because 90% of the time I use `rf` fixture I set `admin_user` on it.

PR #568, opened 2017-12-26
Add fixtures for user request factories, clients

### LINKED ISSUE #565: Create helper fixtures for authenticated states on RequestFactory and client
I think the below are widely enough used on Django projects tests to be part of pytest-django itself.

- [ ] Fixture to create a django user (similar to admin_user, but non-admin)
- [ ] Create a user_client (similar to admin_user, but non-admin, related #553 #554)
- [ ] Create a request factory for unauth users (for ``AttributeError: 'WSGIRequest' object has no attribute 'user'`` errors). Helps mimic user middleware fast of falling back on [AnonymousUser](https://docs.djangoproject.com/en/2.0/ref/contrib/auth/#django.contrib.auth.models.AnonymousUser). See also, [this example in Django docs](https://docs.djangoproject.com/en/2.0/topics/testing/advanced/#example)
- [ ] Request factory for normal django user
- [ ] Request factory for admin user https://github.com/pytest-dev/pytest-django/issues/284

tony:
Fixes #565

This is based on conversations in issues #553, #554, and #284

- Introduces a django_user fixture, which is a user with no additional
  priveleges. Similar to admin_user.
- A user_client, based on django_user. Similar to admin_client.
- 3 new RequestFactory based fixtures, for unauthenticated,
  authenticated, and admin users:

  - rf_unauth: relies on AnonymousUser, similar to django docs
    example.
  - rf_admin: relies on admin_user fixture
  - rf_user: relies on new django_user fixture

  In addition, these 3 fixtures differ from rf in that they:

  - Mimic AuthenticationMiddleware by add the user attribute to
    the request object
  - Mimic SessionMiddleware by adding the 'session' attribute
    to the request object. This is an in-memory session store object
    from ``django.contrib.sessions.backends.base.SessionBase``.

This fills the gaps left over by not having normal and unauthenticated
user objects, clients, and request factories (when admin ones were
available), and also covers a common case likely present in Django
projects - anonymous (not logged in), authenticated, and admin users.


tony:
e.g.  django_rf_user, django_rf_admin, django_rf_unauth?

Is it okay if i 1. rebase against master 2. squash 5fccb31 and 722feac into my commits?

blueyed:
Yes, yes, and yes, please.. :)

tony:
@blueyed Updates

- rebased against master
- renamed new fixtures use ``django_``, including in tests and documentation
- squashed in typo fixes from 5fccb31 and 722feac

PR #575, opened 2018-02-01
Use cascade when truncating

maxnordlund:
This fixes truncation of tables with foreign keys, for example when single table inheriting a model not in INSTALLED_APPS.

I've gone ahead and fixed the issue with this small change. Though it sort of depends on the internals of Django TransactionTestCase but I figured it's OK since the attribute in question is part of the [public API](https://github.com/django/django/blob/1.11.9/django/test/testcases.py#L820-L821) and it's sort of [documented](https://github.com/django/django/blob/1.11.9/django/test/testcases.py#L839-L841).

---

Regarding testing, this happens when you inherit from a model not in INSTALLED_APPS. But I'm not familiar with the tests and thus not sure how to set that up.

In package/module foo.models:
```
from django.db import models

class Foo(models.Model):
    name = models.Charfield()
```

In package/model bar.models:
```
from django.db import models
from foo.models import Foo

class Bar(Foo):
    title = models.Charfield()
```

See #254 for the original PR.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -120,7 +120,11 @@ def _django_db_fixture_helper(transactional, request, django_db_blocker):
     request.addfinalizer(django_db_blocker.restore)
 
     if transactional:
-        from django.test import TransactionTestCase as django_case
+        from django.conf import settings
+        from django.test import TransactionTestCase
+
+        class django_case(TransactionTestCase):
+            available_apps = settings.INSTALLED_APPS
     else:
         from django.test import TestCase as django_case
 


maxnordlund:
@blueyed I need some help testing this. The only way I can think of is to create a fake package and run pytest to see it fail, aka an integration test of some sort. But that seems like overkill.

Do you have any other ides, or how to implement that one?

blueyed:
@maxnordlund 
I think an integration test using testdir Is good for this, there are several other tests in this regard, so you can use them as a template.

maxnordlund:
OK, sorry for not answering sooner. I'll try to get something up this week, probably on Wednesday. Though looking at the tests it seems like I need to modify the main fixture/example in order to test this, i.e. add another Django app. But I'm not sure I like that, the alternative is going to be a bunch of setup, which isn't fun or very readable.

Is there another way save duplicating the testdir fixture?

mbarakaja:
n@blueyed @maxnordlund Is there another way of using cascading when truncating in a transactional tests? I am facing a similar issue.

```console
E   django.core.management.base.CommandError: Database test_your_app couldn't be flushed. Possible reasons:
E     * The database isn't running or isn't configured correctly.
E     * At least one of the expected database tables doesn't exist.
E     * The SQL was invalid.
E   Hint: Look at the output of 'django-admin sqlflush'. That's the SQL this command wasn't able to run.
E   The full error: cannot truncate a table referenced in a foreign key constraint
E   DETAIL:  Table "auth_user_user_permissions" references "auth_permission".
E   HINT:  Truncate table "auth_user_user_permissions" at the same time, or use TRUNCATE ... CASCADE.

```

maxnordlund:
I'm not using Python professionally anymore, so if someone wants to take this PR to the finish line, please do so. Otherwise I suggest closing this.

bluetech:
TBH, I do not really follow how the diff relates to the issue, however the solution is probably to support `available_apps` properly. So I opened #928 and I'll close this PR in favor of that.

PR #576, opened 2018-02-05
Support Python 3.7

blueyed:
Travis does not support 3.7 on the Trusty images: https://docs.travis-ci.com/user/languages/python/#development-releases-support.. :/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,11 @@ dist: trusty
 jobs:
   fast_finish: true
   include:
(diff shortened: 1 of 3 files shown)


blueyed:
Coverage reporting (`coverage xml`) fails, which I could not reproduce using Python 3.7.0b1 - Travis is using a4 still.
https://travis-ci.org/pytest-dev/pytest-django/builds/337570117
Will re-try it later.

PR #577, opened 2018-02-05
tox: use/allow for pytest 3.4

blueyed:


blueyed:
From the first run it looks like coverage decreased because of setup.py being included now?!
Merged master for a rebuild now.

PR #579, opened 2018-02-09
Add pytest as a dependency.

ptressel:
Add pytest as a dependency, in requirements.txt, so that projects using pytest-django do not need to insert pytest before pytest-django in their own requirements.

Without pytest as a dependency for pytest-django, then if a user of pytest-django doesn't have pytest in their own requirements, listed prior to pytest-django, pip and pipenv will not know that pytest is required before pytest-django.  Even if pytest is subsequently installed, if pytest-django is installed before pytest, then an attempt to run tests will fail, with this error:

E   django.core.exceptions.ImproperlyConfigured: Requested setting DEFAULT_INDEX_TABLESPACE, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.

If one attempts to specify the path to the Django settings by using the pytest option --ds, which is supplied by pytest-django, pytest says it has no such option.

This is speculation on my part, because I haven't looked at the pytest-django setup, but it implies pytest-django was unable to add its environment variables and options into pytest, since pytest was not present at the time pytest-django was installed.

Adding the dependency would be helpful to projects using pytest-django so they don't need to know they need to add pytest before pytest-django, or at all, in their own requirements.  This came up for the django-graph-api project recently, when I adapted its installation instructions for Windows, and so was repeatedly testing installation from scratch.  Other folks working on the project had not done a from-scratch installation recently, and so already had pytest installed.  As a temporary workaround, I've added pytest before pytest-django in our own Pipfile (which we use instead of requirements.txt)

Apologies, but I'm including the following explanation to avoid a potential argument that each project should "know" what it depends on, even indirectly, and include all such indirect dependencies in the proper order:  pip, pipenv, and other installers expect that each module specifies its own dependencies, so that they can construct a dependency graph, and install modules in the required order, and so that downstream projects don't need to know all the dependencies needed by the modules they use, nor what order they must be installed in.

Let me know if you'd like an issue added for this, for discussion, or if the pull request is sufficient to have a place to chat.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,4 @@ wheel
 twine
 flake8
 Flask-Sphinx-Themes
+pytest


blueyed:
It is in `setup.py`, so pip should know about it from there:
https://github.com/pytest-dev/pytest-django/blob/e0f59dbe3880a72102a9bb1c94516881d6e6f567/setup.py#L33

`requirements.txt` is only used in the Makefile, and there the `-e .` should bring in pytest already.

ptressel:
Unfortunately, this doesn't seem to happen, perhaps because we are installing with pipenv rather than pip.  I believe pipenv uses requirements.txt if no pipfile is present, and autogenerates a pipfile, but this is only for the immediate dependencies -- it does not traverse the dependency graph at that point -- it relies on other packages dependencies during installation.  The pipfile produced by pipenv from our own earlier requirements files placed pytest-django before pytest, presumably because it didn't find the dependency in our requirements -- technically, we don't have to know about that.  Perhaps pipenv doesn't look at install_requires in setup.py...?  The two -- pip and pipenv -- definitely have differences.  For one thing, the Windows py launcher is able to tell pip which version of Python to use.  That doesn't seem to work for pipenv.  In any case, that's why I mentioned that this would be a kindness to *other* projects.  Folks in our project seem determined to use pipenv.  ;-)

I'm fairly sure nothing used the pytest-django Makefile during our installation, else it would presumably have failed on Windows -- it has *ix commands.  The "." in "-e ." in particular would not be recognized on Windows.

We do have a workaround -- we manually changed the autogenerated pipfile to move pytest before pytest-django.  We can add a warning that this problem may recur if the pipfile is regenerated.  We were hoping to avoid the problem entirely.  ;-)

One thing I haven't done is to rebuild the pytest-django package with the modified requirements.txt.  I don't currently have access to a non-Windows machine.  I might be able to talk someone else into doing it...

blueyed:
> The "." in "-e ." in particular would not be recognized on Windows

Why?

> we manually changed the autogenerated pipfile to move pytest before pytest-django

How was the Pipfile autogenerated?  Just by using pipenv?  How?

Sounds like something very specific to pipenv or your environment to me.
Just imagine that there was no requirements.txt file (or under a different name).
It should really use `install_requires` I guess.

PR #586, opened 2018-03-11
Look in child dirs when discovering django project

voidus:
Hi! :wave: 
In a lot of projects I work on, the django project is in a `src` subfolder of the repository. :fol
These changes allow pytest-django to discover the project when `pytest` is run from the root folder of the repository.

While reading the code, I discovered that `py` is in maintenance mode, so I replaced it with `pathlib`, which is in the standard library.

What do you think about these changes?

PS:
I tried to run `tox` locally, but there were some errors regarding the postgres database. Running one using Docker didn't work either. I also had a little trouble understanding how the django-project-directory-discovery is tested in general, if you feel like this feature deserves a test, please give me some pointers on how to do it.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+Features
+^^^^^^^^
+* pytest-django now searches the immediate subdirectories of the current
+  working directory when trying to find the Django project.
+
 3.1.2
 -----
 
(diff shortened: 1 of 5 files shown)


voidus:
@blueyed Assuming you agree to the code in the first commit, the new functionality is contained in two lines of code, and imo this change makes the auto-discovery much more useful. If you don't like the new feature, I would still suggest the first commit to be included.

Is there anything I can do to move this discussion forward?

blueyed:
Moved the first commit to https://github.com/pytest-dev/pytest-django/pull/631.

blueyed:
https://github.com/pytest-dev/pytest-django/pull/631 has been merged.

blueyed:
btw: https://github.com/pytest-dev/pytest-django/pull/478 - it uses testpaths.

blueyed:
Closing for now.

PR #592, opened 2018-05-04
Add new fixture `q` of type QueryDict

bronsen:
I frequently use this fixture in almost all django/drf projects when I test against APIs. Instead of implementing it again and again, I thought it could be added to pytest-django.

`q` is a [mutable QueryDict](https://docs.djangoproject.com/en/2.0/ref/request-response/#querydict-objects). Its major benefit is [urlencode()](https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.QueryDict.urlencode) which I use for constructing query parameters in URLs.

This PR includes documentation in docs/helpers.rst :)

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -285,6 +285,23 @@ Example
         assert list(m.to) == ['to@example.com']
 
 
+``q``
+~~~~~
+
+A mutable instance of ``django.http.QueryDict``.
+
+Example
+"""""""
+
+::
+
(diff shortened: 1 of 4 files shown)


bronsen:
I just came across `urllib.parse.urlencode(my_dict)` which also seems to do what I want. And `urlencode(q)` is about the same as `q.urlencode()` with regards to typing effort.

Thus I shall close this PR.

Issue #597, opened 2018-05-24
Getting error "setUpClass should be a classmethod"


kleptog:
Recently some magic was added that tries wrap the setUpClass methods for some reason. Not sure why. In any case, it breaks our testsuite. I've attached a test file to reproduce.

The basic issue is that the the setUpClass raises a skip exception sometimes to cause the methods defined in that class to be skipped. This causes the wrapping code to skip restoring the classmethod, leading to errors later on.

The error is in this code:

https://github.com/pytest-dev/pytest-django/blob/a67ab3bd0eda8a2b76b4c5954b04482b4f27e20a/pytest_django/plugin.py#L417-L419

If the setUpClass() throws an exception (in this case Skip), the restore is skipped and chaos ensues.

This is a test file that reproduces it:

```
import pytest
from django.test import SimpleTestCase


class foo(SimpleTestCase):
    @classmethod
    def setUpClass(cls):
        if cls is foo:
            raise pytest.skip("Skip base class")
        super(foo, cls).setUpClass()

    def test_shared_foo(self):
        pass


class bar(foo):
    def test_bar1(self):
        pass


class bar2(foo):
    def test_bar21(self):
        pass
```

blueyed:
Thanks for the report.

I guess it should be wrapped with a `try: … finally` maybe to call `_disable_class_methods` always?

The commit is quite old, but likely was only released recently then? (https://github.com/pytest-dev/pytest-django/commit/96f5fb1817285fbf9ef9b0de89c5190d134f845a)

It's good to have a test case.
Can you please create a PR from it?

kleptog:
Actually, after sleeping on it I think just adding a try/except isn't enough: at the point the exception is thrown the method has been set back, so this can't be the issue.

More digging revealed the actual problem: the classmethod test was incorrect. The testcase can be made even simpler: replacing the setUpClass with just pass still break.

Making a pull request now.

kleptog:
Ok, buildbot passed my patch. Actually two bugs in the end, but now it all works :)

PR #598, opened 2018-05-25
Fix classmethod test.

kleptog:
This test was incorrect in the case where the method was inherited from a
subclass.  Which ironically is precisely what the function was supposed to
test.

Fixes pytest-dev/pytest-django#597

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -277,7 +277,9 @@ def _classmethod_is_defined_at_leaf(cls, method_name):
         f = method.__func__
     except AttributeError:
         pytest.fail('%s.%s should be a classmethod' % (cls, method_name))
-    if PY2 and not (inspect.ismethod(method) and method.__self__ is cls):
+    if PY2 and not (inspect.ismethod(method) and
+                    inspect.isclass(method.__self__) and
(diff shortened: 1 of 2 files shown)


kleptog:
It's not passing the tests. Test case that this fixes:

```
from django.test import SimpleTestCase


class foo(SimpleTestCase):
    @classmethod
    def setUpClass(cls):
        super(foo, cls).setUpClass()

    def test_shared_foo(self):
        pass


class bar(foo):
    def test_bar1(self):
        pass


class bar2(foo):
    def test_bar21(self):
        pass
```

blueyed:
Thanks for fixing this (changed in a67ab3b by myself).

Tests are currently failing, but it looks like it is related to matching the output only.
You could use `*` to make it less strict - OTOH messages/output ("Destroying test database") shouldn't be joined like this in the first place.

kleptog:
Ok, the tests are now fine, but a test case fails. I'm pretty sure this indicates some other bug which I'm not sure how to fix. I'm having difficulty testing on my local setup, but I'll keep trying.

blueyed:
Thanks for fixing this - just noticed that DRF is also affected by this.

Releasing 3.3.1 now.

Issue #596, opened 2018-05-23
Pytest 3.6.0 deprecation: RemovedInPytest4Warning: MarkInfo objects are deprecated


pjdelport:
Pytest 3.6.0 (released 2018-05-23) raises the following deprecation warning with pytest-django (3.2.1):

> RemovedInPytest4Warning: MarkInfo objects are deprecated as they contain merged marks which are hard to deal with correctly.
> Please use node.get_closest_marker(name) or node.iter_markers(name).

* Release notes: https://docs.pytest.org/en/latest/changelog.html#pytest-3-6-0-2018-05-23
* Code update information: https://docs.pytest.org/en/latest/mark.html#marker-revamp-and-iteration

pjdelport:
Awesome, thanks! :)

PR #603, opened 2018-05-29
#596 MarkInfo deprecation

Code0x58:
Here is a quick go at removing the deprecation warning so tests won't break if running with `-W error`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ _build
 /.coverage
 /htmlcov/
 .cache
+.pytest_cache/
 .Python
 .eggs
 *.egg
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -378,10 +378,10 @@ def _django_db_marker(request):
     This will dynamically request the ``db`` or ``transactional_db``
     fixtures as required by the django_db marker.
     """
-    marker = request.keywords.get('django_db', None)
+    marker = request.node.get_closest_marker('django_db')
     if marker:
-        validate_django_db(marker)
-        if marker.transaction:
+        transaction = validate_django_db(marker)
+        if transaction:
             getfixturevalue(request, 'transactional_db')
         else:
             getfixturevalue(request, 'db')
@@ -447,7 +447,7 @@ def mailoutbox(monkeypatch, _dj_autoclear_mailbox):
 @pytest.fixture(autouse=True, scope='function')
 def _django_set_urlconf(request):
     """Apply the @pytest.mark.urls marker, internal to pytest-django."""
-    marker = request.keywords.get('urls', None)
+    marker = request.node.get_closest_marker('urls')
     if marker:
         skip_if_no_django()
         import django.conf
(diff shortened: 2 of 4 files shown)


pytest_django/plugin.py
```diff
@@ -378,10 +378,10 @@ def _django_db_marker(request):
     This will dynamically request the ``db`` or ``transactional_db``
     fixtures as required by the django_db marker.
     """
-    marker = request.keywords.get('django_db', None)
```
Code0x58: The two places like this bump the requirements to `pytest >= 3.6` - I think it would be easy enough to have code branch for `pytest < 3.6` to maintain existing compatibility if that feels like a blocker to this PR.

pytest_django/plugin.py
```diff
@@ -378,10 +378,10 @@ def _django_db_marker(request):
...
+    marker = request.node.get_closest_marker('django_db')
     if marker:
-        validate_django_db(marker)
```
Code0x58: I'm not 100% happy with the naming, but I didn't think of anything better.

.gitignore
```diff
@@ -11,6 +11,7 @@ _build
 /.coverage
 /htmlcov/
 .cache
+.pytest_cache/
```
Code0x58: This was missing, so added it in a separate commit.

pytest_django/plugin.py
blueyed: No, I think requiring pytest 3.6 is fine.

pytest_django/plugin.py
Code0x58: I was thinking `validate_django_db(…)` as it gets a value, and may happen to blow up, but it feels more ok to me now

blueyed:
Don't you see other errors then btw?

Tried it changing the `commands` for the testenv:
```
py36-dj20-postgres runtests: commands[0] | python -W error -m pytest --strict -x
Failed to import the site module
Traceback (most recent call last):
  File "…/Vcs/pytest-django/.tox/py36-dj20-postgres/lib/python3.6/site.py", line 703, in <module>
    main()
  File "…/Vcs/pytest-django/.tox/py36-dj20-postgres/lib/python3.6/site.py", line 683, in main
    paths_in_sys = addsitepackages(paths_in_sys)
  File "…/Vcs/pytest-django/.tox/py36-dj20-postgres/lib/python3.6/site.py", line 282, in addsitepackages
    addsitedir(sitedir, known_paths)
  File "…/Vcs/pytest-django/.tox/py36-dj20-postgres/lib/python3.6/site.py", line 204, in addsitedir
    addpackage(sitedir, name, known_paths)
  File "…/Vcs/pytest-django/.tox/py36-dj20-postgres/lib/python3.6/site.py", line 165, in addpackage
    f = open(fullname, "rU")
DeprecationWarning: 'U' mode is deprecated
```

Looks like this one is coming from Python (3.6.5) itself?!

Code0x58:
I didn't encounter that in the environment I was in. It is strange that the patch is 7 years old - it is included in at least the 3.4+ branches on GitHub, but not 2.7, so I think something is up with the environment like 2.7 files are being used.

nirizr:
Hi guys, just an FYI; This PR turned out to be breaking my builds, as it replaces the pytest required version from (the documented, still in README.md) 2.9 up to 3.6.
I wasn't sure this was intended as it not noted in either the PR or commit messages, so thought I'd note it here.

blueyed:
@nirizr 
Thanks for the notification, I will update the README.

See https://github.com/pytest-dev/pytest-django/pull/603#discussion_r191391113 - we/I've decided that it's OK to bump it, that's also why it became `3.3.0`.
It is also in the release note's title on Github: https://github.com/pytest-dev/pytest-django/releases/tag/3.3.0.

PR #615, opened 2018-06-16
Update documented pytest requirements changed in d9a9e9a8

Code0x58:
This is in response to [this comment](https://github.com/pytest-dev/pytest-django/pull/603#issuecomment-397795443) on #603.

I updated the changelog in case someone goes scanning through it later on, but it is too late for anyone who has the current version. I wouldn't have thought anyone would pin to 3.3.0, and then want to use their local `README.rst` or `docs/changelog.rst`, so I don't think there'd be any practical benifit in making something like a [`3.3.0.post1`](https://www.python.org/dev/peps/pep-0440/#post-releases)


blueyed:
Thanks!
Missed it before pushing f544f60.
Added yours in f667ef9.

PR #618, opened 2018-06-17
Fix classmethod check for mixin case

adamchainz:
I have a case of using a mixin on my test classes that started failing with 3.3.0 in upgrade PR adamchainz/django-mysql#484. It was caused by #598.

It looks like this was actually a latent failure in the way `_classmethod_is_defined_at_leaf` was picking the 'super method' - it would iterate further up the tree despite coming across the next method in the resolution chain. Also by not using `__mro__` the order of the base classes wasn't strictly being observed, although I don't think that affects my mixin case.

Added a test that fails before and passes after.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -263,7 +263,7 @@ def pytest_configure():
 def _classmethod_is_defined_at_leaf(cls, method_name):
     super_method = None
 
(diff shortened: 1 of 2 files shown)


blueyed:
Thank you!
/cc @kleptog (via #598)

adamchainz:
🍝

PR #353, opened 2016-06-24
Add support for serialized rollback

blueyed:
Fixes #329.

TODO:
- [ ] add test(s)

This conflicts (in the sense of touching the same code) with https://github.com/pytest-dev/pytest-django/pull/308.

Continuation/rebase of #331.

PR #619, opened 2018-06-25
Support reset sequences

sliverc:
Picking up work done in #308

This adds support for using the reset_sequences feature of Django's TransactionTestCase.

It will try to reset all automatic increment values before test execution, if the database supports it. This is useful when you have tests that rely on such values, like ids or other primary keys such as snapshot testing of apis with [snapshottest](https://github.com/syrusakbary/snapshottest).

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -45,7 +45,7 @@ test will fail when trying to access the database.
  values (e.g. primary keys) before running the test.  Defaults to
  ``False``.  Must be used together with ``transaction=True`` to have an
  effect.  Please be aware that not all databases support this feature.
- For details see `django.test.TransactionTestCase.reset_sequences`_
+ For details see :py:attr:`django.test.TransactionTestCase.reset_sequences`.
 
 .. note::
 
@@ -63,7 +63,6 @@ test will fail when trying to access the database.
  Test classes that subclass Python's ``unittest.TestCase`` need to have the
  marker applied in order to access the database.
 
-.. _django.test.TransactionTestCase.reset_sequences: https://docs.djangoproject.com/en/dev/topics/testing/advanced/#django.test.TransactionTestCase.reset_sequences
 .. _django.test.TestCase: https://docs.djangoproject.com/en/dev/topics/testing/overview/#testcase
 .. _django.test.TransactionTestCase: https://docs.djangoproject.com/en/dev/topics/testing/overview/#transactiontestcase
 
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -13,10 +13,10 @@ on what marks are and for notes on using_ them.
 .. _using: https://pytest.org/en/latest/example/markers.html#marking-whole-classes-or-modules
 
 
-``pytest.mark.django_db(transaction=False)`` - request database access
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``pytest.mark.django_db`` - request database access
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. :py:function:: pytest.mark.django_db:
+.. :py:function:: pytest.mark.django_db([transaction=False, reset_sequences=False]):
 
 This is used to mark a test function as requiring the database. It
 will ensure the database is set up correctly for the test. Each test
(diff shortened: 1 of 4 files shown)


blueyed:
Thanks.  I assume this would also close https://github.com/pytest-dev/pytest-django/pull/353 then?

Reminded me also of https://github.com/pytest-dev/pytest-django/pull/431 - what do you think?

blueyed:
Would be nice to get the diff covered 100%: https://codecov.io/gh/pytest-dev/pytest-django/compare/db44ccfdcc2fb710a731ca32efb6929824b50c44...db260270b074e8692cbfe51c0825d88b7b319dcf/diff

blueyed:
btw: thanks for improving the docs - I think it would make sense to only have it in one place though, and then use Sphinx' `automethod` directive etc.

sliverc:
* This PR only introduces setting of `reset_sequences` flag - to fix #353 also `serialized_rollback` would need to be set. #431 is a whole different approach. Not sure though whether exposing testcase is really pytesty though...
* Added more tests to fix the coverage issue.
* In terms of documentation. I agree `automethod` would properly be a good idea. Currently this hasn't been used anywhere in the documentation, so potential it needs some restructuring. I have a feeling this is out of scope for this PR but maybe you can give a example what you would use `automethod` for in the scope of this PR?

blueyed:
Thanks for your summary!

> #431 is a whole different approach. Not sure though whether exposing testcase is really pytesty though...

I feel like it would add the most flexibility, and could be a good base for `reset_sequences` (and `serialized_rollback` later).  I've not looked into it deeply though.

Re docs:

- `sphinx.ext.autodoc` needs to be added in `docs/conf.py` to `extendsions`
- I've looked into this when noticing that the `live_server` documentation is duplicated in `docs/helpers.rst` and the docstring, there this would be used:

    .. fixture:: live_server
    .. autofunction:: pytest_django.fixtures.live_server

I agree that it might be out of scope, but wanted to mention it since you are documenting the fixture in two places, but a single one would suffice: https://github.com/pytest-dev/pytest-django/pull/619/files#diff-352a76c86a6f0e0377970c01375e1e39R231 and https://github.com/pytest-dev/pytest-django/pull/619/files#diff-b8d58f3cc3175f8ec7bc9441f7d0bd32R191.

docs/helpers.rst
```diff
@@ -38,14 +38,23 @@ when trying to access the database.
...
  `django.test.TransactionTestCase`_
 
+
+:type reset_sequences: bool
+:param reset_sequences:
+ The ``reset_sequences`` argument will ask to reset auto increment sequence
+ values (e.g. primary keys) before running the test.  Defaults to
+ ``False``.  Must be used together with ``transaction=True`` to have an
+ effect.  Please be aware that not all databases support this feature.
+ For details see `django.test.TransactionTestCase.reset_sequences`_
```
blueyed: You can use intersphinx here:
```diff
diff --git i/docs/helpers.rst w/docs/helpers.rst
index 78e62ee..e2cfede 100644
--- i/docs/helpers.rst
+++ w/docs/helpers.rst
@@ -45,7 +45,7 @@ test will fail when trying to access the database.
  values (e.g. primary keys) before running the test.  Defaults to
  ``False``.  Must be used together with ``transaction=True`` to have an
  effect.  Please be aware that not all databases support this feature.
- For details see `django.test.TransactionTestCase.reset_sequences`_
+ For details see :py:attr:`django.test.TransactionTestCase.reset_sequences`.
 
 .. note::
 
@@ -63,7 +63,6 @@ test will fail when trying to access the database.
  Test classes that subclass Python's ``unittest.TestCase`` need to have the
  marker applied in order to access the database.
 
-.. _django.test.TransactionTestCase.reset_sequences: https://docs.djangoproject.com/en/dev/topics/testing/advanced/#django.test.TransactionTestCase.reset_sequences
 .. _django.test.TestCase: https://docs.djangoproject.com/en/dev/topics/testing/overview/#testcase
 .. _django.test.TransactionTestCase: https://docs.djangoproject.com/en/dev/topics/testing/overview/#transactiontestcase
 ```

You can use `python -msphinx.ext.intersphinx https://docs.djangoproject.com/en/dev/_objects/ | less` to see the available targets.

sliverc: Nice! Adjusted it likewise.

sliverc:
* I have played around a bit with `autodoc`. It seems that it includes the signature of the function as well and I couldn't find a way not too. If I would just use `autofunction` for `django_db_reset_sequences` it would look different to the other fixtures documented. And seeing the signature is also quite confusing as not important to user.
Maybe you know a way how to hide it? Otherwise not sure whether introducing `autofunction` just for this PR makes sense.
* concerning approach in #431 I totally see that this is more flexible than creating a single fixture for each feature. It kind of feels wrong though to expose unit test cases in pytest, but maybe thats just me :wink: The example done in the PR is to support `multi_db` support. I am not fully aware what is really needed for this to work but kind of have a feeling just the flag `multi_db` properly won't add full multi db support as changes will also be done how databases are created (I might be wrong...)
This said I would prefer if `pytest-django` only simply exposes the features it supports with fixtures instead of exposing test cases where some features might not work which can be changed.

What do you think concerning this PR - would you like to change the approach how it is done or do you see it in a merging state whereas issues in #431 and #353 can be solved in a different context?

blueyed:
I think this PR is fine for now!
As for #431 it could only be used internally also, of course and/or creating wrapper fixtures based on it.
But that could be done later still.

cb109:
@sliverc I am happy you have picked this up, thanks and cheers :hugs:

blueyed:
Rebased, planning to add it to the next release / merge it soon.

Issue #624, opened 2018-07-04
cls.setUpClass() -> TypeError: 'NoneType' object is not callable


jedie:
```
    @pytest.fixture(autouse=True, scope='class')
    def _django_setup_unittest(request, django_db_blocker):
        """Setup a django unittest, internal to pytest-django."""
        if django_settings_is_configured() and is_django_unittest(request):
            request.getfixturevalue('django_test_environment')
            request.getfixturevalue('django_db_setup')
    
            django_db_blocker.unblock()
    
            cls = request.node.cls
    
            # implement missing (as of 1.10) debug() method for django's TestCase
            # see pytest-dev/pytest-django#406
            def _cleaning_debug(self):
                testMethod = getattr(self, self._testMethodName)
                skipped = (
                    getattr(self.__class__, "__unittest_skip__", False) or
                    getattr(testMethod, "__unittest_skip__", False))
    
                if not skipped:
                    self._pre_setup()
                super(cls, self).debug()
                if not skipped:
                    self._post_teardown()
    
            cls.debug = _cleaning_debug
    
            _restore_class_methods(cls)
>           cls.setUpClass()
E           TypeError: 'NoneType' object is not callable
```
Maybe related to: https://github.com/pytest-dev/pytest-django/pull/598 ?!?

setup:
```
platform linux -- Python 3.6.5, pytest-3.6.2, py-1.5.4, pluggy-0.6.0 -- /usr/local/bin/python
plugins: django-3.3.2, cov-2.5.1, celery-4.2.0
```

Think it's related to the upgrade:
```
pytest-django==3.3.0 -> 3.3.2
pytest==3.6.1 -> 3.6.2
```

EDIT: I test the down grade to `pytest-django==3.3.0` and `pytest==3.6.1` and the test runs fine.

jedie:
I've looked into this further. I come to the simplified test code to reproduce the error:
```
from django.test import testcases
from cms.test_utils.testcases import CMSTestCase

class FooBarTestCase(testcases.TestCase):
    @classmethod
    def setUpClass(cls):
        super().setUpClass()

class TestContact(CMSTestCase, FooBarTestCase):

    def test_noop(self):
        print("OK?!?")
        return
```
I don't have time to investigate any further right now :(

blueyed:
It is caused by 8a526433a4c601a3e17ce0c90d231dd91a7e7b5b.

Used the following test:
```python
class TestUnittestMethods:
[...]

PR #625, opened 2018-07-04
Fix another unittest regression (#624)

blueyed:
Fixes https://github.com/pytest-dev/pytest-django/issues/624.


blueyed:
Thanks, @adamchainz!
Pushed your fixup.

PR #631, opened 2018-07-26
Remove py

blueyed:


pytest_django/plugin.py
```diff
@@ -120,22 +113,30 @@ def _handle_import_error(extra_message):
...
+
+        if not args:
+            args = [pathlib.Path.cwd()]
+
+        for arg in args:
+            if is_django_project(arg):
+                return arg
+            for parent in arg.parents:
+                if is_django_project(parent):
+                    return parent
```
blueyed: @voidus 
This (looking at parent) is a behavior change already, isn't it?
(at least it is not covered with tests)

blueyed: Oh, sorry.. `arg.parts(reverse=True)` handled this before

blueyed:
@voidus 
Please review.

voidus:
As far as I can see, this looks good. =)

(I just want to mention again that I couldn't get the tests to work on my machine, so don't take my word for it)

blueyed:
Thanks.

Not even with tox?
E.g. `tox -e py36-dj20-postgres` should work.
OTOH I've just noticed that the envlist was broken, maybe that caused trouble? (ab5821b)

voidus:
Ah, I probably need a database server running, right? I didn't really have the time to look into it, and I don't really want to run a shared postgres server on my workstation.

blueyed:
`tox -e py36-dj20-sqlite` should also work.. :)

PR #632, opened 2018-08-06
Remove 'database reuse' as an advantage

adamchainz:
Django can do this too with `--keepdb` since 1.8 so it seems unfair to list this as an advantage.


blueyed:
Thanks!

PR #636, opened 2018-08-18
Use pathlib2

adamchainz:
The dependency on `pathlib` was added in #631. A backport is needed for Python <3.4.

The `pathlib` package [on PyPI](https://pypi.org/project/pathlib/) is maintenance-only and comes with a message about using `pathlib2` for an up-to-date version. Also `pytest` itself uses `pathlib2` so currently I'm getting both packages installed when one would do.

This change in `setup.py` seems to be all that's needed to switch.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -11,7 +11,6 @@ import os
 import sys
 import types
 
-import pathlib2 as pathlib
 import pytest
 
 from .django_compat import is_django_unittest  # noqa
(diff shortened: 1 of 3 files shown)


adamchainz:
Woops, turns out they don't have the same library name.

graingert:
looks like pytest are going to start exposing pathlib: https://twitter.com/ossronny/status/1043438562535006209?s=19

blueyed:
@graingert 
Why is that relevant here?

graingert:
This code should change to `from pytest import pathlib` and will need the pathlib dep removed from setup.py

adamchainz:
Made #660

Issue #641, opened 2018-08-29
Windows: OSError: [WinError 123] due to pathlib changes


toolness:
My `pytest.ini` has the following content:

```ini
[pytest]
DJANGO_SETTINGS_MODULE = project.settings_pytest
addopts = --doctest-modules --cov=. --cov-report html:./coverage/python/html
norecursedirs = .git .venv .vscode node_modules
```

I believe the Unix-like path passed to `--cov-report` is causing the latest version of pytest-django to crash on Windows. (Because it's in an `.ini` file, I don't know of any platform-independent way to specify it).

Here's the traceback that displays when running `pytest`:

```pytb
Traceback (most recent call last):
  File "C:\Program Files\Python37\Lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Program Files\Python37\Lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\Atul\.virtualenvs\tenants2-Gxdtcl1v\Scripts\pytest.exe\__main__.py", line 9, in <module>
  File "c:\users\atul\.virtualenvs\tenants2-gxdtcl1v\lib\site-packages\_pytest\config\__init__.py", line 55, in main
    config = _prepareconfig(args, plugins)
  File "c:\users\atul\.virtualenvs\tenants2-gxdtcl1v\lib\site-packages\_pytest\config\__init__.py", line 180, in _prepareconfig
    pluginmanager=pluginmanager, args=args
  File "c:\users\atul\.virtualenvs\tenants2-gxdtcl1v\lib\site-packages\pluggy\hooks.py", line 258, in __call__
    return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
  File "c:\users\atul\.virtualenvs\tenants2-gxdtcl1v\lib\site-packages\pluggy\manager.py", line 67, in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
  File "c:\users\atul\.virtualenvs\tenants2-gxdtcl1v\lib\site-packages\pluggy\manager.py", line 61, in <lambda>
    firstresult=hook.spec_opts.get('firstresult'),
  File "c:\users\atul\.virtualenvs\tenants2-gxdtcl1v\lib\site-packages\pluggy\callers.py", line 196, in _multicall
    gen.send(outcome)
[...]

PR #648, opened 2018-09-16
Avoid OSError during django path detection

blueyed:
See https://github.com/pytest-dev/pytest-django/issues/641


Rakib1603065:
****Sdfgjhkl;jhgfd**fdsaawf**

Issue #290, opened 2015-11-09
Please included tests in pypi release


jlec:
Downstream developers love to test during installation and packaging. Please include tests.

pelme:
I would be happy to include tests in the distribution, but have not the motivation to work on this right now. Please send a PR and I will make sure to merge it.

blueyed:
@jlec 
This was never fixed - https://github.com/pytest-dev/pytest-django/pull/296/files included "test", but tests are in tests.
Also "pytest_django_test" would need to be included then I assume.

PR #653, opened 2018-09-16
MANIFEST.in: include tests for downstream distros

blueyed:
Ref: https://github.com/pytest-dev/pytest-django/issues/290

[ci skip]


pombredanne:
@blueyed the MANIFEST.ini is used strictly by the sdist to control what goes in the tarball or zip archive. The #665 inclusion of the LICENSE file in the setup.cfg is strictly for wheels AFAIK for now.

So that's not great but you need both for now. And unless you plan to never release sdist anymore (which I think would be mistake) you still need both.
And I am not sure how this commit helps with ##290 .

blueyed:
I've assumed that it has (good) defaults, if the file is absent.

pombredanne:
@blueyed it usually misses a few files.
One way to check is to use a mini setup.py with `from distutils.core import setup` and then run a `python setup.py sdist --manifest-only` and then see if the generated MANIFEST has all that you need.
This is the default https://github.com/python/cpython/blob/master/Lib/distutils/command/sdist.py#L210 and in Pyhon 2.7: https://github.com/python/cpython/blob/2.7/Lib/distutils/command/sdist.py

blueyed:
```diff
diff --git i/setup.py w/setup.py
index 1e551c0..b34b5ee 100755
--- i/setup.py
+++ w/setup.py
@@ -4,7 +4,8 @@
 import codecs
 import os
 
-from setuptools import setup
+# from setuptools import setup
+from distutils.core import setup
 
 
 # Utility function to read the README file.
```

The difference between master and this PR is then that master has `LICENSE` and `AUTHORS`, missing here.

I've amended this PR to readd the file and include tests explicitly.

Thanks for your help!

blueyed:
@pombredanne 
Please ACK if this makes sense.

I've tried to create a tox testenv to use the tests from the sdist, but while it is created as expected (with `tests/` on the same level as `pytest_django/`), tox would not install the tests itself - which makes sense probably, so I've skipped this - sorry, I am obviously not experienced with Python packaging.. ;)

pombredanne:
Acked-by: Philippe Ombredanne <pombredanne@nexb.com>

PR #655, opened 2018-09-16
tests: delete PYTEST_ADDOPTS in testdir already

blueyed:


blueyed:
Increased coverage (+1.59%).
Is pytest-cov being used in inner test runs messing with the outer coverage?!

PR #658, opened 2018-09-21
Back to pytest-cov

blueyed:
The problem was the eager setup, which produced an outer
`.coverage.eager.*` file, next to the `.coverage` one that pytest-cov as
of 2.6.0 generates/combines already, and then `coverage combine` would
delete that - it requires `coverage combine -a` to keep the existing
`.coverage` data.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,6 +1,6 @@
 [run]
 parallel = 1
-source = ${PYTESTDJANGO_COVERAGE_SRC}.
+source = pytest_django,pytest_django_test,tests/
 branch = 1
 
 [report]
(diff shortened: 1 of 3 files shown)


blueyed:
Looks like the compat modules are gone missing (but already before removing the eager setup): https://codecov.io/gh/pytest-dev/pytest-django/compare/5000ff814584e67eda6085f36a3644f6a834f6c2...23d1ce2953413fe3d945a5f5051faed7e74ce986/changes

blueyed:
Closing for now, since it appears to miss the compat module(s).

PR #663, opened 2018-10-15
Add parameter info to fixture assert_num_queries to display additiona…

MrtnBckr:
…l message on failure.

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -286,8 +286,10 @@ Example
 This fixture allows to check for an expected number of DB queries.
 
 It wraps `django.test.utils.CaptureQueriesContext`.  A non-default DB
-connection can be passed in using the `connection` keyword argument, and it
-will yield the wrapped CaptureQueriesContext instance.
+connection can be passed in using the `connection` keyword argument, an
+additional info message which is displayed on fail can be passed in using
+the `info` keyword argument, and it will yield the wrapped
+CaptureQueriesContext instance.
 
 
 Example
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -278,24 +278,23 @@ Example
         assert settings.USE_TZ
 
 
+.. fixture:: django_assert_num_queries
+
 ``django_assert_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. fixture:: django_assert_num_queries
+.. py:function:: django_assert_num_queries(connection=None, info=None)
+
+  :param connection: optional non-default DB connection
+  :param str info: optional info message to display on failure
 
 This fixture allows to check for an expected number of DB queries.
 
-It wraps `django.test.utils.CaptureQueriesContext`.  A non-default DB
-connection can be passed in using the `connection` keyword argument, an
-additional info message which is displayed on fail can be passed in using
-the `info` keyword argument, and it will yield the wrapped
+It wraps `django.test.utils.CaptureQueriesContext` and yields the wrapped
 CaptureQueriesContext instance.
 
 
-Example
-"""""""
-
-::
+Example usage::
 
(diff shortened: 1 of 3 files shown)


MrtnBckr:
@blueyed Yes, you could. But that way the comment won't be part of the error message supplied by giving `-v` parameter. For better understanding of the error, I like to see some hint above the queries about what was expected.

anrie:
@blueyed 

I know that it's possible to customize the error message like this:

`assert a % 2 == 0, "value was odd, should be even"`

But this doesn't seem to be possible, if you do the assertion via a context manager?
Do you see any other way to provide extra context to the error output in this case?

blueyed:
I've meant that you would see it from the traceback already if it was a comment.
I am not completely against this, but do not find it that useful, and think that it only complicates things.

> assert a % 2 == 0, "value was odd, should be even"

Would be nice if this pattern could be used, but it would also miss the original info then, i.e. the number of queries.

MrtnBckr:
Yes, you are right, that it is visible in the traceback. But not visible enough for me. I like the spot on top of the error message above the queries. And I don't think, it is very complicated. We use this parameter in our edited version of assert_num_queries in our projects and it can be a big help for developers running in this errors, who are new to the project.

blueyed:
Ok.

What do you think about this on top?

```diff
diff --git c/docs/helpers.rst i/docs/helpers.rst
index afe2f89..5e8cb17 100644
--- c/docs/helpers.rst
+++ i/docs/helpers.rst
@@ -278,24 +278,23 @@ Example
         assert settings.USE_TZ
 
 
-``django_assert_num_queries``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
 .. fixture:: django_assert_num_queries
 
+``django_assert_num_queries``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. py:function:: django_assert_num_queries(connection=None, info=None)
+
+  :param connection: optional non-default DB connection
+  :param str info: optional info message to display on failure
+
 This fixture allows to check for an expected number of DB queries.
 
-It wraps `django.test.utils.CaptureQueriesContext`.  A non-default DB
-connection can be passed in using the `connection` keyword argument, an
-additional info message which is displayed on fail can be passed in using
-the `info` keyword argument, and it will yield the wrapped
+It wraps `django.test.utils.CaptureQueriesContext` and yields the wrapped
 CaptureQueriesContext instance.
 
 
-Example
-"""""""
-
-::
+Example usage::
 
     def test_queries(django_assert_num_queries):
         with django_assert_num_queries(3) as captured:
@@ -306,20 +305,21 @@ Example
         assert 'foo' in captured.captured_queries[0]['sql']
 
 
+.. fixture:: django_assert_max_num_queries
+
 ``django_assert_max_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. fixture:: django_assert_max_num_queries
+.. py:function:: django_assert_num_queries(connection=None, info=None)
+
+  :param connection: optional non-default DB connection
+  :param str info: optional info message to display on failure
 
 This fixture allows to check for an expected maximum number of DB queries.
 
 It is a specialized version of :fixture:`django_assert_num_queries`.
 
-
-Example
-"""""""
-
-::
+Example usage::
 
     def test_max_queries(django_assert_max_num_queries):
         with django_assert_max_num_queries(3):
```

MrtnBckr:
I'm totally okay with that! Much more understandable now.

blueyed:
Added.

btw: got confused when rebasing this: you had the same commit almost the same, and resolved the conflict in a merge commit.  For the future, please try to push a fixup instead.

MrtnBckr:
Thank you. And I will push a fixup in future PRs. Sry for confusing you.

PR #664, opened 2018-10-16
Avoid crash after OSError during django path detection

### LINKED ISSUE #662: OSError on long command-line args
In the logic of django_find_project there exist these lines:
```
    def find_django_path(args):
        args = map(str, args)
        args = [arg_to_path(x) for x in args if not x.startswith("-")]
        args = [p for p in args if p.is_dir()]
...
```
Unfortunately, this means that if there are really long command-line args that this can get a `OSError: [Errno 36] File name too long: ...` which prevents me from, for example, using really long args to pytest's `-k` option.
Thanks for looking into this.

voidus:
Fixes #662

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -120,7 +120,10 @@ def _handle_import_error(extra_message):
 
 def _add_django_project_to_path(args):
     def is_django_project(path):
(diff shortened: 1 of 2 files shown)


blueyed:
Thank you!

PR #668, opened 2018-10-30
Refine the django.conf module check to see if the settings really are configured

### LINKED ISSUE #599: pytest-django fails pytest (django.core.exceptions.ImproperlyConfigured)
- I installed both pytest and pytest-django libraries in my laptop. Then I wanted to run pytest in a folder in which tests have no dependency of django.
  * Expected: I expected everything to run smoothing
  * What actually happeened? pytest failed with the error 
```
django.core.exceptions.ImproperlyConfigured:` Requested settings, but settings are not configured.
You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

- Pytest version:  3.5.1, Operating System :  Ubuntu Linux 18.04 LTS, python version - 3.6.5

I don't know whether this is an intended feature or a bug.

I reported it in pytest repo. They asked me to report it here.

olasd:
Some modules, e.g. hypothesis, load some Django modules for their own fixtures.
This means that the `django.conf` module sometimes gets loaded, even though it's
never used (and the settings are not initialized).

This causes unrelated test failures when pytest_django is installed, as
pytest_django considers that having a loaded django.conf means the settings are
set up and ready to be modified. The Django settings object provides a flag to
check this condition, which we now use.

Add a regression test that mimics what hypothesis did which makes pytest-django
fail.

Closes #599.

### DIFF
--- a/pytest_django/lazy_django.py
+++ b/pytest_django/lazy_django.py
@@ -15,16 +15,18 @@ def skip_if_no_django():
 
 
 def django_settings_is_configured():
-    # Avoid importing Django if it has not yet been imported
-    if (
-        not os.environ.get("DJANGO_SETTINGS_MODULE")
-        and "django.conf" not in sys.modules
-    ):
-        return False
+    """Return whether the Django settings module has been configured.
 
-    # If DJANGO_SETTINGS_MODULE is defined at this point, Django is assumed to
-    # always be loaded.
-    return True
+    This uses either the DJANGO_SETTINGS_MODULE environment variable, or the
+    configured flag in the Django settings object if django.conf has already
+    been imported.
+    """
+    ret = bool(os.environ.get("DJANGO_SETTINGS_MODULE"))
+
+    if not ret and "django.conf" in sys.modules:
+        return sys.modules["django.conf"].settings.configured
+
+    return ret
 
 
 def get_django_version():
--- a/tests/test_django_settings_module.py
+++ b/tests/test_django_settings_module.py
@@ -406,6 +406,41 @@ def test_no_ds_but_django_imported(testdir, monkeypatch):
     assert r.ret == 0
 
 
+def test_no_ds_but_django_conf_imported(testdir, monkeypatch):
+    """pytest-django should not bail out, if "django.conf" has been imported
+    somewhere, e.g. via hypothesis (#599)."""
+
+    monkeypatch.delenv('DJANGO_SETTINGS_MODULE')
+
+    testdir.makepyfile("""
+        import os
+        import sys
+
+        # line copied from hypothesis/extras/django.py
+        from django.conf import settings as django_settings
+
+        # Don't let pytest poke into this object, generating a
+        # django.core.exceptions.ImproperlyConfigured
+        del django_settings
+
+        from pytest_django.lazy_django import django_settings_is_configured
+
+        def test_django_settings_is_configured():
+            assert django_settings_is_configured() is False
+
+        def test_django_conf_is_imported():
+            assert 'django.conf' in sys.modules
+
+        def test_env():
+            assert 'DJANGO_SETTINGS_MODULE' not in os.environ
+
+        def test_cfg(pytestconfig):
+            assert pytestconfig.option.ds is None
+    """)
+    r = testdir.runpytest_subprocess('-s')
+    assert r.ret == 0
+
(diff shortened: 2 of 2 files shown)


pytest_django/lazy_django.py
```diff
@@ -15,14 +15,19 @@ def skip_if_no_django():
...
-    # If DJANGO_SETTINGS_MODULE is defined at this point, Django is assumed to
-    # always be loaded.
-    return True
+    """Return whether the Django settings module has been configured.
+
+    This uses either the DJANGO_SETTINGS_MODULE environment variable, or the
+    configured flag in the Django settings object if django.conf has already
+    been imported.
+
+    """
```
blueyed: In case this needs to be fixed anyway, please remove the newline at the end.

blueyed:
(Inner) test fails during collection..

This fixes it:

```diff
diff --git i/tests/test_django_settings_module.py w/tests/test_django_settings_module.py
index a7f716d..98b3f8e 100644
--- i/tests/test_django_settings_module.py
+++ w/tests/test_django_settings_module.py
@@ -368,18 +368,24 @@ def test_no_ds_but_django_conf_imported(testdir, monkeypatch):
     testdir.makepyfile("""
         import os
 
-        # line copied from hypothesis/extras/django.py
-        from django.conf import settings as django_settings
-
-        from pytest_django.lazy_django import django_settings_is_configured
-
         def test_django_settings_is_configured():
+            # Line copied from hypothesis/extras/django.py
+            from django.conf import settings as django_settings
+
+            from pytest_django.lazy_django import django_settings_is_configured
+
             assert django_settings_is_configured() is False
 
         def test_env():
+            # Line copied from hypothesis/extras/django.py
+            from django.conf import settings as django_settings
+
             assert 'DJANGO_SETTINGS_MODULE' not in os.environ
 
         def test_cfg(pytestconfig):
+            # Line copied from hypothesis/extras/django.py
+            from django.conf import settings as django_settings
+
             assert pytestconfig.option.ds is None
     """)
     r = testdir.runpytest_subprocess('-s')
```

It still fails then without your patch, but only with `test_django_settings_is_configured`.

Especially for `test_cfg` I wonder if this then still tests the initial issue, assuming that the hypothesis import happens during test collection?!

olasd:
> (Inner) test fails during collection..
>
> [...]
> 
> It still fails then without your patch, but only with `test_django_settings_is_configured`.
> 
> Especially for `test_cfg` I wonder if this then still tests the initial issue, assuming that the hypothesis import happens during test collection?!

I've modified my patch to import django.conf.settings and then `del` it so that pytest doesn't try to poke in it. I've also added an inner test to check that `django.conf` is really imported.

I'm guessing the QA check is broken because of the recent flake8 release? At least it's not in code I touched :-)

olasd:
Ah, and, obviously in the end, having that patch on pytest_django makes it not interfere with my other tests!

tests/test_django_settings_module.py
```diff
@@ -359,6 +359,41 @@ def test_cfg(pytestconfig):
...
+    testdir.makepyfile("""
+        import os
+        import sys
+
+        # line copied from hypothesis/extras/django.py
+        from django.conf import settings as django_settings
+
+        # Don't let pytest poke into this object, generating a
+        # django.core.exceptions.ImproperlyConfigured
+        del django_settings
```
blueyed: Wow.. `inspect.isclass` / `isinstance` triggers the ImproperlyConfigured already..

blueyed:
1. Created https://github.com/pytest-dev/pytest/issues/4266 for discussion about the `isclass` issue.

2. > obviously in the end, having that patch on pytest_django makes it not interfere with my other tests!

What do you mean?
That my patch did not fix the issue for you?

olasd:
>     1. Created [pytest-dev/pytest#4266](https://github.com/pytest-dev/pytest/issues/4266) for discussion about the `isclass` issue.
> 
>     2. > obviously in the end, having that patch on pytest_django makes it not interfere with my other tests!
> 
> 
> What do you mean?
> That my patch did not fix the issue for you?

Ah, no, I did my patch before I saw yours; I just meant that the original issue was indeed fixed :)

blueyed:
I see, rebased.
Will merge once it is green.
Thank you!

PR #676, opened 2018-12-05
Don't add email to extra arguments

iamanikeev:
Email could be a username field. If this is the case, we shouldn't add it into kwargs, because that would cause the same argument to be passed twice. I didn't create test case for that because this would require setting up custom user model and injecting it into a test case. Maybe there is an easier way I didn't see?

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -252,15 +252,17 @@ def admin_user(db, django_user_model, django_username_field):
     """
     UserModel = django_user_model
     username_field = django_username_field
+    username = "admin@example.com" if username_field == "email" else "admin"
 
     try:
-        user = UserModel._default_manager.get(**{username_field: "admin"})
+
+        user = UserModel._default_manager.get(**{username_field: username})
     except UserModel.DoesNotExist:
         extra_fields = {}
         if username_field not in ("username", "email"):
             extra_fields[username_field] = "admin"
         user = UserModel._default_manager.create_superuser(
-            "admin", "admin@example.com", "password", **extra_fields
+            username, "admin@example.com", "password", **extra_fields
         )
     return user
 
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks for your PR!

As you might have expected, we need a test for this.. :)

Check out `tests/test_fixtures.py::test_custom_user_model`, which fails if you comment/remove the existing code that you're changing:
```
        if username_field != "username":
            extra_fields[username_field] = "admin"
```

I think you should either parametrize that test, or adjust it to handle both cases.

iamanikeev:
> Thanks for your PR!
> 
> As you might have expected, we need a test for this.. :)
> 
> Check out `tests/test_fixtures.py::test_custom_user_model`, which fails if you comment/remove the existing code that you're changing:
> 
> ```
>         if username_field != "username":
>             extra_fields[username_field] = "admin"
> ```
> I think you should either parametrize that test, or adjust it to handle both cases.

Oh, I didn't notice that, thanks! And also, wow! It fakes the entire Django app, didn't see such technique before :) Fixture updated and test case added.

pytest_django/fixtures.py
```diff
@@ -252,15 +252,17 @@ def admin_user(db, django_user_model, django_username_field):
     """
     UserModel = django_user_model
     username_field = django_username_field
+    username = "admin@example.com" if username_field == "email" else "admin"
 
     try:
-        user = UserModel._default_manager.get(**{username_field: "admin"})
+
```
blueyed: That extra newline should be removed.

blueyed:
Yes, sometimes/often it just helps to remove some code that you're about to change and see what tests are failing.

pytest_django/fixtures.py
iamanikeev: done.

PR #677, opened 2018-12-10
Use request.config instead of pytest.config

blueyed:
Will be deprecated in pytest 4.1.
Ref: https://github.com/pytest-dev/pytest/commit/b88c3f8f828

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -103,14 +103,14 @@ def django_db_setup(
 
     with django_db_blocker.unblock():
         db_cfg = setup_databases(
-            verbosity=pytest.config.option.verbose,
+            verbosity=request.config.option.verbose,
             interactive=False,
             **setup_databases_args
         )
 
     def teardown_database():
         with django_db_blocker.unblock():
-            teardown_databases(db_cfg, verbosity=pytest.config.option.verbose)
+            teardown_databases(db_cfg, verbosity=request.config.option.verbose)
 
     if not django_db_keepdb:
         request.addfinalizer(teardown_database)


ransford:
Does this merit a pytest-django version bump?

I just added `--disable-pytest-warnings` to a project because the combination of pytest 4.1.0 + pytest-django 3.4.4 was making my tests fail like this:
```
_____________________ ERROR at setup of test_my_thing _____________________
E   _pytest.warning_types.PytestDeprecationWarning: the `pytest.config` global is deprecated.  Please use `request.config` or `pytest_configure` (if you're a pytest plugin) instead.
```
A traceback led a mention of `pytest.config` in `pytest_django/fixtures.py`.

pjdelport:
> Does this merit a pytest-django version bump?

Yes please! 🙏

blueyed:
> PytestDeprecationWarning

Why is this an error?
Are you using `--strict` or something like that?
It should only show up in "warnings summary" normally.

While a new release makes sense I'd like to understand what happens here.

nicoddemus:
Probably they have setup the warnings filter to `error`.

blueyed:
@ransford @pjdelport 
?

ransford:
Yes, sorry I forgot to mention that. I had `error` in the `filterwarnings` stanza of `pytest.ini` (on purpose).

blueyed:
Thanks for the feedback, tagged a new release.  Will take a short while until it appears on PyPI.
https://travis-ci.org/pytest-dev/pytest-django/builds/476522277

ransford:
Thanks @blueyed (for the new release and your work on this project)!

PR #679, opened 2018-12-17
Fix DB renaming fixture for Multi-DB environment

adamantike:
For Django configurations using multiple databases, there's a subtle bug
that doesn't rename databases, if there's an SQLite one that triggers
the `return` statement in the existing feature.

As the return value doesn't seem to be used, we can just move to the
next iteration in the `for`-loop, and evaluate the following DB.

A test was added to consider this scenario.

### DIFF
--- a/tests/test_db_setup.py
+++ b/tests/test_db_setup.py
@@ -215,6 +215,50 @@ class TestSqliteWithXdist:
         result.stdout.fnmatch_lines(["*PASSED*test_a*"])
 
 
+class TestSqliteWithMultipleDbsAndXdist:
+
+    db_settings = {
+        "default": {
+            "ENGINE": "django.db.backends.sqlite3",
+            "NAME": "/tmp/should-not-be-used",
+        },
+        "db2": {
+            "ENGINE": "django.db.backends.sqlite3",
+            "NAME": "db_name",
+            "TEST": {"NAME": "test_custom_db_name"},
+        }
+    }
+
+    def test_sqlite_database_renamed(self, django_testdir):
+        pytest.importorskip("xdist")
+
+        django_testdir.create_test_module(
+            """
+            import pytest
+            from django.db import connections
+
+            @pytest.mark.django_db
+            def test_a():
+                (conn_db2, conn_default) = sorted(
+                    connections.all(),
+                    key=lambda conn: conn.alias,
+                )
+
+                assert conn_default.vendor == 'sqlite'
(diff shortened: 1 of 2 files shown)


adamantike:
Ping @pelme because of #362.

tests/test_db_setup.py
```diff
@@ -213,6 +213,49 @@ def test_a():
...
+            """
+            import pytest
+            from django.db import connections
+
+            @pytest.mark.django_db
+            def test_a():
+                (conn1, conn2) = connections.all()
+
+                if conn1.alias == 'db2':
+                    conn1, conn2 = conn2, conn1
```
blueyed: Is this switching done for predictable order?

tests/test_db_setup.py
```diff
@@ -213,6 +213,49 @@ def test_a():
...
+
+                assert conn1.vendor == 'sqlite'
+                db_name = conn1.creation._get_test_db_name()
+                assert 'file:memorydb' in db_name or db_name == ':memory:'
```
blueyed: Why the `or` here?
I think it should be just one assertion per line.  Maybe a leftover?

tests/test_db_setup.py
adamantike: Yes, it's just for that. I can sort `connections.all()` instead to simplify the code.

tests/test_db_setup.py
adamantike: This is copy-pasted from the previous test. Let me remove it and check if CI doesn't complain 🙂

tests/test_db_setup.py
blueyed: Would be good, yes.

blueyed: Or add a comment.

adamantike:
@blueyed comments addressed! I can see Travis is only failing because `pytest-xdist` is optional now, so some environments don't recognize the `-n1` option. I can investigate what's happening, and open a different PR to address it.

adamantike:
Just discovered the `pytest.importorskip("xdist")` notation added to tests now, so it should pass now.

adamantike:
@blueyed this is ready to review!

blueyed:
Thank you, great work!

blueyed:
Will be released in 3.4.8 - waiting for CI/CD.

adamantike:
Thanks for the assistance @blueyed !

PR #680, opened 2018-12-17
Rename test databases when running parallel Tox

### LINKED ISSUE #678: Correctly support DB access in parallel Tox testing
Hi!

Recently, I have migrated some Django projects to run their tests using [Tox](https://tox.readthedocs.io/en/latest/index.html). Because of the number of tests and Tox environments, it was configured to run tests in parallel. This required a similar approach as what's currently handled by `pytest-django` when running tests with `pytest-xdist`: **Database renaming to avoid collisions**.

A simple fixture like `django_db_modify_db_settings_xdist_suffix`, to add a suffix to database names, was enough to fix database collisions, so I wonder if `pytest-django` should also handle Tox testing in these scenarios. And, now more than ever, with Tox implementing real support for [parallel execution](https://github.com/tox-dev/tox/pull/1102).

I have the time to work on this, and provide what I've implemented so far, if this request is accepted!

adamantike:
When tests are executed using Tox in parallel, modify the test database
names, to avoid name collisions between processes.

This change renames the existing
`django_db_modify_db_settings_xdist_suffix` fixture, to a generic
`django_db_modify_db_settings_parallel_suffix` one, in case more
scenarios/tools have to be considered in the future.

It also handles projects where both `pytest-xdist` and parallel `tox`
are being using, generating database names like
`test_default_py37-django21_gw0`.

This change is based on `tox` new parallel mode [0].

Resolves #678.

[0] https://github.com/tox-dev/tox/pull/1102

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ _build
 .env
 /.coverage.*
 /.coverage
+/coverage.xml
 /htmlcov/
 .cache
 .pytest_cache/
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -191,10 +191,32 @@ If you need to customize the location of your test database, this is the
 fixture you want to override.
 
 The default implementation of this fixture requests the
-:fixture:`django_db_modify_db_settings_xdist_suffix` to provide compatibility
+:fixture:`django_db_modify_db_settings_parallel_suffix` to provide compatibility
 with pytest-xdist.
 
 This fixture is by default requested from :fixture:`django_db_setup`.
+
+django_db_modify_db_settings_parallel_suffix
+""""""""""""""""""""""""""""""""""""""""""""
+
+.. fixture:: django_db_modify_db_settings_parallel_suffix
+
+Requesting this fixture will add a suffix to the database name when the tests
+are run via `pytest-xdist`, or via `tox` in parallel mode.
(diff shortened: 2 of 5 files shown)


adamantike:
@blueyed updated to avoid removing the previous fixture, and avoided code duplication between the two new fixtures.

This now depends on https://github.com/tox-dev/tox/pull/1139 being merged, for the `TOX_PARALLEL_ENV` to be present, and we can identify a parallel run.

blueyed:
https://github.com/tox-dev/tox/pull/1139 was merged for tox, so we can proceed here.
I would like to have this covered 100%.

blueyed:
Please rebase / fix coverage.

adamantike:
@blueyed, rebased and improved coverage! There's only one line without coverage at the moment, which requires the database engine not to be `django.db.backends.sqlite3` (and is also uncovered in `master`). Trying to use a different engine requires specific test classes for MySQL or Postgres in `test_db_setup.py`, something that hasn't been done yet.

Let me know how you'd like to proceed :)

adamantike:
@blueyed bump!

blueyed:
@adamantike 
Thanks!
(and sorry for the slow response)

PR #687, opened 2019-01-10
Admin user pw

sreyemnayr:
This one should pass (except for 2.7)


blueyed:
It did not.

bluetech:
I believe c0d10af86eb737f5ac4ba42b1f4361b66d3c6c18 fixes this (if I understand what the problem here is correctly). Will be in 4.1.0.

PR #697, opened 2019-02-01
install_requires: do not allow pytest 4.2.0

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/4704

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ setup(
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     setup_requires=['setuptools_scm>=1.11.1'],
     install_requires=[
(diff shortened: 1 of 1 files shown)


nicoddemus:
This probably should be mentioned in the CHANGELOG right?

nicoddemus:
Hmm on second thought it would probably be nice to mention that there's an incompatibility that's being work out.

blueyed:
@nicoddemus 
Yes, it might make sense to mention that it is a known issue, and not pinned just-because.

But it should also have been pinned to `<4.2` - I've done it a bit hastly due to having no time after all, but considered it to be rather critical.

nicoddemus:
Sure thing, no worries, appreciate taking the time to do it even if you don't have it. 👍

Issue #698, opened 2019-02-01
Incompatible with pytest 4.2


blueyed:
https://github.com/pytest-dev/pytest/issues/4704#issuecomment-459693344

https://github.com/pytest-dev/pytest-django/pull/697#issuecomment-459702597 disallows pytest 4.2.0 (hoped to have it fixed there with 4.2.1, but it is likely a bug/issue in pytest-django itself)

PR #700, opened 2019-02-01
Fix/skip disabling/handling of unittest methods with pytest 4.2+

blueyed:
Fixes https://github.com/pytest-dev/pytest-django/issues/698.


blueyed:
Well, not that easy: using pytest 4.2's method will not block database access anymore, and there seems to be another failing test.
https://travis-ci.org/pytest-dev/pytest-django/jobs/487630467#L233

PR #711, opened 2019-03-15
[WIP] Prefix all fixtures with "django_"

blueyed:
TODO:

- [ ] doc


pytest_django/plugin.py
```diff
@@ -44,6 +45,17 @@
...
 
+# For backward compatibility.
+admin_client = django_admin_client
+admin_user = django_admin_user
+client = django_client
+db = django_db
+live_server = django_live_server
+rf = django_rf
+settings = django_settings
+transactional_db = django_transactional_db
```
blueyed: This is problematic for `x in request.funcargnames` checks being used.. :/

blueyed:
Updated docs, and started to use new fixtures only internally, but it is quite messy.

What also should be considered is using prefixes with the option names..

PR #717, opened 2019-04-13
Fix #716 -- Check if models are ready

### LINKED ISSUE #716: Site cache clear does not work
If you use `pytest-django` with `django.contrib.sites` inside your `INSTALLED_APPS` you will get, the following exception:

```
venv/lib/python3.7/site-packages/pytest_django/plugin.py:702: in _django_clear_site_cache
    from django.contrib.sites.models import Site
venv/lib/python3.7/site-packages/django/contrib/sites/models.py:78: in <module>
    class Site(models.Model):
venv/lib/python3.7/site-packages/django/db/models/base.py:103: in __new__
    app_config = apps.get_containing_app_config(module)
venv/lib/python3.7/site-packages/django/apps/registry.py:252: in get_containing_app_config
    self.check_apps_ready()
venv/lib/python3.7/site-packages/django/apps/registry.py:135: in check_apps_ready
    raise AppRegistryNotReady("Apps aren't loaded yet.")
E   django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.
```

I could fix this, be overwriting the extension and actually checking if the models are ready:

```python
import pytest
from django.apps import apps
from pytest_django.lazy_django import django_settings_is_configured


@pytest.fixture(autouse=True, scope="function")
def _django_clear_site_cache():
    """Clears ``django.contrib.sites.models.SITE_CACHE`` to avoid
    unexpected behavior with cached site objects.
    """

    if django_settings_is_configured() and apps.models_ready:
        from django.conf import settings as dj_settings

        if "django.contrib.sites" in dj_settings.INSTALLED_APPS:
            from django.contrib.sites.models import Site

            Site.objects.clear_cache()
```

codingjoe:
You can not access the `Site` model before the models are ready.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -718,8 +718,9 @@ def _django_clear_site_cache():
     """Clears ``django.contrib.sites.models.SITE_CACHE`` to avoid
     unexpected behavior with cached site objects.
     """
+    from django.apps import apps
 
-    if django_settings_is_configured():
+    if django_settings_is_configured() and apps.models_ready:
         from django.conf import settings as dj_settings
 
         if "django.contrib.sites" in dj_settings.INSTALLED_APPS:


codingjoe:
Never mind, this was due to another package interfering.

PR #725, opened 2019-05-05
Add django_debug_sql ini option

blueyed:
TODO:

- [ ] doc
- [ ] use a command line option instead?

I've thought about having a fixture also, but it would require to set
"force_debug_cursor" on the connections then, and it probably not
useful; typically you want to use this for a short time only - therefore
a command line option might be better suited also (but you can use `-o
django_debug_sql = 1`).  And on the other hand, it will only show up on
failures, and is therefore maybe good to set it in general.


fopina:
I'd really like to see this baked into pytest-django!

I hadn't seen this PR, but I'm currently using it through conftest.py like [this](https://gist.github.com/fopina/6b6adf195c41a287da421dda5b73aaa6)

danihodovic:
This would indeed be very handy.

Issue #710, opened 2019-03-13
Error with Django internal when using pytest >= 4.2


debnet:
Hello,

Since I updated pytest to 4.2, I've got errors inside Django internals for some tests.
These tests worked fine with 4.1 and still work fine with Django embedded unittests.

Here the error:

```
______________________________________________________________________________ UserPageCheckTest.test_get_view_user_by_organisation _______________________________________________________________________________

self = <drop.front.tests.test_pages.UserPageCheckTest testMethod=test_get_view_user_by_organisation>, url = <URLPattern 'user/<int:pk>/' [name='view_user']>, pk_required = True, login_required = True

    def test_get_by_organisation(self, url=url, pk_required=pk_required, login_required=login_required):
>       base_url = self.get(url, pk_required, login_required)

drop/front/tests/test_pages.py:69: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
drop/front/tests/test_pages.py:49: in get
    self.client.force_login(self.user)
../../.pyenv/versions/drop/lib/python3.7/site-packages/django/test/client.py:611: in force_login
    self._login(user, backend)
../../.pyenv/versions/drop/lib/python3.7/site-packages/django/test/client.py:624: in _login
    login(request, user, backend)
../../.pyenv/versions/drop/lib/python3.7/site-packages/django/contrib/auth/__init__.py:132: in login
    user_logged_in.send(sender=user.__class__, request=request, user=user)
../../.pyenv/versions/drop/lib/python3.7/site-packages/django/dispatch/dispatcher.py:175: in send
    for receiver in self._live_receivers(sender)
../../.pyenv/versions/drop/lib/python3.7/site-packages/django/dispatch/dispatcher.py:175: in <listcomp>
    for receiver in self._live_receivers(sender)
../../.pyenv/versions/drop/lib/python3.7/site-packages/django/contrib/auth/models.py:20: in update_last_login
    user.save(update_fields=['last_login'])
[...]

PR #740, opened 2019-05-30
Revisit unittest handling

blueyed:
While looking into https://github.com/pytest-dev/pytest-django/issues/710.


pytest_django/plugin.py
```diff
@@ -512,43 +512,42 @@ def _django_db_marker(request):
...
-            if not skipped:
-                self._pre_setup()
-            super(cls, self).debug()
-            if not skipped:
-                self._post_teardown()
+        if not skipped:
+            self._pre_setup()
+        super(cls, self).debug()
+        if not skipped:
+            self._post_teardown()
```
blueyed: Only `not skipped` is covered here.  But also previously.

PR #744, opened 2019-06-04
WIP: drop support for pytest < 4.2

blueyed:


boxed:
This is actually a pretty huge deal! The import of pkg_resources is pretty expensive. It amounts to something like 20% of the overhead of my test runner hammett when using this plugin!

blueyed:
@boxed 
Yeah, importing/using pkg_resources just for this is really bad after all.  pytest should probably have a `version_info` tuple.

I'll look into removing this separately already.

boxed:
In this case it seems like it would be safe to just do `.startswith('4.')` anyway. I doubt anyone cares about 3, 2 or 1.

blueyed:
@boxed done, thanks for pointing it out!

blueyed:
@boxed as for performance in pytest itself see https://github.com/pytest-dev/pytest/pull/4376.
I might take this into blueyed/pytest soon, you might want to try adding your "test case" / scenario there already.

PR #751, opened 2019-06-29
Fix compat with pytest 5.x

rpkilby:
The following is preventing compatibility with the newly release pytest 5.0:
```python
pytest.PytestDeprecationWarning: The `funcargnames` attribute was an alias for `fixturenames`, since pytest 2.3 - use the newer attribute instead.
```

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -129,7 +129,7 @@ def _django_db_fixture_helper(
     if is_django_unittest(request):
         return
 
-    if not transactional and "live_server" in request.funcargnames:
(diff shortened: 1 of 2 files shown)


rpkilby:
Note that pytest 5.x has dropped py27 and py34 support, although, the related builds seem to be passing. I'm assuming that pytest-django is just not hitting those code paths where pytest would otherwise fail.

blueyed:
> Note that pytest 5.x has dropped py27 and py34 support, although, the related builds seem to be passing. I'm assuming that pytest-django is just not hitting those code paths where pytest would otherwise fail.

It is using pytest 4.6 there.

Issue #757, opened 2019-08-29
Provide easy access to version info.


fermezz:
I understand [`setuptools_scm` provides programmatic access to version info](https://github.com/pypa/setuptools_scm#programmatic-usage), but I think it would be nice if we could provide easy access in here, kinda following what `pytest` is doing. I can work on a PR, although if this is not something we'd want to do, let me know!

blueyed:
Just out of interest: what's your use case?

fermezz:
I have a few projects still running old versions of Django which latest `pytest-django` doesn't really support, so they don't have access to newer features. The idea is to port the implementation into those projects only when running `pytest-django` with a version lower than `<some-version-of-pytest-django>` so that once we can get them updated, those feature-ports are easily identifiable and removable. At the moment there's no easy way to do that. We _could_ work around it by checking the Django version but other variables might come into play so I'm not sure that's optimal.

Does that make sense?

PR #758, opened 2019-08-29
Generate _version.py file with setuptools_scm

fermezz:
Solves #757 by generating a `_version.py` file with `setuptools_scm`. I think using `post-release` as `version-scheme` is good because it lets you know the current version of the project, without doing anything more than splitting a string. Feedback is appreciated!


blueyed:
I'd rather recommend using https://github.com/zsimic/setupmeta/.

fermezz:
Ok nice! I was not aware of the `setupmeta` project. I have added a `__version__` attribute under `pytest_django/__init__.py` which matches what should be the version of this project if this is squashed and merged (3.5.1.post2). We can now have a workflow which would look like this for tagging a new version:

```(pytest-django) [fermezz@thinkpad pytest-django]$ python setup.py v→
  running version
  Not pushing bump, use --push to push
  Running: git add pytest_django/__init__.py
  Running: git commit -m "Version 3.5.2"
  [master 92eda6e] Version 3.5.2
   1 file changed, 1 insertion(+), 1 deletion(-)
  Running: git tag -a v3.5.2 -m "Version 3.5.2"
  (pytest-django) [fermezz@thinkpad pytest-django]$ cat pytest_django/__→
  __version__ = '3.5.2'```

_and_ we can get the version of the project as simple as `import pytest_django; pytest_django.__version__` which is great.

I didn't feel like it was necessary to put constraints to the version of `setupmeta` but if you think otherwise let me know!

fermezz:
@blueyed ping for awareness! :)

blueyed:
@fermezz 
Sorry for late reply, and thanks for this!

I'm not too sure about setupmeta anymore given that setuptools-scm is used already.

Also `pytest-django-3.7.1.dev3+g2bb0135` (master) looks better than `pytest-django-3.7.0.post4` (your cherry-picked PR), so it should use `versioning=devcommit` (resulting in `3.7.1.dev4-g64fd5d6-dirty`).

I'll leave this open for now (and can amend/change it to "devcommit" myself of course).

What do you think?

fermezz:
Totally fine, @blueyed! I can take on that tomorrow, or the day after :)

fermezz:
@blueyed Ok, so I went ahead and changed the versioning to `devcommit`. It's not as-straightforward to have that same versioning scheme with setuptools-scm and I don't see a big issue on changing it, but if you have a strong opinion on that, please let me know!

blueyed:
@fermezz 
I've meant to keep using setuptools-scm more likely, but am still not sure.

What's your motiviation btw to have the version there?

bluetech:
Done as part of #880.

PR #759, opened 2019-09-05
marker to ensure that fresh version of .mo files are present

imomaliev:
Sometimes there is a need to make sure that tests are using fresh version of .mo files. So tests should make sure that `compilemessages` is called before such tests

TODO:

- [x] languages arg for marker
- [ ] tests

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -6,6 +6,7 @@ import os
 import warnings
 from contextlib import contextmanager
 from functools import partial
+from io import StringIO
 
 import pytest
 
@@ -151,6 +152,24 @@ def _django_db_fixture_helper(
     test_case = django_case(methodName="__init__")
     test_case._pre_setup()
     request.addfinalizer(test_case._post_teardown)
+
+
+@pytest.fixture(scope="session")
+def django_locale(request):
+    return request.config.getvalue("locale")
+
+
+def django_i18n_fixture_helper(locale):
+    from django.core.management import call_command
+
+    call_command("compilemessages", locale=django_locale, stderr=StringIO())
+
+
+@pytest.fixture(scope="session")
+def django_i18n_setup(request, django_locale):
+
+    if not request.config.getvalue("reuse_i18n"):
+        django_i18n_fixture_helper(django_locale)
 
 
 def _disable_native_migrations():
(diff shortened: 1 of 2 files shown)


blueyed:
> So tests should make sure that `compilemessages` is called before such tests

Hmm.  This seems to be rather specific - I do not think it should be done by default.

Instead it might be good to have a code snippet in the docs that allows for doing it yourself in/via your conftest.

What do you think?

blueyed:
Alternatively I think we could have a fixture that would do it when being requested/used?

imomaliev:
Hi @blueyed. There already is marker. That when enabled will be run this command. It works similar to `django_db` marker

bluetech:
Hi @imomaliev,

`compilemessages` is really only meant to be called once. It also likely affects the git status which seems inappropriate for a test suite. So this PR doesn't seem like the right approach to me.

> Sometimes there is a need to make sure that tests are using fresh version of .mo files

This perhaps makes sense to have (or document as a recipe as @blueyed said), but I suggest opening an issue to discuss the approach.

imomaliev:
@bluetech Hi, ok I will start with creating an issue. But few things to note

1.  .mo shouldn't be stored in git. 
2. `compilemessages` should be rerun after .po files have been changed.
3. I may be wrong but for me it almost the same as applying migrations  - this is an expected state of a project and when you are running tests you could have test failures just because your migrations weren't applied or your .mo files weren't compiled properly

bluetech:
1. I for one perfer to commit them. I run `compilemessages` after I change the po files and commit them. Seems like django itself [does the same](https://github.com/django/django/tree/main/django/contrib/sessions/locale/af/LC_MESSAGES). Both approaches are valid, but I wouldn't say they shouldn't be stored in get.
2. Definitely!
3. It's completely valid to do as you suggest, however not everyone work this way. I prefer to do the `compilemessages` only once, instead of having it slow down every test run.

imomaliev:
> 1. I for one perfer to commit them. I run compilemessages after I change the po files and commit them. Seems like django itself does the same. Both approaches are valid, but I wouldn't say they shouldn't be stored in git.

Yeah, django stores them as well, but I don't think this is the right way, because it is an artefact like .pyc files. But this very well may be subjective. Either way this is one of current use-cases and should be taken into consideration. This is why marker I made runs only when applied to ensure that .mo files are up to date.

Ok, let's continue this discussion in an issue. May be the right way would be just put this in separate plugin

PR #768, opened 2019-10-07
Add "python_paths_early" and "site_dirs_early" settings.

pakal:
These settings allow - contrary to pytest-pythonpath plugin - to modify sys.path BEFORE django.setup() is called by early pytest initialization.

This PR is only a first draft, if the solution looks OK to maintainers, I'll reformat and document all this.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -6,6 +6,7 @@ test database and provides some useful text fixtures.
 
 import contextlib
 import inspect
+import site
 from functools import reduce
 import os
 import sys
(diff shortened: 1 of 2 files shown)


pakal:
The build failed with a Flakes error expected mapping but got sequence", but on a line I didn't touch  :?
It's weird furthermore, because "%" also works with sequences, especially in this case :

        raise ValueError(
            "{} is not a valid value for {}. "
            "It must be one of {}." % (x, name, ", ".join(possible_values.keys()))
        )

bluetech:
> These settings allow - contrary to pytest-pythonpath plugin - to modify sys.path BEFORE django.setup() is called by early pytest initialization.

pytest-pythonpath patches the pythonpath like this:

```py
@pytest.mark.tryfirst
def pytest_load_initial_conftests(args, early_config, parser):
```

and django first runs `django.setup()` like this:

```py
def pytest_load_initial_conftests(early_config, parser, args):
```

Because of the `tryfirst`, `pytest-pythonpath` is supposed to run first.

Does this not happen?

pakal:
Whoups, when I had checked the pythonpaths plugin it used pytest_configure(), but I see now that they have switched to "tryfirst" pytest_load_initial_conftests for quiet a long time already... I must have checked the wrong version *facepalm*

No need anymore for this PR so  ^^'

PR #776, opened 2019-10-23
Repro issue 713

prescod:
I built on Michel Samia's work to reformat into the format suggested by @blueyed. 

#713

PR #777, opened 2019-10-23
[WIP] for issue 713: INTERNALERROR due to DB access in repr

blueyed:
Followup-to https://github.com/pytest-dev/pytest-django/pull/776.

- adjusted test


blueyed:
Note that this passes with https://github.com/pytest-dev/pytest/pull/6047, but I'll get something with a shorter test.

Issue #341, opened 2016-06-14
_django_db_setup fails with data migrations


kuboschek:
## Situation

In my project, I have a data migration that transfers one value of a certain model to another property of the same model. During test setup, all migrations are run as per default fixture of pytest. 
## Expected behavior

pytest-django runs all migrations, including data migrations
## Actual behavior

All tests end with errors, due to DB use being disallowed during test setup
## Versions

platform linux2 -- Python 2.7.6, pytest-2.9.2, py-1.4.31, pluggy-0.3.1
django settings: project.mysite.settings.local (from ini file)
rootdir: /vagrant, inifile: pytest.ini
plugins: django-2.9.1, pythonpath-0.7
## Possible resolution

Is there a way of still using that default fixture, but allowing DB access, or not running this specific migration?

pelme:
Without seeing the code and exception traceback, it is very hard to figure out what is going on and what the problem is.

Please provide steps and information for how to reproduce the error. A PR with a test case that shows the issue would of course be even better and increase the chance that we can get the issue fixed faster. :)

tovmeod:
I'm having a similar issue, maybe it is caused by the same bug.

I have a data migration providing initial value for a model, populating the table.

When running old style unittest Class based tests it runs correctly

When running a parametrized test it doesn't apply the data migration, causing the test to fail.

Below is the test:

@pytest.mark.skipif(os.environ.get('SELENIUM', False), reason='Selenium env is set to 1')
@pytest.mark.parametrize('url', _get_noargs_urls())
@pytest.mark.django_db(transaction=True)
def test_no_args_urls_loggedin(url, client, userpsw):
    """All URLs without arguments should return 200 if logged in"""
    user, psw = userpsw
    client.login(username=user, password=psw)
    response = client.get(url)
    # assert response.status_code == 200

Can I tell it to force running migrations?

davo-pkools:
I think I have a similar situation - essentially it appears the migration is applied to the database after completion of the first test case, before execution of the second. This code demonstrates the issue I am seeing:

```
import pytest
from django.core.management import call_command
from foo import bar

@pytest.fixture(scope='session')
def django_db_setup(django_db_setup, django_db_blocker):
    LOGGER.info('ran call_command')
    with django_db_blocker.unblock():
        call_command('loaddata', 'XXX.json')

@pytest.mark.django_db(transaction=True)
def test_t1():
    assert len(bar.objects.all())

[...]

PR #778, opened 2019-10-23
Unblock DB in case of error for reports

blueyed:
This mainly works around pytest causing an INTERNALERROR for when using
`repr()` on the `pytest.fail.Exception` that is used when the DB is
blocked.

While this will be fixed via
https://github.com/pytest-dev/pytest/pull/6047 likely, it might be good
to have this anyway (also for older pytest versions).

On the other hand this might cause the DB to be accessed/changed then,
which might show different results from when the test failed due to
this?!

Fixes https://github.com/pytest-dev/pytest-django/issues/713.
Fixes https://github.com/pytest-dev/pytest-django/issues/341.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -427,6 +427,16 @@ def pytest_runtest_setup(item):
             _disable_class_methods(item.cls)
 
 
+@pytest.hookimpl(hookwrapper=True, tryfirst=True)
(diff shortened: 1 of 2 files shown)


blueyed:
Going with https://github.com/pytest-dev/pytest-django/pull/781 instead.

PR #784, opened 2019-11-18
Catch exception when explicitly enabling thread-shareability on sqlite3 memory

### LINKED ISSUE #783: Live server error with sqlite3
When using the live_server in an integration test fixture:
```python
@pytest.fixture(scope='session', name='server_url')
def get_server_url(live_server, environment):
    """Start a django server to test against and return the url of the server."""
    yield live_server.url
```
the following error is generated:

```

self = <[AttributeError("'LiveServer' object has no attribute 'thread'") raised in repr()] LiveServer object at 0x80526a8d0>
addr = 'localhost'

    def __init__(self, addr):
        import django
        from django.db import connections
        from django.test.testcases import LiveServerThread
        from django.test.utils import modify_settings
    
        connections_override = {}
        for conn in connections.all():
            # If using in-memory sqlite databases, pass the connections to
            # the server thread.
            if (
                conn.settings_dict["ENGINE"] == "django.db.backends.sqlite3"
                and conn.settings_dict["NAME"] == ":memory:"
            ):
                # Explicitly enable thread-shareability for this connection
>               conn.allow_thread_sharing = True
E               AttributeError: can't set attribute

/home/goldeneagle/venv/lib/python3.7/site-packages/pytest_django/live_server_helper.py:27: AttributeError
```

cleder:
Fixes #783

### DIFF
--- a/pytest_django/live_server_helper.py
+++ b/pytest_django/live_server_helper.py
@@ -24,8 +24,11 @@ class LiveServer(object):
                 and conn.settings_dict["NAME"] == ":memory:"
             ):
                 # Explicitly enable thread-shareability for this connection
-                conn.allow_thread_sharing = True
-                connections_override[conn.alias] = conn
+                try:
+                    conn.allow_thread_sharing = True
+                    connections_override[conn.alias] = conn
+                except AttributeError:
+                    pass
 
         liveserver_kwargs = {"connections_override": connections_override}
         from django.conf import settings


pytest_django/live_server_helper.py
```diff
@@ -24,8 +24,11 @@ def __init__(self, addr):
                 and conn.settings_dict["NAME"] == ":memory:"
             ):
                 # Explicitly enable thread-shareability for this connection
-                conn.allow_thread_sharing = True
-                connections_override[conn.alias] = conn
+                try:
+                    conn.allow_thread_sharing = True
+                    connections_override[conn.alias] = conn
+                except AttributeError:
+                    pass
```
blueyed: It should use `inc_thread_sharing()` with Django 2.2+ I think (https://github.com/django/django/commit/76990cbbda5d93fda560c8a5ab019860f7efaab7).

blueyed:
Closing in favor of https://github.com/pytest-dev/pytest-django/pull/793.

PR #787, opened 2019-11-29
WIP: Fixtures for Template

erayerdin:
This isa feature pull request.

# Rationale

`pytest-django` already covers most of the helpers and fixtures we need, yet it lacks some especially for testing templates and template tags. So, this PR aims to ease testing these aspects. For now, this PR contains some simple fixtures mentioned in their own sections, all of which are implemented in a factory like pattern. The pattern is as below:

```python
@pytest.fixture(scope="function")
def foo_factory():
	"""A factory fixture (returning a function) that generates `Foo` instance.
	"""

	def factory():
		# this function might also take parameters to ease object generation
		
		# assuming we have an instance of class `Foo` named `foo`
		foo = Foo()
		return foo
	
	return factory
```

While this is the pattern for the fixtures implemented yet, it does not have to be necessarily so.

## Fixtures Already Implemented

Below are fixtures that are already implemented.

### `template_factory`

A factory fixture to ease the initialization of `django.template.Template`. An example:

```python
def test_whatever(template_factory):
	template = template_factory("foo")
	assert "foo" in template.source
```

This simply creates a `django.template.Template` instance with source `"foo"`.

### `context_factory`

A factory fixture to ease the initialization of `django.template.Context`. This is usually to be used with `template_factory` fixture. In order to render a `Template` instance, you also need a `Context` instance. See the example:

```python
def test_whatever(template_factory, context_factory):
	template = template_factory("Hello, {{ name }}.")
	rendered_template = template.render(context_factory({"name": "world"))
	assert rendered_template == "Hello world."
```

## Fixtures to be Implemented

Until now, it just does some imports from `django.template` package and removes *barely little* boilerplate code while testing templates, which might seem unpractical. However, I have just implemented fixtures above to (i) present the idea of testing templates and introduce it to the codebase as early as possible and (ii) discuss the design choices (factory-like pattern, or naming as `_factory` suffix instead of initials as in `RequestFactory`'s `rf`) or even whether it is necessary to implement it. If it is decided that it's not necessary, I would not like to invest more time.

To decide if this is necessary or a good thing to merge, I also would like to present a couple of fixtures in my mind that will ease testing of templates and template tags. These can also be subject to change.

### `tag_template_factory`

`tag_template_factory` will simply create a template, including `load` and the tag itself with its *args* and *kwargs*. To demonstrate:

```python
# assuming there is a `simple_tag` named `foo` under `myapp.templatetags.mytagmod`

def test_whatever(tag_template_factory, context_factory):
	template = tag_template_factory(
		module="mytagmod",
		name="foo",
		args=("lorem", "ipsum"),
		kwargs={"foo": "bar"}
	)
	# below would pass
	assert template.source == "{% load mytagmod %}{% foo 'lorem' 'ipsum' foo='bar' %}"
```

### Fixture(s) for Testing Tag Filters

I haven't thought about it enough. Reading [this section](https://docs.djangoproject.com/en/2.2/howto/custom-template-tags/#writing-custom-template-filters) to understand.

### Fixture(s) for Testing Inclusion Tags

I haven't give this one enough thinking either. Reading [this section](https://docs.djangoproject.com/en/2.2/howto/custom-template-tags/#inclusion-tags) to understand. This one will probably be similar to simple tags because they have similarities.


bluetech:
Hi @erayerdin, can you explain how the new factory fixtures are better than using the types directly?

bluetech:
I'll close this now, but feel free to reopen if you want to go back to it.

PR #793, opened 2020-01-05
live_server: fix in-memory DB detection, dj 2.2 compat

blueyed:
Fixes https://github.com/pytest-dev/pytest-django/issues/783.

TODO:

- `dec_thread_sharing()` ?  (also for dj < 2.2 then likely)

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -9,7 +9,6 @@ from functools import partial
 
 import pytest
 
-from . import live_server_helper
 from .django_compat import is_django_unittest
 from .lazy_django import skip_if_no_django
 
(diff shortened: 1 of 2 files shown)


githubdd7:
Are there any plans to merge this change ? We think this can fix our pytest errors. Thank You.

asfaltboy:
I'd love to pick it up, @blueyed could you provide some background on this PR, i.e with the changes that remove the thread logic, how will this still work?

boxed:
I hit this exact problem today! Would love it if this was merged and released!

bluetech:
Fixed in #973 (didn't see this one before...)

PR #802, opened 2020-01-17
Added missing decorators to tests that depended on the DB but weren't marked as such

SalmonMode:
Just a simple fix, but some tests were missing the marker that tells the `_django_db_marker` fixture that they need the db.

### DIFF
--- a/tests/test_fixtures.py
+++ b/tests/test_fixtures.py
@@ -617,6 +617,7 @@ class Test_django_db_blocker:
             with pytest.raises(RuntimeError):
                 Item.objects.exists()
 
+    @pytest.mark.django_db
     def test_unblock_manually(self, django_db_blocker):
         try:
             django_db_blocker.unblock()
@@ -624,6 +625,7 @@ class Test_django_db_blocker:
         finally:
             django_db_blocker.restore()
 
+    @pytest.mark.django_db
     def test_unblock_with_block(self, django_db_blocker):
         with django_db_blocker.unblock():
             Item.objects.exists()


tests/test_fixtures.py
```diff
@@ -617,13 +617,15 @@ def test_block_with_block(self, django_db_blocker):
...
         try:
             django_db_blocker.unblock()
             Item.objects.exists()
         finally:
             django_db_blocker.restore()
 
+    @pytest.mark.django_db
     def test_unblock_with_block(self, django_db_blocker):
         with django_db_blocker.unblock():
             Item.objects.exists()
```
blueyed: Thanks, but I think the point of the tests is to actually ensure that the unblocking can be used manually, without requesting the `db` fixture / using the mark, which basically only triggers the unblocking for the test outside already.

It might be a good case for the pytest issue maybe, if you/we can figure out what is happening here maybe?

SalmonMode: @blueyed Ah, my mistake. I didn't realize. I'll dig into it and see what I can find

SalmonMode: @blueyed I think I've figured. Try running `Test_django_db_blocker.test_unblock_manually` on pytest 5.3.2 so my changes aren't included. One would expect the test to pass.

SalmonMode: If it fails, it's because it's not actually setting up the DB.

blueyed: @SalmonMode 
That fails with "django.db.utils.OperationalError: unable to open database file" for me, since it tries to access the "'/should_never_be_accessed'" DB (initial setting, adjusted when setting up the DB).
So this is also kind of expected in general - although should not fail in general.
pytest-django itself is used as a plugin etc and tests are rather fragile (and therefore this ("running that test alone") is not surprising).
It is the same with both 5.3.2 and 5.3.3 in that regard, isn't it?

SalmonMode: @blueyed yes. That's the problem. A test should be able to run entirely on its own. In this case, the `django_db_setup` fixture is being called out of order, and outside the fixture system by `_django_setup_unittest`, where it does `request.getfixturevalue("django_db_setup")`. Before, this "worked out" because the fixture system wasn't properly cleaning up after itself, and when I ran `pytest --setup-plan`, there were several fixtures that weren't even listed even though they were definitely running.

If the fixture system is not used in the normal requesting/autouse way, then it can't possibly track down when things should and shouldn't be torn down or even set up.

blueyed: The test kind of works, but needs to handle also this exception: it is meant to allow access to the DB.  If you use the `db` fixture before, pytest-django sets up the test DB, if you don't do that it does not.  Since pytest-django itself is a plugin once the "db" fixture is used the DB name is adjusted already (a flaw with the tests, and also kind of with pytest's global state in this regard).
I.e. I think the test is fine in general - it allows access to the DB in its current state.
Keep in mind that it fails the same way in 5.3.2, so there is nothing that regressed with this test.

blueyed: As said before the xdist test failure might be more interesting for debugging / figuring out the issue (also in pytest-django then maybe, of course).

There are other issues which go into an endless recursion, also without pytest-django.
They might also "misuse" the fixture system, but it clearly is a regression then still, and we need to investigate how to fix/address this.

SalmonMode: @blueyed Sorry, I edited my previous comment a bunch while you were typing yours it seems.

The cause of the issue is using `request.getfuncargvalue("fixture_name")` to execute fixtures prevents the relevant those fixtures from being listed in that `_dependee_fixture_argnames` list (really the `request.fixturenames` list). Because a fixture is not in there when executed by this method, when it executes it adds its finalizer to every fixture that _is_ in there, so whenever any of them teardown, it gets torn down as well, because the `request.fixturenames` list can't be properly sliced

As far as I can tell, `django-pytest` has been using the `request.getfuncargvalue('django_db_setup')` approach for at least 7 years, and `django_db_setup` isn't the only affected fixture. I'm not familiar with what pytest was like back then, but I'm guessing it was a far cry from what it is today, and the fixture system wasn't as feature rich as it is now.

For a short term solution, it _might_ be possible to insert the fixture being executed by `request.getfuncargvalue` into the `request.fixturenames` list, but getting it into the right spot would be pretty tricky if not impossible, and may have plenty of consequences.

It doesn't sound like too fun of an exercise, but I think the best solution is to refactor the plugin to properly use the fixture system.

blueyed: @SalmonMode thanks for the clarification.  AFAICS pytest-django needs to dynamically request fixtures, and as you can see from the issues in pytest others might do something similar themselves (and the recursionerror appears to be another bug/issue in itself).
I really think we should revert this for now and then try it again, hopefully with some tests that replicate what we've seen here and elsewhere.  /cc @nicoddemus

blueyed: I agree though however that the code is old, and that it could be improved.
I've tried several times to e.g. not require `runpytest_subprocess` for own tests, but it is really hairy since pytest itself loads pytest-django already etc.  There are several WIPs in that regard, but some of them not even wanted by pytest (so only in my fork currently).

blueyed: (@SalmonMode 
I also totally can imaging/feel your sentiment about the issue in general - unfortunately a lot of corner cases in pytest are not covered in its own tests.)

SalmonMode: @blueyed I think I know the solution for the recursion stuff. I'm guessing when an extreme amount of fixtures are involved a fixture can end up adding it's finalizer to another fixture numerous times, so it's not an infinite recursion issue, but an exponentially increasing recursion issue (which might as well be infinite haha). To fix it, I just have to throw in a check to see if it's already registered its finalizer before adding it to the fixture it's looking at.

As for the original issue, I agree that we should revert it so I can polish it up. But the core issue is still present. Pytest-django isn't calling fixtures in an appropriate way. Assuming I polish up my changes, they're still necessary to make pytest fixtures setup and teardown in a deterministic way when used in the intended way. I don't think there's any way around that.

I don't think pytest itself should be modified further to accommodate pytest-django, because that will only create more opportunities for others to use those workarounds digging ourselves into a deeper hole.

I think a potential solution (for now) would be to have (in pytest-django) every fixture that calls other fixtures dynamically, first add themselves to the list of fixture names in an intelligent manner (assuming they aren't already in it), and then add the fixture they want to call to that list in the appropriate spot before calling it. I believe this will "teach" the fixtures how to manage their teardowns appropriately.

SalmonMode: @blueyed I found an additional culprit of the problem, which is that the tests are being forced into an order by a pytest hook. This is causing numerous tests to be dependent on things they aren't in control of. There are even tests meant to ensure this is the case, and they prioritize running tests with this marker, or tests that request the `db` fixture first, which made sure this problem stayed hidden (unintentionally, of course).

This is why the original 2 failing tests fail when they are run on their own. Because there are other tests in that test class with the `django_db` marker, this ordering logic ensures they happen first, and this taints the environment for the tests without that marker.

Those 2 failing tests should have already been failing, because if a user doesn't put any tests with that marker in place, hoping to rely on manually blocking/unblocking, then it won't work. There's something going on with tests that _are_ marked, where they aren't properly cleaning up after themselves, and this was the case before my changes. I'm not sure how wide spread the impact is for users of `django` and `pytest-django`, but it's very possible they have tests which aren't failing but should be.

This may also be why you found interesting results with `pytest-xdist` involved, as things could easily happen in a different order, resulting in exposing problems that would have otherwise gone unnoticed.

I have to remove the reordering logic in order to properly debug this. I would like to keep it removed once I'm done.

blueyed: The re-ordering logic is there to ensure the same order of tests like Django does it, e.g. transactional_db tests should be grouped / run first (IIRC).

blueyed: As for the 2 failing tests: I've started to fix them by running them in a subprocess/testdir.

SalmonMode: @blueyed I understand that's how django does it. That doesn't mean it's a good thing to do, nor is it something django should do. These failing tests are a good example of how it can hide problems.

As for a fix for them, I'm adjusting the fixtures provided by the plugin to maintain their original purpose but also make sure they have appropriate dependencies in terms of what fixtures they request.

blueyed: I've created https://github.com/pytest-dev/pytest-django/pull/806 to address the problem.
But I think it makes sense to have another test for what we've been seeing due to it.
Also I must have confused something since I've thought / said they were failing with 5.3.2 already.

SalmonMode: I just looked into why django orders their tests. They say it's because they need to ensure a clean database for every TestCase. The reason they go about it the way they do, is because of limitations with the `unittest.TestCase` fixture system. I'm not sure why they don't put the onus on the test writer to clean up after themselves, but they understandably want good test hygiene.

Pytest fixtures have no such limitation, and no such excuse. WIth pytest fixtures, it's easy to provide a fixture that tears itself down accordingly giving each test case a clean starting point.

SalmonMode: As for taking out the reordering logic, I think it would be best to have a more graceful departure from it. I would put out a warning that there are potentially hidden problems by relying on test order, and the plugin should discourage the reliance on it. And eventually the logic should be removed.

blueyed: It's mostly about speed also: transactional_db tests are especially slow.

SalmonMode: That I can understand. I do, however, have a plan for that as well in `pytest-xdist`. 

https://github.com/SalmonMode/pytest-xdist/tree/set-test-group

I intend to add another hook to allow reordering of "work groups" so that the test writer can dictate which ones to hand off to workers first. So if they have their slow tests in one work group, then can move that to the top of the ordered dict.

SalmonMode: Revised solution here: https://github.com/pytest-dev/pytest-django/pull/807

blueyed:
Closing this, following up in https://github.com/pytest-dev/pytest-django/pull/807 etc.

PR #804, opened 2020-01-17
Skip pytest 5.3.3

blueyed:
Ref: https://github.com/pytest-dev/pytest/issues/6492


blueyed:
Well, let's rather let pytest itself handle this.  Not wanting to cause more trouble (as can be seen by tox freaking out).

blueyed:
Doing it for CI only in https://github.com/pytest-dev/pytest-django/pull/808.

PR #806, opened 2020-01-18
tests: run Test_django_db_blocker in testdir

blueyed:
Ref: https://github.com/pytest-dev/pytest-django/pull/802#discussion_r368134433


blueyed:
@SalmonMode thanks!
Do you want to create a PR for this (in case you have it ready already locally)?
Otherwise I'll do so later.

I think this PR/change still makes sense (separately), since it simulates running the tests alone.  And it could be extended then to test what your renaming/re-arranging would fix (in a separate test).

SalmonMode:
I'd be happy to :grin: 

There's some already existing tests I can make tweaked copies of to cover this.

Sorry for causing all the ruckus, btw.

blueyed:
> I'd be happy to 

Great.  I assume this can be done on top of this one?
I.e. run pytest also with selecting only a single test (not using the db) from it.

> There's some already existing tests I can make tweaked copies of to cover this.

I'd rather avoid new tests if existing ones can be extended / hardened - the test suite is rather slow already, especially due to all the `runpytest_subprocess` tests.

> Sorry for causing all the ruckus, btw.

No worries.. :)

blueyed:
Just to clarify:

> Placing them in a subprocess fix the issue that `django_db_blocker` currently has. WHen `django_db_blocker` is requested on its own, the DB is not set up. These tests pass because of the other two tests that trigger the session scope-level `django_db_setup` fixture.  Because they always run first (due to the test reordering logic), the DB will always be prepared for the other two tests (that would fail if run on their own).

Here they are changed to fail.
But what you describe happens due to the test suite in itself (altogether) then:

> because of the other two tests that trigger the session scope-level `django_db_setup` fixture.

Which are those?

(just thought to clarify this, in case you've meant the ones in here with "the other two")

blueyed:
Forget my previous comment - I've mixed that up I guess.  Will delete it.

SalmonMode:
>I'd rather avoid new tests if existing ones can be extended / hardened - the test suite is rather slow already, especially due to all the runpytest_subprocess tests.

Not a problem. I can just tweak them in place.

>I assume this can be done on top of this one?
I.e. run pytest also with selecting only a single test (not using the db) from it.

Sorry, I'm not sure I follow. You mean if I can make changes on top of the current change you're make here?

SalmonMode:
Here's the PR https://github.com/pytest-dev/pytest-django/pull/807

PR #223, opened 2015-03-29
Ordering of tests

blueyed:
This is for #214.

TODO:

- [x] optimize (inline calls for ordering; rename)
- [ ] allow to skip this (config setting?)

PR #820, opened 2020-02-27
Don't break --failed-first when re-ordering tests

### LINKED ISSUE #819: Test re-ordering breaks Pytest's --failed-first and --stepwise options
The test re-ordering introduced in response to #214 seems to execute after pytest's own `--failed-first`, `--new-first`, `--stepwise`, etc. ordering options, breaking them.

We ran across this in mdn/kuma#6531, where even with `--failed-first` pytest was running dozens of known good tests before executing the failed tests.

Removing the `pytest_collection_modifyitems` function or decorating it with `@pytest.hookimpl(tryfirst=True)` seems to resolve this, though I'm not familiar enough with pytest to know if that's an appropriate solution, or if there's something we should be doing on our end instead.

callahad:
Pytest-django 3.5.0 attempts to run tests in the same order as Django:
Issue #223, Commit 5a79fba3705376c3f89f5a9e26b8244937d61768

Pytest itself can re-order tests to prioritize previously failed tests
(`--failed-first`), incremental runs of the test suite (`--stepwise`), etc.
However, pytest-django's re-ordering clobbers pytest's, which can break
those options as reported in #819.

By applying the `@pytest.hookimpl(tryfirst=True)` decorator to
`pytest_collection_modifyitems()`, pytest-django's re-ordering happens
first, and does not clobber the later re-ordering performed by pytest.

Fixes #819

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -415,6 +415,7 @@ def pytest_runtest_setup(item):
             _disable_class_methods(item.cls)
 
 
+@pytest.hookimpl(tryfirst=True)
 def pytest_collection_modifyitems(items):
     def get_order_number(test):
         marker_db = test.get_closest_marker('django_db')


blueyed:
Thanks!

A fix for pytest also makes sense additionally: https://github.com/blueyed/pytest/pull/262

PR #822, opened 2020-03-07
Remove Django version from --nomigrations heading

adamchainz:
1.7 is a long time ago.


blueyed:
Makes sense to remove it from the heading, but could have mentioned it in the text though (just for reference, something à la "Django migrations (added in Django 1.7) …"), but ok.

adamchainz:
Yeah sorry for impatiently merging. The whole document mentions "migrations" without referring to 'which' migrations, but I think it's pretty obvious for Django users at this point. I'm not aware of any other migrations tool having much use.

Issue #338, opened 2016-05-25
db fixture always returns None


gjcarneiro:
Is it me being stupid, or the db fixture is always None?  Looking at the code:

```python
@pytest.fixture(scope='function')
def db(request, _django_db_setup, _django_cursor_wrapper):
    return _django_db_fixture_helper(False, request, _django_cursor_wrapper)
```
But `_django_db_fixture_helper` always returns None, it has no return statement at the end of the function!  

It is my first time using this plugin, so I could be missing something obvious.  My django settings have this:
```python

DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': ':memory:',
        }
}
```
The test is along these lines:
```python
@pytest.mark.parametrize("username,expected_credit_lim", [
...
])
def test_get_customer_credit_limit(db, username, expected_credit_lim):
	_create_tables(db)
	cur = db.cursor()
        ....
```

gjcarneiro:
I'm beginning to think db being None is by design, and instead I need to just use `django.db.connection` directly.  If so, I must say this is not helpful. 

1) having a fixture with a value of None is very unintuitive, I suggest you just remove it as we then should use `@pytest.mark.django_db` instead

2) on the other hand, having a `db` fixture return a valid django connection that can be used for the test would be useful.

blueyed:
> return a valid django connection that can be used for the test would be useful. 

Agreed.

blueyed:
Well, there might be several connections involved (see https://github.com/pytest-dev/pytest-django/blob/485590ce41e6e41f710bf75de6de06245d6ab921/pytest_django/fixtures.py#L93-L100).
Should this be a list then?

As for `transactional_db` I just looked into getting to the Django `TransactionTestCase` to skip flushing tables on teardown - so in this specific case (where using `monkeypatch` seems to be better anyway) it would have been useful to return/have something in that regard.

So we could return a list of connections and the Django TestCase instance here maybe?!
On the other hand the fixture would need to get the list of connections itself for newer Django versions, so I am not really convinced.

pelme:
The db fixture is mostly used internally to tie into the fixture system. The `django_db` mark basically exists to activate the `db` fixture, because the db fixture is weird by itself (since it does not return anything meaningful). We cannot get rid of the db fixture, because it is using the pytest fixture system to manage all of the database, which is a bit more involved.

[...]

PR #832, opened 2020-03-24
Docs - explain how to get a db connection

christianmlong:
I was thinking that I could use the `db` fixture to access a database connection object within a test. However, the `db` fixture returns None. 

This documentation update makes it clear that the `db` fixture does not return an object, and explains that users should use from `django.db import connection` instead.

Reference:
See [this comment](https://github.com/pytest-dev/pytest-django/issues/338#issuecomment-227402089) on issue #338.


blueyed:
Thanks.

However, it could actually return the default connection maybe?

bluetech:
> However, it could actually return the default connection maybe?

Regardless of this, the change itself looks good. Thanks @christianmlong.

Issue #457, opened 2017-02-02
If user model changes the username field, admin_user fixture will pass it twice


TauPan:
Consider a model like this:

```python
class User(AbstractBaseUser):
    name = models.CharField(max_length=255, unique=True)
    USERNAME_FIELD = 'name'
[...]
```

(And of course ```AUTH_USER_MODEL``` points at this model.)

Then the `admin_user` fixture will raise an error:
```python
db = None, django_user_model = <class 'sis.models.User'>, django_username_field = 'name'

    @pytest.fixture()
    def admin_user(db, django_user_model, django_username_field):
        """A Django admin user.
    
        This uses an existing user with username "admin", or creates a new one with
        password "password".
        """
        UserModel = django_user_model
        username_field = django_username_field
    
        try:
            user = UserModel._default_manager.get(**{username_field: 'admin'})
        except UserModel.DoesNotExist:
            extra_fields = {}
            if username_field != 'username':
                extra_fields[username_field] = 'admin'
            user = UserModel._default_manager.create_superuser(
>               'admin', 'admin@example.com', 'password', **extra_fields)
E           TypeError: create_superuser() got multiple values for keyword argument 'name'

UserModel  = <class 'sis.models.User'>
db         = None
django_user_model = <class 'sis.models.User'>
django_username_field = 'name'
extra_fields = {'name': 'admin'}
username_field = 'name'

../../../virtualenvs/sis/lib/python2.7/site-packages/pytest_django/fixtures.py:245: TypeError
================================== 1 error in 3.03 seconds ==================================

Compilation exited abnormally with code 1 at Thu Feb  2 09:41:48
```

This is because the name `admin` is passed as first argument and another time in the `**extra_fields`.

Either it should not be assumed that the username is the first argument (which seems safer) '''or''' the name should be dropped from `extra_fields`. My intuition is to prefer the first, i.e. only passing it in `extra_fields`.

TauPan:
This fixes that one issue

```python
@pytest.fixture()
def admin_user(db, django_user_model, django_username_field):
    """A Django admin user.

    This uses an existing user with username "admin", or creates a new one with
    password "password".
    """
    UserModel = django_user_model
    username_field = django_username_field

    try:
        user = UserModel._default_manager.get(**{username_field: 'admin'})
    except UserModel.DoesNotExist:
        extra_fields = {}
        if username_field != 'username':
            extra_fields[username_field] = 'admin'
[...]

PR #843, opened 2020-05-19
Make admin_user & admin_client fixtures compatible with custom user models

jnns:
Using the `admin_client` fixture with a custom user model that doesn't have a username field doesn't work currently. The fixture assumes that the user model always has a `username` field present. For this case, Django has `User.USERNAME_FIELD` and `User.get_username()`. 

Unfortunately, the existing test didn't catch this issue because a custom user model _with_ `User.username` is used. 

This pull request changes the test setup to use a custom user without `username` and makes `admin_client` use `User.get_username()`.

There are a few merge requests and issues about this problem already. I hope this one meets the requirements to be merged. If not, please let me know and I see what I can do.  

- https://github.com/pytest-dev/pytest-django/pull/246
- https://github.com/pytest-dev/pytest-django/pull/484
- https://github.com/pytest-dev/pytest-django/pull/748
- https://github.com/pytest-dev/pytest-django/issues/457

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -291,11 +291,12 @@ def admin_user(db, django_user_model, django_username_field):
     try:
         user = UserModel._default_manager.get(**{username_field: username})
     except UserModel.DoesNotExist:
-        extra_fields = {}
-        if username_field not in ("username", "email"):
-            extra_fields[username_field] = "admin"
         user = UserModel._default_manager.create_superuser(
-            username, "admin@example.com", "password", **extra_fields
+            **{
+                username_field: username,
+                "email": "admin@example.com",
+                "password": "password",
+            }
         )
     return user
 
@@ -306,7 +307,7 @@ def admin_client(db, admin_user):
     from django.test.client import Client
 
     client = Client()
-    client.login(username=admin_user.username, password="password")
+    client.login(username=admin_user.get_username(), password="password")
     return client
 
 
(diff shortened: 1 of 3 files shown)


pytest_django/fixtures.py
```diff
@@ -291,11 +291,12 @@ def admin_user(db, django_user_model, django_username_field):
...
     except UserModel.DoesNotExist:
-        extra_fields = {}
-        if username_field not in ("username", "email"):
-            extra_fields[username_field] = "admin"
         user = UserModel._default_manager.create_superuser(
-            username, "admin@example.com", "password", **extra_fields
+            **{
+                username_field: username,
+                "email": "admin@example.com",
+                "password": "password",
```
bluetech, line 298: In case `username_field` is email, it should override the `email` below, so should come after it (even though the values are the same, but just for good measure).

```suggestion
                "email": "admin@example.com",
                "password": "password",
                username_field: username,
```

bluetech (changes requested):
Thanks @jnns! This looks good to me and fixes what seems to be a long-standing issue.

If you could do the following that would be great:

1. Fix my one tiny comment.
2. Rebase the PR on latest master.
3. Add an entry to the changelog (`docs/changelog.rst`).
4. Edit the PR to say "Closes #XXX" on old the old PRs/issues that this fixes, so they're automatically closed once merged.

pytest_django/fixtures.py
jnns: Thanks for taking the time to review! I will add the requested changes and push again.

bluetech:
Hey @jnns, I am just about to release version 4.0.0 and want to include these changes, so I re-submitted this PR with the requested changes in #885. Thanks again!

PR #855, opened 2020-08-02
Run tests against Django 3.1

michael-k:
Django 3.1 is due on Monday 2020-08-03 if there isn't another RC release.
https://code.djangoproject.com/wiki/Version3.1Roadmap

This PR also includes two commits with unrelated changes.  They are necessary for the tests to pass.  (They are necessary to let the baseline stage pass on travis and actually run the new job.)

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -240,7 +240,7 @@ def _get_boolean_value(x, name, default=None):
     except KeyError:
         raise ValueError(
             "{} is not a valid value for {}. "
-            "It must be one of {}." % (x, name, ", ".join(possible_values.keys()))
+            "It must be one of {}.".format(x, name, ", ".join(possible_values.keys()))
         )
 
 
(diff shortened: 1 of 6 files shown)


pytest_django/plugin.py
```diff
@@ -239,7 +239,7 @@ def _get_boolean_value(x, name, default=None):
         return possible_values[x.lower()]
     except KeyError:
         raise ValueError(
-            "{} is not a valid value for {}. "
+            "{} is not a valid value for {}. "  # noqa: F502
             "It must be one of {}." % (x, name, ", ".join(possible_values.keys()))
```
michael-k, line 243: This doesn't work.  Earlier today I didn't really look at the code and just silenced the warning to unblock the PR.  But `{}` don't work in combination with `%s`.  This was introduced 4 years ago in commit https://github.com/pytest-dev/pytest-django/commit/3ec54d69e0ad14569c6d9e56d76ae7a29e772005#diff-87c0c2cb9262dd6c4384e140e51d8e64L165-R166 and nobody ever noticed, despite the TypeError being raised.

I'm going to change the `%` to `.format` which works with Python 2.7 and 3.2+, see https://pyformat.info/

bluetech: Hmm indeed, I stared at it when reviewing but missed that it used `%`. That makes more sense.

PR #857, opened 2020-08-14
Fix support for Django 3.2

untitaker:


pytest_django/compat.py
```diff
@@ -1,8 +1,29 @@
 # This file cannot be imported from until Django sets up
-try:
-    # Django 1.11
-    from django.test.utils import setup_databases, teardown_databases  # noqa: F401
-except ImportError:
+
+
+def _get_setup_and_teardown_databases():
```
untitaker, line 4: Refactor code such that we can use `return` to avoid deeply nested `try-except` blocks.

untitaker:
Not sure how the tests broke and I don't really understand them...

bluetech:
The tests should be fixed by #861, which is unrelated, so we can merge this. Thanks!

Issue #860, opened 2020-08-16
duplicate key value violates unique constraint "pg_extension_name_index" after update pytest-xdist to 2.0.0


MuslimBeibytuly:
pytest==6.0.1
pytest-asyncio==0.14.0
pytest-cov==2.10.1
pytest-benchmark==3.2.3
pytest-django==3.9.0
pytest-flake8==1.0.6
pytest-mock==3.2.0
pytest-xdist==2.0.0
pytest-bdd==3.4.0
python:3.8.5-alpine3.12
Django 3.1

Upgraded pytest-xdist from 1.34.0 to 2.0.0

tolomea:
Same thing here, also failing on mysql but with a different error.
Both errors kinda imply that per thread test db naming has broken, so all the threads are sharing a single db.
In the interim you can work around it by forcing single threaded or pinning pytest-xdist<2.0

tolomea:
The mysql error is ```MySQLdb._exceptions.OperationalError: (1050, "Table 'django_migrations' already exists")```

PR #861, opened 2020-08-16
support pytest-xdist 2

graingert:
Fixes https://github.com/pytest-dev/pytest-django/issues/860

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -46,7 +46,7 @@ def django_db_modify_db_settings_tox_suffix():
 def django_db_modify_db_settings_xdist_suffix(request):
     skip_if_no_django()
 
-    xdist_suffix = getattr(request.config, "slaveinput", {}).get("slaveid")
+    xdist_suffix = getattr(request.config, "workerinput", {}).get("workerid")
     if xdist_suffix:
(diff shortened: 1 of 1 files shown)


pytest_django/fixtures.py
```diff
@@ -46,7 +46,7 @@ def django_db_modify_db_settings_tox_suffix():
 def django_db_modify_db_settings_xdist_suffix(request):
     skip_if_no_django()
 
-    xdist_suffix = getattr(request.config, "slaveinput", {}).get("slaveid")
+    xdist_suffix = getattr(request.config, "workerinput", {}).get("workerid")
```
bluetech, line 49: I haven't followed the slave/worker work, but is it safe to just switch, or should we keep compatibility with the old names for a while? I.e. like this:

```suggestion
    workerinput = getattr(request.config, "workerinput", getattr(request.config, "slaveinput", {}))
    xdist_suffix = workerinput.get("workerid", workerinput.get("slaveid"))
```

graingert: pytest-cov switched wholesale, so I think pytest-django should too

bluetech: Right, looks like it should be fine from the graph provided in https://github.com/pytest-dev/pytest-cov/pull/412#discussion_r440068695.

Issue #864, opened 2020-08-25
Fixture for Async Client


ResleyR:
With Django 3.1 it would be useful to have an `async_client` fixture similar to the existing `client` fixture.          
 The following code works but would be nice if it was part the plugin itself

```python
@pytest.fixture()
def async_client():
    """A Django async test client instance."""
    skip_if_no_django()

    if get_django_version() < (3, 1):
        pytest.skip("no Django settings")
    from django.test.client import AsyncClient

    return AsyncClient()
```

This allows for testing async views when used along with [pytest-asyncio](https://pypi.org/project/pytest-asyncio/) with the alternative being using the django TestCase instead.

bluetech:
Fixed in #865, will be in 4.1 (along with `async_rf`).

The fixture as written doesn't do the automatic `pytest.skip` on django<3.1. Personally not sure automatic skip is a good idea...

PR #865, opened 2020-08-26
add async client fixture

pbelskiy:


pbelskiy:
Related to https://github.com/pytest-dev/pytest-django/issues/864

docs/helpers.rst
```diff
@@ -170,6 +170,22 @@ To use `client` as an authenticated standard user, call its `login()` method bef
...
+An instance of a `django.test.AsyncClient`_
+
+.. _django.test.AsyncClient: https://docs.djangoproject.com/en/dev/topics/testing/tools/#the-test-client
+
+Example
+"""""""
+
+::
+
+    async def test_with_async_client(async_client):
```
graingert, line 185: are we recommending people use:

```suggestion
    @asgiref.sync.async_to_sync
    async def test_with_async_client(async_client):
```
or

```suggestion
    @pytest.mark.asyncio
    async def test_with_async_client(async_client):
```

or something else?

pbelskiy: Yes! I forgot about it

graingert: I wonder if `@asgiref.sync.async_to_sync` would actually work better

pbelskiy: > I wonder if `@asgiref.sync.async_to_sync` would actually work better

So, is you have some standard? Or comparison of it?

graingert: It's listed in the Django docs

graingert: Like, if you're using a single database connection and a tx for each new test then when Django supports async database backends and pytest-asyncio makes a new loop, the connection will be lost

docs/helpers.rst
```diff
@@ -170,6 +170,23 @@ To use `client` as an authenticated standard user, call its `login()` method bef
         response = client.get('/private')
         assert response.content == 'Protected Area'
 
+``async_client`` - ``django.test.AsyncClient``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An instance of a `django.test.AsyncClient`_
+
+.. _django.test.AsyncClient: https://docs.djangoproject.com/en/dev/topics/testing/tools/#the-test-client
```
bluetech, line 178:
```suggestion
.. _django.test.AsyncClient: https://docs.djangoproject.com/en/stable/topics/testing/tools/#the-test-client
```

docs/helpers.rst
bluetech: When Django's test runner encounters an `async def` test case, it automatically wraps it in `asgiref.async_to_sync()`. I don't know if it does anything special that pytest-asyncio doesn't. On the other hand it might be better not to mix the two. Really I am not an expert on this, I prefer gevent myself.

bluetech (changes requested):
Thanks @pbelskiy, this looks good.

It will be good to rebase on latest master, the CI is overhauled and should be green now.

We need to figure out the recommendation. But if nothing clears up, we can stick with pytest-asyncio setup. It is just an example so can be fixed later.

I think for completion we should also add an `async_rf` fixture which creates an `AsyncRequestFactory`.

bluetech:
@pbelskiy You requested my review but not sure if you've seen my other comments. CI fails as well.

pbelskiy:
Rebased, tests is okay now 

@bluetech 
> I think for completion we should also add an async_rf fixture which creates an AsyncRequestFactory

Do you suggest to add this to this PR or new one?

docs/helpers.rst
```diff
@@ -168,7 +168,23 @@ To use `client` as an authenticated standard user, call its
         response = client.get('/private')
         assert response.content == 'Protected Area'
 
-.. fixture:: admin_client
```
bluetech, line 171: This line should remain, right above the `admin_client` title.

And above the `async_client` title, should add

```
.. fixture:: async_client
```

bluetech (commented):
> Do you suggest to add this to this PR or new one?

If you don't mind, would be good to do it in this PR already.

pbelskiy:
@bluetech Done. But I removed `:class:` prefix from docs for new fixtures:

An instance of a `django.test.AsyncClient`
An instance of a `django.test.AsyncRequestFactory`

Becase it leads to build documentation error, I don't understand why.

Issue #868, opened 2020-09-15
Drop support for Django<2.2, Python<3.5, pytest<5.4?


bluetech:
I propose we release a major version (4.0.0) which drops all of the unsupported combinations of the above.

In practice this means:

- Drop support for Django 1.8, 1.9, 1.10, 1.11, 2.0, 2.1. All of these versions are no longer [supported by Django](https://www.djangoproject.com/download/#supported-versions).

- Drop support for Python 2.7 and Python 3.4. No longer supported by Python and not supported by Django.

- Keep support for Python 3.5. While it's no longer supported by Python, it is supported by Django 2.2 LTS.

- Drop support for pytest 3.5 to 5.3.

Rationale:
- Reduce the huge CI matrix
- Drop compat hacks
- Be able to utilize a modern baseline

Users using Python 2 and 3.4 will automatically continue to get pytest-django 3. 

Users using older Django's or pytest's will need to upgrade their version, or require pytest-django<4.

bluetech:
Was fixed by #869. Will try to do a release soon.

PR #869, opened 2020-09-19
Drop support for Python<3.5, Django<2.2, pytest<5.4

### LINKED ISSUE #660: Import pathlib from pytest
As discussed on #636 when all supported versions of pytest offer an import of `pathlib`, use that rather than direct dependency on `pathlib2`

bluetech:
This is my proposal for version 4.0.0 of the plugin, per issue #868. The idea is to have a clean slate for further development in 4.0.0 by dropping support for unsupported Python, Django and pytest versions. Development for the 3.x series can continue on a separate branch if such need arises.

After this is merged I intend to also switch CI from Travis to github actions.

Closes #763.
Closes #796.
Closes #744.
Closes #660.


bluetech:
:tada: finally got the CI to pass with all the changes...

Issue #870, opened 2020-09-24
Order of setting and transactional_db


bdauvergne:
I use both in my tests and I created a media fixture (creating a temporary media directory and using the settings fixture) with autouse=True, due to alphabetic order it made the settings fixture run before the transactional_db fixture, but my post_migrate handler creating basic models depend upon the settings, in a test testing modification of those settings the teardown of transactional_db calling the flush() run with the modified settings instead of the default settings, modifying the default state of the database.

I think that a way of forcing run of setting before transactional_db must be found, as the post_migrate handler can depend upon the settings.

bluetech:
I tried to follow what you are saying but the paragraph is a bit long and hard to understand. Can you rephrase it such that a simpleton like me can understand? :smile: The deputy likes dots!

bdauvergne:
Simplification would be _all pytest-django Django fixtures can depend upon the settings in a way or another but the ordering between the settings fixtures and other fixtures is mainly random_. It would be useful that this order is fixed and documented. You can ignore my particular solution in the pull request, it just illustrates the problem. Note that it is a fundamental problem with the way pytest fixtures works, I think that without explicit dependencies between fixtures their ordering depend on their name or point of definition, I don't remember exactly. There is no way to indicate such fixture must run around another without depending on it.

PR #871, opened 2020-09-24
Force ordering of settings and transactional_db fixtures #870

bdauvergne:


pytest_django/fixtures.py
```diff
@@ -354,10 +354,13 @@ def finalize(self):
 
 
 @pytest.yield_fixture()
-def settings():
+def settings(request):
     """A Django settings object which restores changes after the testrun"""
     skip_if_no_django()
 
+    if 'transactional_db' in request.fixturenames:
```
bluetech, line 361: I suggest adding a comment explaining the rationale:

```suggestion
    # Order the `settings` fixture after DB.
    # We don't want overridden settings to be in effect during
    # DB setup/teardown/post_migrate.
    if 'transactional_db' in request.fixturenames:
```

bluetech (commented):
Thanks for the PR, I understand the issue now, and the change LGTM.

I suggested a comment, and this also needs a rebase against the latest master. Afterwards, if the CI is green, we can merge.

PR #885, opened 2020-10-16
Make admin_user & admin_client fixtures compatible with custom user models

### LINKED ISSUE #457: If user model changes the username field, admin_user fixture will pass it twice

bluetech:
Replaces PR #843 with some minor updates. Commits by @jnns.

---

Using the `admin_client` fixture with a custom user model that doesn't have a username field doesn't work currently. The fixture assumes that the user model always has a `username` field present. For this case, Django has `User.USERNAME_FIELD` and `User.get_username()`. 

Unfortunately, the existing test didn't catch this issue because a custom user model _with_ `User.username` is used. 

This pull request changes the test setup to use a custom user without `username` and makes `admin_client` use `User.get_username()`.

There are a few merge requests and issues about this problem already. I hope this one meets the requirements to be merged. If not, please let me know and I see what I can do.  

Closes #246
Closes #484
Closes #748
Fixes #457

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -293,11 +293,12 @@ def admin_user(db, django_user_model, django_username_field):
         # so we can assume so as well.
         user = UserModel._default_manager.get_by_natural_key(username)
     except UserModel.DoesNotExist:
-        extra_fields = {}
-        if username_field not in ("username", "email"):
-            extra_fields[username_field] = "admin"
         user = UserModel._default_manager.create_superuser(
-            username, "admin@example.com", "password", **extra_fields
+            **{
+                "email": "admin@example.com",
+                "password": "password",
+                username_field: username,
+            }
         )
     return user
 
--- a/pytest_django_test/settings_mysql_innodb.py
+++ b/pytest_django_test/settings_mysql_innodb.py
@@ -2,6 +2,7 @@ from os import environ
 
 from .settings_base import *  # noqa: F401 F403
 
+
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.mysql",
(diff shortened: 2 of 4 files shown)


bluetech:
It turns out mysql has some restriction on the max_size of unique indexes, so I lowered the EmailField max_size from 254 to 100.

PR #890, opened 2020-10-22
bit more explicit use case for `configure` + `settings` fixture

minusf:


docs/configuring_django.rst
```diff
@@ -85,6 +86,19 @@ This can be done from your project's ``conftest.py`` file::
...
+
+Settings can be overriden easily by using the ``settings`` fixture::
+
+    @pytest.fixture(autouse=True)
+    def use_dummy_cache_backend(settings):
+        settings.CACHES = {
+            "default": {
+                "BACKEND": "django.core.cache.backends.dummy.DummyCache",
+            }
+        }
```
bluetech, line 100:
```suggestion
Settings can be overridden by using the :fixture:`settings` fixture::

    @pytest.fixture(autouse=True)
    def use_dummy_cache_backend(settings):
        settings.CACHES = {
            "default": {
                "BACKEND": "django.core.cache.backends.dummy.DummyCache",
            }
        }

Here `autouse=True` is used, meaning the fixture is automatically applied to all tests,
but it can also be requested individually per-test.
```

bluetech (commented):
I left a suggestion, but LGTM, thanks!

PR #893, opened 2020-11-03
Add query count option

MRigal:
When you have many tests using the assert_num_queries fixtures, it can be tedious to identify the ones impacted by the change of a factory for example.

Now you can run `pytest --only-query-count` and only the tests using these fixtures will be run. (Also works on the pytest-django test suite itself)

@blueyed what do you think about it? Do you like it? Shall I add another option to exclude these tests? (I could see less use cases...). If you think it is a nice addition, I would submit a documentation change in a following commit on this branch


MRigal:
Please note that this somehow does not work properly with `--lf`. If running `--check-query-count` returns many errors, rerunning it with `--lf` will only rerun the first of the last failed tests... :-/

bluetech:
I think adding a full-blown command line argument for this is a bit too ad-hoc.

What you can do instead in your project is define your own fixture which wraps `django_assert_num_queries` but also applies a [pytest mark](https://docs.pytest.org/en/stable/mark.html) to the test. Then you can use `pytest -m` to select or deselect them.

MRigal:
@bluetech sorry for the late response. We have this in our repo, indeed.

However it requires the tests to be ran from exactly the place where the conftest lies, which is a (small) disadvantage.

I thought this could be beneficial for big projects with CI where tests are ran into batches, allowing to group by this kind of test. However I see your point of trying to avoid the plugin to bloat, at least without having a couple thumbs up from other people awaiting this feature ;-)

bluetech:
I still think it's too specialized and overlaps with pytest's markers and/or hooks functionality, so I'll close this now. Of course, we can still discuss if there are other ideas though.

Issue #297, opened 2015-12-07
Error message for database access in setup_X is misleading


pelme:
The error message when triggering the database from setup_X methods should clearly point out that database access from such methods is not supported, rather than advising to put the django_db marker on the test.

See http://stackoverflow.com/questions/34089425/django-pytest-setup-method-database-issue for original problem.

TauPan:
May I also suggest some documentation, similar to what @pelme has written in his very helpful SO answer. At least I couldn't find that setup-methods unsupported in the pytest-django documentation anywhere.

pelme:
Yes that would be very helpful. If you (or anyone else) are interested in contributing a documentation change based on my SO answer, that would be very helpful!

PR #896, opened 2020-12-15
[MINOR] add opt-in multi-database support for db fixtures

averagechris:
Adds opt-in multi-database support for database fixtures (db, transactional_db) to cleanup in databases other than `default`.

This approach uses a setting in `settings.DATABASES["db_name"]["TEST"]["PYTEST_DJANGO_ALLOW_TRANSACTIONS"]` to determine whether the database fixtures should create a transaction, or cleanup that database after each test. It does this by extending the collection of database names on the relevant django TestCase's `databases` attribute or by setting `multi_db=True` for compatibility with django version <=1.8.x.

TODO:
- [x] update documentation
- [ ] ask maintainers if they'd like these tests to run against other databases and add the DATABASES config if so
- [x] I found #431 and will be adjusting the approach based on the feedback in that PR

### DIFF
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -71,6 +71,7 @@ other than ``default`` by adding a special key to your database test configurati
 in your django settings.
 
 for example::
+
     DATABASES = {
         "default": {
             # ... normal default settings
@@ -82,6 +83,7 @@ for example::
             }
         }
     }
+
 
 With that database configuration in your test environment, tests that depend on the
 ``db`` or ``transactional_db`` fixtures can make changes in the "secondary" database
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -66,21 +66,31 @@ select using an argument to the ``django_db`` mark::
 Tests requiring multiple databases
 ----------------------------------
 
-Currently ``pytest-django`` does not specifically support Django's
-multi-database support.
+You can configure ``pytest-django`` to create transactions (or cleanup) databases
+other than ``default`` by adding a special key to your database test configuration
+in your django settings.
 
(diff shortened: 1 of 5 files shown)


bluetech:
Hi @mistahchris,

I don't have much experience with non-replica mutli-db but I'll try to find time to research the issues and look into your PR. Any more details/considerations you can add would be useful.

In the meantime, I noticed you have some code for `django versions <= 1.8.X`. Current pytest-django only supports Django>=2.2 so you don't need to worry about <=1.8.

mistahchris:
Hi @bluetech,

Thanks for taking the time to look into this PR! Let me know if I can help answer any questions. I've had to work with multiple databases quite a bit for work, which prompted me to make this PR in the first place.

I'll ditch the django 1.8.x stuff and fix the failing check on Monday or Tuesday this week (in next couple of days).

aijorgenson:
This PR may also indirectly resolve https://github.com/pytest-dev/pytest-django/issues/297

mistahchris:
Sorry about ghosting on this PR. My computer died and I just got my new one a few days ago. I'm working on setting up my python environment today so I can make the tweaks to this PR this weekend.

jgb:
@mistahchris any update on this PR? Would love to try it out.

bluetech:
So I'm going over all of the multi-db issues and PRs, trying to understand the material.

The approach you've taken here is to have a global setting, which affects all of the tests. I think this is too coarse -- one should be able to configure the `databases` on a per-test way, for efficiency and control. This is what the underlying Django `TestCase` provides. This means it should be configured on the mark/fixture.

I'll close this PR then, but I'll open an issue discussing multi-db support and will cc this.

PR #902, opened 2021-01-24
try pytest 6.2

auvipy:


bluetech:
Any special reason to add a special case for pytest 6.2.x? If `pytest54` is not specified, the latest version is run.

auvipy:
then this can be closed

PR #905, opened 2021-02-21
Add `request.user` to example

guettli:
Newcomers get very confused if `request.user` is not set. I think it makes sense to underline this in the example.

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -132,6 +132,10 @@ Example
 
     def test_details(rf, admin):
         request = rf.get('/customer/details')
+        # Remember that when using RequestFactory, the request does not pass
+        # through middleware. If your view expects fields such as request.user
+        # to be set, you need to set them explicitly.
+        # The following line sets request.user to an admin user.
         request.user = admin
         response = my_view(request)
         assert response.status_code == 200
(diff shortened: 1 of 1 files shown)


docs/helpers.rst
```diff
@@ -130,8 +130,9 @@ Example
 
     from myapp.views import my_view
 
-    def test_details(rf):
+    def test_details(rf, admin):
         request = rf.get('/customer/details')
+        request.user = admin
```
bluetech, line 135:
```suggestion
        # Remember that when using RequestFactory, the request does not pass
        # through middleware. If your view expects fields such as request.user
        # to be set, you need to set them explicitly.
        # The following line sets request.user to an admin user.
        request.user = admin
```

bluetech (commented):
Thanks! I left a suggestion which explains a bit the purpose of the line, what do you think?

docs/helpers.rst
benjaoming: This is a great addition :+1: 

I can vouch for that, have just spent some time finding out about how to use `django.contrib.messages` with RequestFactory.

guettli: Looks good. Thank you.

bluetech (approved):

PR #910, opened 2021-03-02
Disable atomic durability check on non-transactional tests

### LINKED ISSUE #909: Handle transaction.atomic(durable=True)
This argument will be introduced in Django 3.2: https://docs.djangoproject.com/en/3.2/topics/db/transactions/#controlling-transactions-explicitly

It's used to make sure that atomic blocks aren't nested and naturally this fails when running the test suite. The django `TestCase`-class handles this by setting `transaction.Atomic._ensure_durability = False` in `setUpClass` and setting it back to `True` on `tearDownClass`.

I'm not that familiar with the codebase but I wonder if it would make sense to adjust the non-transactional block in `_django_db_fixture_helper` with something like:
```python
transaction.Atomic._ensure_durability = False
def reset_durability():
    transaction.Atomic._ensure_durability = True
request.addfinalizer(reset_durability)
```

hannseman:
Fixes #909

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -147,6 +147,12 @@ def _django_db_fixture_helper(
             django_case = ResetSequenceTestCase
     else:
         from django.test import TestCase as django_case
+        from django.db import transaction
+        transaction.Atomic._ensure_durability = False
+
+        def reset_durability():
+            transaction.Atomic._ensure_durability = True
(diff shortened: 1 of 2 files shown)


bluetech (approved):
Thanks, LGTM. Django does this in `setUpClass`, but pytest-django doesn't really support class scope, so should be good.

hannseman:
Thanks @bluetech

PR #913, opened 2021-03-13
add some special characters to fix rendering of docs

fgebhart:
I know this is a tiny change and and maybe not worth a PR, but my OCD kicked in and forced me to raise this PR.

Also, after using pytest-django more and more I hope that this might be the first but not last PR and I wanted to get started somewhere.

Happy to receive feedback.


bluetech (approved):
Thanks! Always happy to get fixes.

The CI failure is unrelated, so this is good to go!

PR #917, opened 2021-03-20
Less opinionated wording

guettli:
I shorted the text to be less opinionated.

I disagreed to the previous text: the ORM is a tool like any other tool. If it helps you, then use it. The only reason not to use it, is that
it might slow your tests down.

### DIFF
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -5,9 +5,7 @@ Database creation/re-use
 access.  By default your tests will fail if they try to access the
 database.  Only if you explicitly request database access will this be
 allowed.  This encourages you to keep database-needing tests to a
-minimum which is a best practice since next-to-no business logic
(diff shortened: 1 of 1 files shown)


docs/database.rst
```diff
@@ -5,9 +5,7 @@ Database creation/re-use
 access.  By default your tests will fail if they try to access the
 database.  Only if you explicitly request database access will this be
 allowed.  This encourages you to keep database-needing tests to a
-minimum which is a best practice since next-to-no business logic
-should be requiring the database.  Moreover it makes it very clear
-what code uses the database and catches any mistakes.
+minimum which and it makes it very clear what code uses the database.
```
bluetech, line 8:
```suggestion
minimum which makes it very clear what code uses the database.
```

bluetech (approved):
Sure, no need to be too opinionated.

PR #930, opened 2021-05-08
Add initial experimental multi database support

bluetech:
This PR adds experimental support for multi-database configurations - refs #924.

Currently, we only add a `databases` argument to the `django_db` marker, which corresponds to the `databases` attribute of Django's `TransactionTestCase`/`TestCase`. We probably also want some fixture to be able to control this more nicely for an entire file, module, session etc, but that can be done later.

Since I'm not really sure about the API, I marked it as experimental for now. If it works out fine, we can stabilize it.

### DIFF
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -64,21 +64,25 @@ select using an argument to the ``django_db`` mark::
 Tests requiring multiple databases
 ----------------------------------
 
+.. caution::
+
+    This support is **experimental** and is subject to change without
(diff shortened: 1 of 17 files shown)


julianmoji:
EY @bluetech, I have a function test that access to a Mongo and Postgres, This is a mix between TransactionalTestCase and TestCase, but it seems to be that the entire test is wrapped in a TransactionalTestCase because the Postgres is flushed too, This should not happen since transaction=False is being specified in the marker

PR #933, opened 2021-05-24
fixtures: add django_capture_on_commit_callbacks fixture

### LINKED ISSUE #752: on_commit support
It would be nice if pytest-django had on_commit support that didn't require a TransactionTestCase, something like this maybe.
https://medium.com/gitux/speed-up-django-transaction-hooks-tests-6de4a558ef96

bluetech:
Similar to Django's `TestCase.captureOnCommitCallbacks`. Documentation
is cribbed from there.

Fixes #752.

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -425,6 +425,51 @@ Example usage::
             Item.objects.create('foo')
             Item.objects.create('bar')
 
+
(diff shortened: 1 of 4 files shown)


adamchainz:
Thanks for adding this!

PR #946, opened 2021-09-15
assertTemplateUsed return type

meshy:
Hi! Thank you for this invaluable library.

I recently added a test to my project that looked something like this:

```python
def test_template(client: Client) -> None:
    with assertTemplateUsed("app/template.html"):
        client.get(URL)
```

This caused a typing error from MyPy:

```
"assertTemplateUsed" does not return a value  [func-returns-value]
```

This is because the [type hints added for this function in `pytest_django/asserts.py`](https://github.com/pytest-dev/pytest-django/blob/cf6b229bce0c717bc590d8fd7cef281f3aa2c1fa/pytest_django/asserts.py#L100-L113) incorrectly indicate a return type of `None`.

I hope this change will fix the error. While I was there, I noticed that the return type of `assertTemplateNotUsed` was similarly incorrect, so I've addressed that too.

I realise that I have imported an underscore-prefixed class here, and that might be a cause of some concern. I notice the class has existed for around 10 years, so I assume changes to the import are unlikely. That said please let me know if you would prefer me to replace the return type with a more generic `typing.ContextManager`, or if there is another approach that I might not have considered.

For reference, here is where Django returns instances of these classes:

https://github.com/django/django/blob/a7be74d01775f862d18add17b10296c8583e35d8/django/test/testcases.py#L653

https://github.com/django/django/blob/a7be74d01775f862d18add17b10296c8583e35d8/django/test/testcases.py#L682

### DIFF
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -41,6 +41,7 @@ for assert_func in assertions_names:
 
 if TYPE_CHECKING:
     from django.http import HttpResponse
+    from django.test.testcases import _AssertTemplateUsedContext
 
     def assertRedirects(
         response: HttpResponse,
@@ -102,7 +103,7 @@ if TYPE_CHECKING:
         template_name: Optional[str] = ...,
         msg_prefix: str = ...,
         count: Optional[int] = ...,
-    ) -> None:
+    ) -> _AssertTemplateUsedContext:
         ...
 
     def assertTemplateNotUsed(
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -41,7 +41,7 @@ for assert_func in assertions_names:
 
 if TYPE_CHECKING:
     from django.http import HttpResponse
-    from django.test.testcases import _AssertTemplateUsedContext
+    from django.test.testcases import _AssertTemplateNotUsedContext, _AssertTemplateUsedContext
 
     def assertRedirects(
         response: HttpResponse,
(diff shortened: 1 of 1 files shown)


bluetech (commented):
Thanks @meshy, good catch.

Even though the types have been around for a long time, I am still uneasy about importing them. Also, since Django isn't typed it becomes Any anyway...

My suggestion is would be this:

- Change the incorrect `-> None:` to `:` (i.e. no annotation). This indicates "incomplete annotation". Alternatives are:
    - Use `ContextManager` - I don't like this because the real type has extra attributes that may be accesses and `ContextManager` will prevent that.
    - Copy/paste the type - probably a bad idea.

- Add a "pytest native" fixture for this functionality, like [`django_assert_num_queries`](https://pytest-django.readthedocs.io/en/latest/helpers.html#django-assert-num-queries) which we have.

WDYT?

meshy:
> Even though the types have been around for a long time, I am still uneasy about importing them.

Seems reasonable to me :)

> Change the incorrect -> None: to : (i.e. no annotation).

Good plan. I've pushed a commit that removes the annotation. Please let me know if you would like me to merge my commits into one.

> Add a "pytest native" fixture for this functionality

This sounds like a good idea, but I don't know if I will have the time to implement this soon. Will you be happy merging this PR without the new feature?

bluetech:
That's OK, thanks!

PR #949, opened 2021-10-13
Fix settings fixture leak by resetting mutated values

marcgibbons:
Mutated attributes (i.e. dictionaries, lists) on the `settings` fixture leak into subsequent tests. 

```python
# settings.py
MY_DICT = {'a': 1}

# test_settings.py
def test_setting(settings):
    assert settings.MY_DICT == {'a': 1}
    settings.MY_DICT['a'] = 2  # Mutate the dict


def test_setting_resets(settings):
    assert settings.MY_DICT == {'a': 1}  # Fails
```

Django's `override_settings` context manager re-assigns the setting and, the framework provides `modify_settings` to mutate dicts and lists without having to redefine them in their entirety. 

This PR adds a mechanism on `SettingsWrapper` which copies the original value of the setting on `__getattr__`. During the finalize step, an additional check is made to all attributes "checked out". If the present value no longer equals to the original, it is reset.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -1,4 +1,5 @@
 """All pytest-django fixtures"""
+import copy
 import os
 from contextlib import contextmanager
 from functools import partial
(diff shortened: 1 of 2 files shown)


bluetech:
Thanks @marcgibbons, this does seem like a bad gotcha.

Another problem with modifying the settings as in `settings.MY_DICT['a'] = 2` is that the `setting_changed` signal won't fire (done by `override_settings`/`modify_settings`).

I wonder if perhaps we can handle this. I was thinking that `__getattr__` could return a wrapper itself when the value is a mutable collection (dict/list/set), linking back to the top-level wrapper. And this would be done recursively. Maybe it's too much?

I'd guess Django was careful not to expose an interface like pytest-django does due to such problems...

PR #952, opened 2021-11-04
Add Python 3.10 to test matrix.

hramezani:


bluetech:
Thanks, seems the new job seems to pass (the failing job is unrelated).

PR #954, opened 2021-11-06
fix  DJANGO_SETTINGS_MODULE docs

Yassine-cheffai:
using point to sepcify the path to the settings file, instead of underscore


bluetech (approved):
Thanks!

PR #958, opened 2021-11-15
Add USE_TZ to test settings.

hramezani:
This patch will remove below warning for django main
```
RemovedInDjango50Warning: The default value of USE_TZ will change from False to True in Django 5.0. Set USE_TZ to False in your project settings if you want to keep the current default behavior.
```

### DIFF
--- a/pytest_django_test/settings_base.py
+++ b/pytest_django_test/settings_base.py
@@ -30,4 +30,4 @@ TEMPLATES = [
 
 DATABASE_ROUTERS = ['pytest_django_test.db_router.DbRouter']
 
-USE_TZ = False
+USE_TZ = True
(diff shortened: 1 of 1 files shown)


pytest_django_test/settings_base.py
```diff
@@ -29,3 +29,5 @@
 ]
 
 DATABASE_ROUTERS = ['pytest_django_test.db_router.DbRouter']
+
+USE_TZ = False
```
bluetech, line 33: If we're setting it, probably better to set it to `True`, that's the common/useful value.

```suggestion
USE_TZ = True
```

bluetech (approved):
Thanks!

PR #959, opened 2021-11-17
Drop Django 3.0 support.

hramezani:
Since Django 3.0 doesn't receive an update, we can also drop it's support.
https://www.djangoproject.com/download/


bluetech (approved):
Thanks!

PR #962, opened 2021-11-20
FIX: Configure django to skip migrations if --nomigrations flag is set

timthelion:
Not sure why, but `--nomigrations` is a no-op for me. At least this code works in Djang>=3.1 . It is also cleaner.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -180,8 +180,16 @@ def _django_db_fixture_helper(
 
 
 def _disable_native_migrations() -> None:
+    # https://github.com/django/django/commit/f5ebdfce5c417f9844e86bccc2f12577064d4bad
+    if get_django_version() >= (3, 1):
+        from django.db import connections
+        for alias in connections:
+            connection = connections[alias]
+            test_settings = connection.settings_dict['TEST']
+            test_settings['MIGRATE'] = False
+        return
+
     from django.conf import settings
-    from django.core.management.commands import migrate
 
     class DisableMigrations:
         def __contains__(self, item: str) -> bool:
@@ -191,6 +199,8 @@ def _disable_native_migrations() -> None:
             return None
 
     settings.MIGRATION_MODULES = DisableMigrations()
+
+    from django.core.management.commands import migrate
 
     class MigrateSilentCommand(migrate.Command):
         def handle(self, *args, **kwargs):
--- a/tests/test_db_setup.py
+++ b/tests/test_db_setup.py
@@ -456,6 +456,8 @@ class TestNativeMigrations:
     """ Tests for Django Migrations """
 
     def test_no_migrations(self, django_testdir) -> None:
+        skip_if_sqlite_in_memory()
+
         django_testdir.create_test_module(
             """
             import pytest
@@ -467,16 +469,30 @@ class TestNativeMigrations:
         """
         )
 
+        result_first = django_testdir.runpytest_subprocess("-v", "--reuse-db")
+        assert result_first.ret == 0
+
         migration_file = django_testdir.project_root.join(
             "tpkg/app/migrations/0001_initial.py"
         )
         assert migration_file.isfile()
         migration_file.write(
-            'raise Exception("This should not get imported.")', ensure=True
+            """
+from django.db import migrations
+
+def fail(*args):
+    raise Exception("This should not get run.")
+
+class Migration(migrations.Migration):
+    operations = [
+        migrations.RunPython(fail),
+    ]
+            """
+            , ensure=True
         )
 
         result = django_testdir.runpytest_subprocess(
-            "--nomigrations", "--tb=short", "-vv", "-s",
+            "--reuse-db", "--nomigrations", "--tb=short", "-vv", "-s",
         )
         assert result.ret == 0
         assert "Operations to perform:" not in result.stdout.str()


timthelion:
The test passes for me, no idea why it's broken here. The code works, unlike the code that passes the test making this an ironic example of when automated testing doesn't save you ;)

pytest_django/fixtures.py
```diff
@@ -180,8 +180,16 @@ class PytestDjangoTestCase(test_case_class):  # type: ignore[misc,valid-type]
 
 
 def _disable_native_migrations() -> None:
```
bluetech, line 182: (I removed "native" from here and elsewhere in master)

pytest_django/fixtures.py
```diff
@@ -180,8 +180,16 @@ class PytestDjangoTestCase(test_case_class):  # type: ignore[misc,valid-type]
...
+    # https://github.com/django/django/commit/f5ebdfce5c417f9844e86bccc2f12577064d4bad
+    if get_django_version() >= (3, 1):
+        from django.db import connections
+        for alias in connections:
+            connection = connections[alias]
+            test_settings = connection.settings_dict['TEST']
+            test_settings['MIGRATE'] = False
+        return
```
bluetech, line 190: The `MigrateSilentCommand` stuff still seems relevant, for silencing the `migrate` output.

tests/test_db_setup.py
```diff
@@ -456,6 +456,8 @@ class TestNativeMigrations:
     """ Tests for Django Migrations """
 
     def test_no_migrations(self, django_testdir) -> None:
+        skip_if_sqlite_in_memory()
```
bluetech, line 459: Why?

tests/test_db_setup.py
```diff
@@ -467,16 +469,30 @@ def test_inner_migrations():
...
 
+        result_first = django_testdir.runpytest_subprocess("-v", "--reuse-db")
+        assert result_first.ret == 0
+
         migration_file = django_testdir.project_root.join(
             "tpkg/app/migrations/0001_initial.py"
         )
         assert migration_file.isfile()
         migration_file.write(
-            'raise Exception("This should not get imported.")', ensure=True
```
bluetech, line 475: (changed this in master so can be written with nice indentation)

bluetech (commented):
Thanks @timthelion. While reviewing this I made some changes in master (see comments). You can find a rebased version here: https://github.com/bluetech/pytest-django/commit/34e5fbf8e36340818b51fcae73ab0035c3f540e7

While testing it, however, I see that the new test passes also before the change, so it seems that the nomigrations functionality works fine. So maybe the problem is something else?

In regards to the patch itself, I'm not really sure that disabling migrations by setting `MIGRATE = False` for each db is better than the current way which sets `MIGRATION_MODULES`. The `MIGRATION_MODULES` seems a bit cleaner actually. Is there a reason to prefer the `MIGRATE` approach?

pytest_django/fixtures.py
timthelion: I thought that the whole point of `_disable_native_migrations()` is that the migrations are disabled. Why would you want to silence them? Isn't the purpose of the `--nomigrations` flag to prevent migrations from running? If so, the the `migrate` command shouldn't run and there should be no output to suppress. What does the word `native` mean in this context? Is this only supposed to suppress some migrations but not all?

tests/test_db_setup.py
timthelion: When the code actually works, migrations are disabled. In order for the test to actually load, we need to first run migrations, and then run the tests with `--nomigrations`. But there is no way to save the migrated state of the db using `sqlite_in_memory` the db is not maintained from one run to the next.

timthelion:
Just to be clear, the reason that I created this PR is that I have a project with a LOT of migrations and it takes > 100 seconds to run them. So every time I run pytest I have to wait 100 seconds. The `--nomigrations` flag doesn't help. But with this PR  and `--nomigrations` my test run time goes down to 2 seconds, which is much more ergonomic. So to answer your question about reason for using `MIGRATE`, I investigated this and made the PR because the current approach does not work at all. The current approach doesn't speed up anything for me. I included a link in my PR to this django commit https://github.com/django/django/commit/f5ebdfce5c417f9844e86bccc2f12577064d4bad

The commit message is "Added an option to allow disabling of migrations during test database creation." It seems like the whole point of the `MIGRATE` flag is to do exactly what I want `--nomigrations` to do, except, for the fact that I'm using this in conjunction with `--reusedb` and therefore it isn't during database creation (I'm not clear when you would ever want to not run migrations during test db creation but perhaps it is useful for testing the migrations themselves.

bluetech:
Ah I understand, it's not that it doesn't *work*, just that it's less efficient since it still computes the migration graph. That makes more sense. So this looks like a good idea.

BTW, since Django now has added `MIGRATE`, perhaps we should deprecate  `--no-migrations`, since the user can now just specify it in their Django settings?

bluetech:
Actually, I understand the confusion now. The Django code has changed since the commit you linked - see https://github.com/django/django/commit/77caeaea888d1744416b213036ff29699758de76 (part of Django 3.2). The migrate command still runs now when `MIGRATE = False`, in order to sync the DB; in fact `MIGRATE = False` just sets `MIGRATION_MODULES = <empty>`, basically what pytest-django has been doing.

timthelion:
OK, so there seems to be a general misunderstanding between various parties. It seems that there are two use cases for disabling migrations.

1. To test migrations
2. To speed up test running when there is a large migration DAG

And the two parties are going back and forth, merging code over each others work, thinking that the other parties work is buggy.

That said, a flag `--no-migrations` doesn't make any sense for case `#1`. If you need to disable migrations in order to test migrations, then you should be doing this with a settings override in the test suit and NOT as a flag. Flags are temporary configuration, not permanent configuration. So `--no-migrations` should only be used for case `#2`.

The way I imagine using `--no-migrations` is always in conjunction with `--reuse-db`. The first time I run the test suit, I run without it, and the migrations are run. Subsequent runs of the test suit, however, can be "fast runs". In which migrations are no longer necessary.

For this workflow, setting `MIGRATION_MODULES = <empty>` doesn't do me any good at all, because there aren't any migrations to run. It is the gathering of migrations that is slow, painfully slow.

Maybe though, it would be better to solve `#2` in a different way.  After all right now I'm treating the symptom rather than the disease. Why should it take 100 seconds to run the `migrate` command when there are no migrations to be run?

What purpose do you see in the `--no-migrations` flag? Why would you want to "sync the modules", whatever that means?

bluetech:
To better understand the question, it would help to understand what is your intended workflow. Do you use `--reuse-db`? Do you use `--no-migrations` always, or only after the first run?

timthelion:
After the first run and always with `--reuse-db`. Does not doing any migrations without `--reuse-db` ever make sense?

bluetech:
Sorry, I see you actually answered these question already.

So: I think you should point this question to Django itself, since pytest-django merely uses Django's test infrastructure under the hood. Essentially, I believe what you want is for `./manage.py migrate --run-syncdb` with `MIGRATION_MODULES = <empty>` to skip migration importing/planning/whatever's the slow part before the migrations are run. Maybe it's possible, or maybe there's a reason why the slow parts needs to be done even in this case, I don't know myself.

Some other options for you:

- Reset your migrations - delete all migrations, create new initial ones, truncate the `django_migrations` table, fake-apply the new ones. Without hundreds of migrations, you'll probably no longer need to worry about any of this.

- Patch pytest-django to just outright stub out the `migrate` command. I think something like this would work (not tested, put in your conftest.py):

```py
def django_db_use_migrations(django_db_use_migrations: bool) -> bool:
    if not django_db_use_migrations:
        from django.core.management.commands import migrate
        
        class MigrateNoopCommand(migrate.Command):
            def handle(self, *args, **kwargs):
               pass
        
        migrate.Command = MigrateNoopCommand
```

timthelion:
I don't understand why you like `MIGRATION_MODULES = <empty>` it is so clearly wrong to me that there must be some sort of horrible misunderstanding. 

If I set `--nomigrations` **I don't want any migrations to run, period.** That's what the flag `nomigrations` means to me and it is what I want. I tried looking into it, and I managed to wrangle out some verbosity. When migrations are run this happens:

```
Operations to perform:
  Synchronize unmigrated apps: admin_numeric_filter, admin_tools, adminactions, admindocs, adminfilters, bootstrap4form, bootstrapform, chart_tools, colorfield, corsheaders, dbbackup, django_extensions, django_filters, django_grapesjs, django_nvd3, django_spaghetti, django_wysiwyg, djangobower, djcelery_email, drf_yasg, humanize, import_export, inline_static, markdown_deux, messages, mptt, multiselectfield, nested_admin, notifications_edit, polymorphic, rangefilter, raven_contrib_django, related_admin, rest_framework, sekizai, staticfiles, stdimage, theming, tinymce, wiki_macros
  Apply all migrations: account, admin, admin_tools_stats, advanced_filters, aklub, auth, computedfields, contenttypes, dashboard, django_celery_beat, django_nyt, events, flexible_filter_conditions, helpdesk, import_export_celery, import_export_celery_edit, interactions, menu, notifications, oauth2_manager, oauth2_provider, pdf_storage, pinax_invitations, pinax_teams, post_office, repolinks, reversion, sessions, sites, smmapdfs, smmapdfs_edit, thumbnail, wiki, wiki_attachments, wiki_images, wiki_notifications
Running pre-migrate handlers for application polymorphic
Running pre-migrate handlers for application django_grapesjs
Running pre-migrate handlers for application admin_tools_stats
Running pre-migrate handlers for application admin_tools
Running pre-migrate handlers for application theming
Running pre-migrate handlers for application menu
Running pre-migrate handlers for application dashboard
Running pre-migrate handlers for application admin
Running pre-migrate handlers for application auth
Running pre-migrate handlers for application contenttypes
Running pre-migrate handlers for application sessions
Running pre-migrate handlers for application sites
Running pre-migrate handlers for application django_nyt
Running pre-migrate handlers for application stdimage
Running pre-migrate handlers for application bootstrapform
Running pre-migrate handlers for application bootstrap4form
Running pre-migrate handlers for application django_extensions
Running pre-migrate handlers for application django_wysiwyg
Running pre-migrate handlers for application tinymce
Running pre-migrate handlers for application chart_tools
Running pre-migrate handlers for application post_office
Running pre-migrate handlers for application raven_contrib_django
Running pre-migrate handlers for application import_export_celery
Running pre-migrate handlers for application import_export_celery_edit
Running pre-migrate handlers for application rangefilter
Running pre-migrate handlers for application computedfields
Running pre-migrate handlers for application adminactions
Running pre-migrate handlers for application notifications
Running pre-migrate handlers for application djangobower
Running pre-migrate handlers for application advanced_filters
Running pre-migrate handlers for application interactions
Running pre-migrate handlers for application events
Running pre-migrate handlers for application aklub
Running pre-migrate handlers for application oauth2_manager
Running pre-migrate handlers for application helpdesk
Running pre-migrate handlers for application pdf_storage
Running pre-migrate handlers for application django_celery_beat
Running pre-migrate handlers for application djcelery_email
Running pre-migrate handlers for application nested_admin
Running pre-migrate handlers for application smmapdfs
Running pre-migrate handlers for application smmapdfs_edit
Running pre-migrate handlers for application repolinks
Running pre-migrate handlers for application mptt
Running pre-migrate handlers for application sekizai
Running pre-migrate handlers for application thumbnail
Running pre-migrate handlers for application wiki
Running pre-migrate handlers for application wiki_attachments
Running pre-migrate handlers for application wiki_notifications
Running pre-migrate handlers for application wiki_images
Running pre-migrate handlers for application flexible_filter_conditions
Running pre-migrate handlers for application oauth2_provider
Running pre-migrate handlers for application account
Running pre-migrate handlers for application pinax_invitations
Running pre-migrate handlers for application pinax_teams
Running pre-migrate handlers for application reversion
Synchronizing apps without migrations:
  Creating tables...
    Running deferred SQL...
Running migrations:
  No migrations to apply.
  Your models have changes that are not yet reflected in a migration, and so won't be applied.
  Run 'manage.py makemigrations' to make new migrations, and then re-run 'manage.py migrate' to apply them.
Running post-migrate handlers for application polymorphic
Running post-migrate handlers for application django_grapesjs
Running post-migrate handlers for application admin_tools_stats
Running post-migrate handlers for application admin_tools
Running post-migrate handlers for application theming
Running post-migrate handlers for application menu
Running post-migrate handlers for application dashboard
Running post-migrate handlers for application admin
Running post-migrate handlers for application auth
Running post-migrate handlers for application contenttypes
Running post-migrate handlers for application sessions
Running post-migrate handlers for application sites
Running post-migrate handlers for application django_nyt
Running post-migrate handlers for application stdimage
Running post-migrate handlers for application bootstrapform
Running post-migrate handlers for application bootstrap4form
Running post-migrate handlers for application django_extensions
Running post-migrate handlers for application django_wysiwyg
Running post-migrate handlers for application tinymce
Running post-migrate handlers for application chart_tools
Running post-migrate handlers for application post_office
Running post-migrate handlers for application raven_contrib_django
Running post-migrate handlers for application import_export_celery
Running post-migrate handlers for application import_export_celery_edit
Running post-migrate handlers for application rangefilter
Running post-migrate handlers for application computedfields
Running post-migrate handlers for application adminactions
Running post-migrate handlers for application notifications
Running post-migrate handlers for application djangobower
Running post-migrate handlers for application advanced_filters
Running post-migrate handlers for application interactions
Running post-migrate handlers for application events
Running post-migrate handlers for application aklub
Running post-migrate handlers for application oauth2_manager
Running post-migrate handlers for application helpdesk
Running post-migrate handlers for application pdf_storage
Running post-migrate handlers for application django_celery_beat
Running post-migrate handlers for application djcelery_email
Running post-migrate handlers for application nested_admin
Running post-migrate handlers for application smmapdfs
Running post-migrate handlers for application smmapdfs_edit
Running post-migrate handlers for application repolinks
Running post-migrate handlers for application mptt
Running post-migrate handlers for application sekizai
Running post-migrate handlers for application thumbnail
Running post-migrate handlers for application wiki
Running post-migrate handlers for application wiki_attachments
Running post-migrate handlers for application wiki_notifications
Running post-migrate handlers for application wiki_images
Running post-migrate handlers for application flexible_filter_conditions
Running post-migrate handlers for application oauth2_provider
Running post-migrate handlers for application account
Running post-migrate handlers for application pinax_invitations
Running post-migrate handlers for application pinax_teams
Running post-migrate handlers for application reversion
System check identified no issues (0 silenced).
```

A full minute of nothing useful. Squashing the migrations doesn't help.  Why in Dickens name would anyone ever want to run post-migrate handlers for a full minute when their DB is already migrated and set up. We want our tests to run fast, not slow.

My PR fixes things. You say that your current code is cleaner, but not only is that not true, your current code is wrong.

timthelion:
Nevermind, this is pointless because my code no longer fixes things in most recent django: https://code.djangoproject.com/ticket/33177

PR #963, opened 2021-11-21
Add test for mirror database

mkniewallner:
While taking a look at the project, I saw that there was no test ensuring that mirroring of the database worked as expected when setting `"MIRROR": "default"` option, so I thought it could be nice to add one.


bluetech (approved):
Good idea, thanks!

PR #967, opened 2021-11-23
Drop support for Python 3.5

hramezani:


graingert:
might be worth applying pyupgrade eg https://github.com/pytest-dev/pytest/blob/df9e94fbca59d981df78141f4beebf41affd8ad0/.pre-commit-config.yaml#L36-L40

hramezani:
Thanks @graingert. I've applied `pyupgrade` and updated my patch with code update in a separate commit.

bluetech:
Thanks @hramezani. I notice though that we still support Django 2.2 which supports Python 3.5, as well as pytest 5.4 which does. I think that to keep for the purpose of having a common baseline we should wait until Django 2.2 is no longer supported ([April 2022](https://www.djangoproject.com/download/#supported-versions)), then drop Django 2.2, pytest 5.4 and Python 3.5. WDYT?

hramezani:
@bluetech Yes, Django 2.2 supports Python3.5 but Python3.5 is no longer receiving updates.
In Jazzband we decided to remove Python3.5 support but I am fine to wait and drop the Python3.5 support after Django2.2.
Let's keep the PR open and I will refresh it again.

hramezani:
@bluetech I've refreshed the PR and added another commit to remove Django 2.2 support.
Please take another look whenever you have time.
Thanks!

tox.ini
```diff
@@ -24,6 +22,7 @@ deps =
     coverage: coverage-enable-subprocess
 
     pytest54: pytest>=5.4,<5.5
+    psycopg28: psycopg2-binary==2.8.6
```
bluetech, line 25: Why is this needed?

bluetech (approved):
Thanks!

I left a question, but feel free to merge.

tox.ini
hramezani: Seems we don't need it now. I can't remember why I added this line.
Removed!

hramezani:
@bluetech Thanks for your review.
I am not able to merge the PR because there are some github action test which has to be removed from pipeline.

bluetech:
I removed the outdated checks, but the coverage one still blocks, so I'll force merge myself.

PR #972, opened 2021-11-30
Wire up cls.setUpTestData for class level fixtures

### LINKED ISSUE #514: How to use django_db mark at session fixture?
I have a situation where I need to load huge db fixture which is created by a function. The fixture is needed in all api tests. So I made a session fixture at `conftest.py` which would do it. But the problem is `pytest` throws following exception even though I have marked `django_db`:

`E       Failed: Database access not allowed, use the "django_db" mark to enable it.
`
Below is my code snippet.

```python
from permission.helpers import update_permissions

pytestmark = [
        pytest.mark.django_db(transaction = True),]

@pytest.fixture(scope="session", autouse = True)
def permission(request):
        load_time_consuming_db_fixture()
```

paultiplady:
Reimplement the Django TestCase's setUpTestData. Since we are
already calling the setUpClass machinery from TestCase, it's a
simple step to arrange for PytestDjangoTestCase to call the real
test class's setUpTestData classmethod.

This allows us to defer to the existing Django lifecycle hook machinery,
and also use the Django 3.2 implementation for TestData which
uses a descriptor to handle re-loading model instances between tests.
(It does this by memoizing the pre-test instances, so this avoids
us having to add a DB transaction or refresh_from_db() to reset the
testData.)

Partially fixes #514

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -182,6 +182,21 @@ def _django_db_helper(
         if _databases is not None:
             databases = _databases
 
+    if VERSION >= (3, 2):
+        from django.test.testcases import TestData
+        # If the test class has defined a Django-style setUpTestData,
+        # arrange for it to be called by the DjangoTestCase classmethods that we're calling here.
+        if request.cls and hasattr(request.cls, 'setUpTestData'):
+            def wrappedSetUpTestData(klass):
+                pre_attrs = request.cls.__dict__.copy()
+                request.cls.setUpTestData()
+                for name, value in request.cls.__dict__.items():
+                    if value is not pre_attrs.get(name):
+                        setattr(request.cls, name, TestData(name, value))
+
+            # Re-bind our new classmethod to the class we're attaching it to using the function's descriptor.
+            PytestDjangoTestCase.setUpTestData = wrappedSetUpTestData.__get__(PytestDjangoTestCase, PytestDjangoTestCase)
+
     PytestDjangoTestCase.setUpClass()
     if VERSION >= (4, 0):
         request.addfinalizer(PytestDjangoTestCase.doClassCleanups)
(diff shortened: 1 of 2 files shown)


paultiplady:
WIP: So far I've just verified that unit tests run on Py 3.7, I want to get some directional feedback before spending time polishing this and getting everything green.

paultiplady:
I looked for solutions to the more general problem of "run session-scoped Django fixtures", but it turns out to be a bit challenging. The core problem is thusly:

## Session-scoped fixtures

To write the most "pytestian" fixtures, I'd want to write a session-scoped fixture, something like

```python
def item_1():
    item_1 = Item.objects.create()
    yield item_1
```

However, there are a couple problems:

1. This is session-scoped fixture, so it can't depend on the existing `db` function-scoped fixture. We'd need to have a new fixture to unblock the DB and do the testcase setup.

@blueyed suggested something like this:

```
def django_db_setup(django_db_setup, django_db_blocker):
        with transaction.atomic():  # XXX: could/should use `TestCase_enter_atomics` for multiple dbs
            load_time_consuming_db_fixture()
            yield
```

Which has the follow-on issue:

2. this unblocks the DB for the whole session; so if you have mixed "true UTs" (no DB) and integration tests, you can't rely on pytest-django's DB locking any more. (Maybe this isn't a blocking issue? Django itself doesn't attempt to do the DB blocking, so there's an argument that it could be worth giving up the DB blocker to gain session-scoped Django fixtures.

Also, 3. This does not let us use the state-of-the-art `TestData` caching from Django 3.2. If we wrap everything in a transaction, even if we get this to play nicely with a `TestCase` (non-`TransactionTestCase`), we still need some way to reload the model instances to their pre-test state. (Else we're back to Django pre-3.2, where the test `setUp` needs to call `obj.reload_from_db()` on each model instance. This last objection is not insurmountable, but note that we need to do some sort of registration of model instances, and arrange for the underlying objects to get refreshed somehow -- it's not clear how we'd actually figure out in a generic way what model instances were created.

Maybe I'm overcomplicating this though -- I'm not familiar with the pytest machinery, so it's very likely there's an approach I didn't think of.

PR #984, opened 2022-01-02
correct pytest.ini spelling

mjh-sakh:
No whitespaces allowed around `=`

### DIFF
--- a/docs/configuring_django.rst
+++ b/docs/configuring_django.rst
@@ -36,7 +36,7 @@ Example::
 Example contents of pytest.ini::
 
     [pytest]
-    DJANGO_SETTINGS_MODULE = test.settings
+    DJANGO_SETTINGS_MODULE=test.settings
 
 Order of choosing settings
 --------------------------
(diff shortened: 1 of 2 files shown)


bluetech:
> No whitespaces allowed around =

They are allowed - can you show the error you get?

mjh-sakh:
@bluetech, thanks for the attention to this PR. Please find error details at [StackOverflow](https://stackoverflow.com/questions/67936854/how-to-solve-pytestconfigwarning-unknown-config-option-django-settings-module/70553672). I had the very same problem with 'no spaces around' ini file resolving it.

hoefling:
I forked your repo, added spaces back and ran the tests in CI. [The settings are parsed just fine](https://github.com/hoefling/python-project-lvl4/runs/4852052600?check_suite_focus=true#step:8:35). BTW the reason for the error in the SO question you linked is just `pytest-django` not being installed. If you run `pip uninstall pytest-django` and then `pytest`, you should get the `Unknown config option: DJANGO_SETTINGS_MODULE` warning as well since the setting is not registered in any `pytest` plugin installed.

mjh-sakh:
@hoefling, thank you for the effort to reproduce and extra details!

PR #988, opened 2022-01-24
Add Django version to report header

### LINKED ISSUE #987: Add Django version to report header
When pytest is run, it displays all the versions of the platform and plugins in the report header. Currently the `pytest-django` report header only adds the Django settings and (if found) configuration from `django-configuration`. It would be very useful if `pytest-django` also included the version of Django.

I recently had to debug an issue where I was swapping between two versions of Django and, while it is a small thing, having this info when running the tests would have been a nice DX improvement.

This should be an easy patch to write. Opening this issue to gauge interest.

(I searched the issues for a similar request, but did not find any. Apologies if this has been brought up before.)

joshuadavidthomas:
closes #987

Gets the current Django version installed and adds it to the report header.

Since it just uses functions already available in `pytest-django` and `django`, I didn't add any additional tests, but I can if needed.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -43,7 +43,7 @@ from .fixtures import rf  # noqa
 from .fixtures import settings  # noqa
 from .fixtures import transactional_db  # noqa
 from .fixtures import validate_django_db
-from .lazy_django import django_settings_is_configured, skip_if_no_django
+from .lazy_django import django_settings_is_configured, get_django_version, skip_if_no_django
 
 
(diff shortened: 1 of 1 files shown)


joshuadavidthomas:
Oh yikes, I guess I should have ran the tests locally before opening the pull request. 🤦‍♂️ Need to clean up the failures before it can get merged.

joshuadavidthomas:
I've been swamped at work, so I have been unable to work any on fixing the failing tests. I'm going to close this PR for now until I can dedicate some time to it.

Issue #995, opened 2022-02-28
self-tests fail against Pytest 7


stanislavlevin:
1 test fails on migration to Pytest 7:
```
=================================== FAILURES ===================================
______________ test_django_project_found_invalid_settings_version ______________

django_testdir = <Testdir local('/usr/src/tmp/pytest-of-builder/pytest-0/test_django_project_found_invalid_settings_version0')>
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f2af96481f0>

    @pytest.mark.django_project(project_root="django_project_root", create_manage_py=True)
    def test_django_project_found_invalid_settings_version(django_testdir, monkeypatch) -> None:
        """Invalid DSM should not cause an error with --help or --version."""
        monkeypatch.setenv("DJANGO_SETTINGS_MODULE", "DOES_NOT_EXIST")
    
        result = django_testdir.runpytest_subprocess("django_project_root", "--version", "--version")
        assert result.ret == 0
>       result.stderr.fnmatch_lines(["*This is pytest version*"])
E       Failed: remains unmatched: '*This is pytest version*'

/usr/src/RPM/BUILD/python3-module-pytest-django-4.5.2/tests/test_manage_py_scan.py:121: Failed
----------------------------- Captured stdout call -----------------------------
running: /usr/src/RPM/BUILD/python3-module-pytest-django-4.5.2/.tox/py3/bin/python -mpytest --basetemp=/usr/src/tmp/pytest-of-builder/pytest-0/test_django_project_found_invalid_settings_version0/runpytest-0 django_project_root --version --version
     in: /usr/src/tmp/pytest-of-builder/pytest-0/test_django_project_found_invalid_settings_version0
This is pytest version 7.0.1, imported from /usr/lib/python3/site-packages/pytest/__init__.py
setuptools registered plugins:
  pytest-django-4.5.2 at /usr/src/RPM/BUILD/python3-module-pytest-django-4.5.2/pytest_django/plugin.py
```

https://docs.pytest.org/en/7.0.x/changelog.html#breaking-changes:
> [pytest#8246](https://github.com/pytest-dev/pytest/issues/8246): --version now writes version information to stdout rather than stderr.

PR #996, opened 2022-02-28
tests: Sync expected stream for Pytest' version

stanislavlevin:
https://docs.pytest.org/en/7.0.x/changelog.html#breaking-changes:
> [pytest#8246](https://github.com/pytest-dev/pytest/issues/8246): --version now writes version information to stdout rather than stderr.

Fixes: https://github.com/pytest-dev/pytest-django/issues/995

### DIFF
--- a/tests/test_manage_py_scan.py
+++ b/tests/test_manage_py_scan.py
@@ -118,7 +118,12 @@ def test_django_project_found_invalid_settings_version(django_testdir, monkeypat
 
     result = django_testdir.runpytest_subprocess("django_project_root", "--version", "--version")
     assert result.ret == 0
-    result.stderr.fnmatch_lines(["*This is pytest version*"])
+    if hasattr(pytest, "version_tuple") and pytest.version_tuple >= (7, 0):
+        version_out = result.stdout
+    else:
+        version_out = result.stderr
+
+    version_out.fnmatch_lines(["*This is pytest version*"])
 
     result = django_testdir.runpytest_subprocess("django_project_root", "--help")
     assert result.ret == 0


tests/test_manage_py_scan.py
```diff
@@ -118,7 +118,13 @@ def test_django_project_found_invalid_settings_version(django_testdir, monkeypat
...
     result = django_testdir.runpytest_subprocess("django_project_root", "--version", "--version")
     assert result.ret == 0
-    result.stderr.fnmatch_lines(["*This is pytest version*"])
+    if result.errlines:
+        version_out = result.stderr
+    else:
+        # Pytest 7.0+
+        version_out = result.stdout
+
+    version_out.fnmatch_lines(["*This is pytest version*"])
```
bluetech, line 127: I prefer to use an explicit pytest version check, since then when we drop support for pytest<7 it is much easier to find and remove no longer needed code.

```suggestion
    if hasattr(pytest, 'version_info') and pytest.version_info >= (7, 0):
        version_out = result.stdout
    else:
        version_out = result.stderr
    version_out.fnmatch_lines(["*This is pytest version*"])
```

bluetech (commented):
Thanks! I left a suggestion.

hramezani (approved):
Thanks @stanislavlevin :+1:

tests/test_manage_py_scan.py
hramezani: Yes, this is better! Thanks @bluetech

stanislavlevin: @bluetech, thank you for the suggestion! I thought about this.
I prefer to not rely on version-check but check for actual behavior. For future cleanup I added the comment in code. But of course, your pov as project's maintainer is higher than my own.

stanislavlevin: updated.
Note: `version_info` => `version_tuple` (requires Pytest 7.0+).

PR #997, opened 2022-02-28
Fix byte and string comparison

farooqaaa:
`response.content` is a `bytes` object and doing a string comparison throws an error


bluetech (approved):
Indeed, thanks!

Issue #1001, opened 2022-03-21
Provide Hook for custom django settings configuration


JulianFeinauer:
We use a setup where we do not use djangos "default" initialization but create our own settings via Djangos `settings.configure(xxx)`.
Currently this is not supported via pytest-django but a very minor change could allow such a functionality.

I think all that would be necessary is the possibility to call a callback before Line 351 (https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/plugin.py#L351).

This change should not have any side effect for existing projects and by default the behavior of the library would not be changed at all.

I will provide a suggestion in the next days for a MWE.

JulianFeinauer:
Hmmmm... looks like that... RTFM again, hm?
I will try to adopt my code : )

NiklasMerz:
I was just looking for a place to document your new feature and found
this that sounds like something very similar 😉

On March 21, 2022, GitHub ***@***.***> wrote:
> Hmmmm... looks like that... RTFM again, hm?
>  I will try to adopt my code : )
>

—
Reply to this email directly, view it on GitHub
<https://github.com/pytest-dev/pytest-django/issues/1001#issuecomment-
1073961182>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AA3LORGRVQTSBHP2FXQ3XZDVBCA4PANCNFSM5RHNN7VQ>.
You are receiving this because you are subscribed to this thread.Message
ID: ***@***.***>

JulianFeinauer:
Feedback here...
It seems that it is at the moment impossible to run a custom database initialization when using pytest-django together with pytest-xdist.
The worker-nodes are then initialized differently and in the `pytest_load_initial_conftests` Hook the "default" initialization is done for django.

So I think we have to stick to something like this approach

PR #1002, opened 2022-03-21
add DJANGO_CONFIGURATION_HOOK as optional parameter to allow customly…

JulianFeinauer:
… configured settings instances

This is a first draft implementation for #1001 to have a starting point for a discussion.

See here for how this would be applied: https://github.com/pragmaticindustries/open-machine-plattform/tree/feature/new-launcher-with-pytest-fixes
Especially https://github.com/pragmaticindustries/open-machine-plattform/blob/feature/new-launcher-with-pytest-fixes/pytest_setup.py
which is the hook that would be called and would internally call `settings.configure({...})` to prepare a valid `settings` singleton for later.


NiklasMerz:
I guess you can close this now? ;-)

JulianFeinauer:
I guess so : )

PR #1022, opened 2022-08-25
Add Django 4.1 support

adamchainz:


.github/workflows/main.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install tox==3.24.4
+          pip install tox==3.25.1
```
adamchainz, line 51: upgrading tox as well here

.github/workflows/main.yml
```diff
@@ -76,24 +80,32 @@ jobs:
           python: '3.10'
           allow_failure: false
 
-        - name: py39-dj32-postgres-xdist-coverage
+        - name: py39-dj41-mysql_innodb-coverage
```
adamchainz, line 83: did a little rearranging of order so these are always in descending order by python version, then by django version - they were previously inconsistent

adamchainz (commented):
Tests show no new warnings 👍

hramezani (approved):

.github/workflows/main.yml
```diff
@@ -76,24 +80,32 @@ jobs:
...
           allow_failure: false
 
-        - name: py39-djmain-sqlite-coverage
-          python: 3.9
+        - name: py38-dj32-sqlite-xdist-coverage
+          python: 3.8
+          allow_failure: false
+
+        - name: py310-djmain-sqlite-coverage
+          python: '3.10'
```
samamorgan, line 108: Shouldn't this be a float, not a string?

adamchainz: No because then it will try to install Python 3.1, which is long dead

bluetech (approved):
Thanks!

adamchainz:
Thank you for merging.

Would you be open to adding me as a maintainer?

bluetech:
Happy to, you should have received an invite.

auvipy:
hope new maintainer can work towards a new release!

adamchainz:
I'm in, thanks! And 👋 @pelme as other maintainer.

adamchainz:
Accepted, thanks!

PR #1024, opened 2022-09-06
Feature/verbose liveserver

mpasternak:
Basing on my usage and on StackOverflow questions, the world needs ``--liveserver-verbose`` command line argument, which this branch implements. 

Default live server is quiet. Django guys made it quiet. And, well, for some, maybe this is good. Because theory says that you're supposed to have a lot of unit tests and when you start integration testing all your code should be A1-OKAY and 100% tested... in an ideal world. 

... but in this world, I tend to stumble on "500 ERROR" spewed by LiveServer at my code. Unable to debug, unable to print stuff to the console. Helpless, really. Especially helpless, when I know that everything in the code is okay and the error is because of environment problems. But still, unable to debug it, I feel helpless. 

And I don't like the feeling of being helpless. 

So I present to you the ``--liveserver-verbose`` argument. 

It prints stuff on the stdout!

It can even print out a backtrace. 

Is it ugly? Yes. 

Can it be improved? Yes! Just like pytest-durations, it would be lovely to have some frames and a nicely formatted output. 

But... I almost pressed "CREATE PULL REQUEST" here... I can improve it already. 

Another option I present to you is ``liveserver-debug``. This option drops to debugger in the console when a traceback occurs. 

Am I cool now? Am I the good guy? Can I take constructive criticism? Was I able to do all of that without switches on the command line? LET'S SEE!

What I would really like to see with this branch is how to collect those stdout texts nicely. If I don't use ``-s`` switch it prints out stderr on failing tests nicely. Perhaps you can help me improve this. Thanks!

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+vunreleased ()
+--------------
+
+Improvements
+^^^^^^^^^^^^
+
+* ``--liveserver-verbose`` command-line parameter, which allows to see
+  logs from live_server on the standard output, including tracebacks.
+  Useful for debugging.
+
 v4.5.2 (2021-12-07)
 -------------------
 
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -374,6 +374,10 @@ created in data migrations, you should add the
 You can also use ``--liveserver-verbose`` command-line argument, to outputs
 the liveserver logs to the standard output, including tracebacks. This is
 useful for debugging live server behaviour and environment-related problems.
+Take note, logs will be written to standard output, which is being supressed
+by pytest by default, so you will need to use ``-s`` or ``--capture=no``
+parameter too.
+
 
 .. note:: Combining database access fixtures.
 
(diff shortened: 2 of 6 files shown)


mpasternak:
Looks like I can't request review. @bluetech @hramezani I saw you helped with a few recent merges, would you like to check it out?

mpasternak:
Review required!

AngelOnFira:
Thanks for doing this @mpasternak, I hope it gets merged because it does just work :)

I did just try using this locally, and I ran into a problem with the following code:
```python
class VerboseLiveServerThread(LiveServerThread):
    request_handler_class = VerboseWSGIRequestHandler

    def _create_server(self):
        return ThreadedWSGIServer(
            (self.host, self.port),
            self.request_handler_class,
            allow_reuse_address=False
        )
```
due to this commit to Django: https://github.com/django/django/commit/855f5a36e7c8e7a8ce3f62d6ef8c9ae3e073ae3d making a change to how _create_server is called:
```python
self.httpd = self._create_server(
    connections_override=self.connections_override,
)
```
I fixed it by just adding an optional argument:
```python
def _create_server(self, connections_override={}):
```

I don't know if this is the cleanest fix, but it works for me right now :)

mpasternak:
Thanks for some kind words. It’s been almost a year since I posted this. Obviously nobody runs into problems with live server except me. And maybe some people on StackOverflow 🤣

Issue #294, opened 2015-12-02
live_server unable to serve static files under Django 1.8.7


bobsilverberg:
I've been trying to get a test running using selenium and the `live_server` fixture, and I couldn't seem to get the server to serve up static files. I compared the code in `LiveServerHelper.__init__` [1] to the code for `LiveServerTestCase` in Django [2] and it looked like the problem was with the StaticFilesHandler that was being passed in to `LiveServerThread`.

I changed the code locally to remove the path that is specific to Django >= 1.7 [3] and that seems to have solved the problem for me.

Is this in fact a bug, or am I just doing something wrong?

[1] https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/live_server_helper.py#L14
[2] https://github.com/django/django/blob/stable/1.8.x/django/test/testcases.py#L1214
[3] https://github.com/pytest-dev/pytest-django/blob/master/pytest_django/live_server_helper.py#L29-L35

pelme:
@bobsilverberg Do you have `django.contrib.staticfiles` in INSTALLED_APPS?

The idea is that you should get static files with the `live_server` fixture automatically, as long as staticfiles is installed.

bobsilverberg:
@pelme Yes, I do have `django.contrib.staticfiles` in `INSTALLED_APPS`, and with the current code it won't serve static files for me. It only works if I comment out that code path and allow it to use `_StaticFilesHandler` from `django.test.testcases` instead. I assume the current code has been tested with Django 1.8.7? What is that other code path supposed to buy me?

bobsilverberg:
FTR, a colleague of mine was able to reproduce the same results with my codebase.

pelme:
The idea is that you should be able to use the `live_server` fixture and have static files just work. With standard Django testcases you have to specifically subclass from `django.contrib.staticfiles.StaticLiveServerTestCase` if you use staticfiles and not want to run collectstatic manually. We look in INSTALLED_APPS and enable static files for live_server.

The relevant discussion (which is a bit verbose) for this can be found in issue #120 and PR #140 . There is a test case that verifies that this should be working:

https://github.com/pytest-dev/pytest-django/blob/d28ecf2e48cb3f58509e5f4a8ea81a75ac6c3872/tests/test_fixtures.py#L148-L187

I created an empty project on my computer and got working, both with Django 1.8.7 and 1.9 and got it working. Can you please try to create a minimal project and try to reproduce your problem in it?

```
[...]

PR #1032, opened 2022-10-06
Move thread activation to start method

samamorgan:
This moves activation of `LiveServer.thread` to a method. My primary motivation here is to delay thread activation so I can change the parameters of the thread in an overriden `live_server` fixture. This provides a work-around to issue #294 using the following override:

```python
import os

import pytest
from django.test.testcases import _StaticFilesHandler
from pytest_django import live_server_helper
from pytest_django.lazy_django import skip_if_no_django


def live_server(request):
    skip_if_no_django()

    addr = (
        request.config.getvalue("liveserver")
        or os.getenv("DJANGO_LIVE_TEST_SERVER_ADDRESS")
        or "localhost"
    )

    server = live_server_helper.LiveServer(addr)
    server.thread.static_handler = _StaticFilesHandler  # Force _StaticFilesHandler
    server.start()
    request.addfinalizer(server.stop)

    return server
```

### DIFF
--- a/pytest_django/live_server_helper.py
+++ b/pytest_django/live_server_helper.py
@@ -8,7 +8,7 @@ class LiveServer:
     The ``live_server`` fixture handles creation and stopping.
     """
 
-    def __init__(self, addr: str) -> None:
+    def __init__(self, addr: str, *, start: bool = True) -> None:
         from django.db import connections
         from django.test.testcases import LiveServerThread
         from django.test.utils import modify_settings
@@ -20,8 +20,6 @@ class LiveServer:
             # If using in-memory sqlite databases, pass the connections to
             # the server thread.
             if conn.vendor == "sqlite" and conn.is_in_memory_db():
-                # Explicitly enable thread-shareability for this connection.
-                conn.inc_thread_sharing()
                 connections_override[conn.alias] = conn
 
         liveserver_kwargs["connections_override"] = connections_override
@@ -51,6 +49,16 @@ class LiveServer:
         # `_live_server_helper`.
 
         self.thread.daemon = True
+
+        if start:
+            self.start()
+
+    def start(self) -> None:
+        """Start the server"""
+        for conn in self.thread.connections_override.values():
+            # Explicitly enable thread-shareability for this connection.
+            conn.inc_thread_sharing()
+
         self.thread.start()
         self.thread.is_ready.wait()
 


samamorgan:
Note: The failing tests here aren't failing, it's just an issue with the codecov upload.

bluetech (changes requested):
I'm not opposed to this, but needs a couple of changes:

1. To keep backward compat (not that it's a public interface, but who knows), let's add a `*, start: bool = True` argument to `__init__`, and add a call to `self.start` if `start` is true. You will then need to pass `start=False`.
2. The `inc_thread_sharing` part needs to be done in `start`, not in `__init__`, to match the `stop`.

samamorgan:
> I'm not opposed to this, but needs a couple of changes:
> 
> 1. To keep backward compat (not that it's a public interface, but who knows), let's add a `*, start: bool = True` argument to `__init__`, and add a call to `self.start` if `start` is true. You will then need to pass `start=False`.
> 2. The `inc_thread_sharing` part needs to be done in `start`, not in `__init__`, to match the `stop`.

Thanks for the quick feedback! I did both of these, then went a little ham and took it a step beyond.

This code was a little smelly. Imports inside the constructor, messing around with threads, etc. I refactored it to move all imports to the top, and it's now subclassing `LiveServerThread` rather than implementing a thread internally. This is functionally the same, but cleans up a lot of unnecessary calls to `self.thread`/`live_server.thread`. The class behaves like a thread, so why not just make it one?

If I went too hard here, I'm happy to back off and just implement the requested changes.

samamorgan:
Alright, tests failing now. Looks like I'm misunderstanding something about these imports. Is there an explanation I missed somewhere about why imports are happening in a function instead of the top of the module?

bluetech:
importing `django` has some side-effects, sometimes it requires `DJANGO_SETTINGS_MODULE` and such to be defined, etc. So it must always be imported lazily in pytest-django.

samamorgan:
Backed off on the subclassing decision. I spent _way too much time_ trying to implement a lazy import mechanism just to have it fall on its face.

I very much dislike the imports inside methods. Makes my heart hurt. I don't know enough about this library to really understand what a good alternative would be though, so the code smells get to stay!

samamorgan:
@bluetech Anything blocking this one?

pytest_django/fixtures.py
```diff
@@ -538,6 +538,7 @@ def live_server(request):
 
     server = live_server_helper.LiveServer(addr)
     request.addfinalizer(server.stop)
+
```
bluetech, line 541:
```suggestion
```

pytest_django/live_server_helper.py
```diff
@@ -1,56 +1,59 @@
...
+        from django.test.testcases import LiveServerThread, _StaticFilesHandler
         from django.test.utils import modify_settings
 
-        liveserver_kwargs = {}  # type: Dict[str, Any]
+        try:
+            host, port_str = addr.split(":")
+            port = int(port_str)
+        except ValueError:
+            host = addr
+            port = 0
```
bluetech, line 20: I think the entire `liveserver_kwargs` business was to avoid this line, i.e. not duplicate Django's default in case it changes.

bluetech:
Thanks @samamorgan. I rebased and removed unrelated changes from the diff. Merging now.

Issue #993, opened 2022-02-22
_dj_autoclear_mailbox ends quite often with AttributeError: module 'django.core.mail' has no attribute 'outbox'


mcepl:
When running the test suite for [whitenoise](https://github.com/evansd/whitenoise) plenty of tests ends with the following error (using pytest-6.2.5, pytest-django-4.5.2):

```
[   31s] _______ ERROR at setup of test_last_modified_not_set_when_mtime_is_zero ________
[   31s]
[   31s]     @pytest.fixture(scope="function", autouse=True)
[   31s]     def _dj_autoclear_mailbox() -> None:
[   31s]         if not django_settings_is_configured():
[   31s]             return
[   31s]
[   31s]         from django.core import mail
[   31s]
[   31s] >       if mail.outbox:
[   31s] E       AttributeError: module 'django.core.mail' has no attribute 'outbox'
[   31s]
```

When reading on the `mail.outbox` I found that it [doesn’t exist quite often](https://stackoverflow.com/q/5424498/164233), so this fixture has protection against it. When I apply this patch:
```diff
---
 pytest_django/plugin.py |    3 ++-
 1 file changed, 2 insertions(+), 1 deletion(-)

--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -502,7 +502,8 @@ def _dj_autoclear_mailbox() -> None:

     from django.core import mail

-    del mail.outbox[:]
+    if hasattr(mail, "outbox"):
+        del mail.outbox[:]


 @pytest.fixture(scope="function")
```
Tests pass without a problem.

hramezani:
Based on [Django documentation](https://docs.djangoproject.com/en/dev/topics/testing/tools/#django.core.mail.django.core.mail.outbox):

_During test running, each outgoing email is saved in django.core.mail.outbox. This is a list of all [EmailMessage](https://docs.djangoproject.com/en/dev/topics/email/#django.core.mail.EmailMessage) instances that have been sent. The outbox attribute is a special attribute that is created only when the locmem email backend is used. It doesn’t normally exist as part of the [django.core.mail](https://docs.djangoproject.com/en/dev/topics/email/#module-django.core.mail) module and you can’t import it directly._

Could you please add a test to show in what condition the above mentioned error happens?

mcepl:
The above error is from running test suite of https://github.com/evansd/whitenoise

mcepl:
@evansd Does this whole ticket make sense to you?

[...]

PR #1033, opened 2022-10-07
fix: don't crash if mail.outbox doesn't exist

symwell:
Fixes https://github.com/pytest-dev/pytest-django/issues/993

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -502,7 +502,8 @@ def _dj_autoclear_mailbox() -> None:
 
     from django.core import mail
 
-    del mail.outbox[:]
+    if hasattr(mail, "outbox"):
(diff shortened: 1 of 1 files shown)


samamorgan (commented):
Change looks good, needs a test though!

symwell:
It's unlikely I'll write a test.

mskoenz:
This fix would be helpful, I currently use the workaround from @wouter-vdb
```python
# see: https://github.com/pytest-dev/pytest-django/issues/993
@pytest.fixture(scope="function", autouse=True)
def _dj_autoclear_mailbox() -> None:
    pass
```

bluetech:
Thanks for the PR. This issue has been fixed in a separate PR #1187 so I'll close this one.

PR #1036, opened 2022-11-08
Fix assert function signatures that expect HTTP response

mschoettle:
Updating to the latest version(s) of `django-stubs` from `1.10.1` reveals errors such as:

> Argument 1 to "assertContains" has incompatible type "_MonkeyPatchedWSGIResponse"; expected "HttpResponse"  [arg-type]

`django-stubs` expects a `HttpResponseBase` instead of `HttpResponse` (see: https://github.com/typeddjango/django-stubs/blob/0381fdb0227f0db96a584d8e916ae6d8b3d025a3/django-stubs/test/testcases.pyi#L102)

This PR changes the signatures to be in line with what `django-stubs` defines.

### DIFF
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -40,10 +40,10 @@ for assert_func in assertions_names:
 
 
 if TYPE_CHECKING:
-    from django.http import HttpResponse
+    from django.http.response import HttpResponseBase
 
     def assertRedirects(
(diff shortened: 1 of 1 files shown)


adamchainz:
Great, thanks!

mschoettle:
@adamchainz Any chance a new version could be released soon that includes this fix?

PR #1037, opened 2022-12-04
Add comment about `assertTemplateUsed`

belegnar:


belegnar:
I just can't get why linter is failed

bluetech:
This part of the documentation just points to Django docs, so I don't think we should add this text here. It would be better to submit this change to Django.

PR #1040, opened 2022-12-16
Add `pattern` argument to django_assert_num_queries fixture

qarlosh:
This optional argument allows to filter the queries to take into account for the assertion.

It may be useful to check for the count of only certain queries. For example, a use case of mine is to skip SAVEPOINT queries by using:
```
NO_SAVEPOINT_PATTERN=r'(?!^(ROLLBACK TO |RELEASE )?SAVEPOINT [`"].+[`"]$)'
...
with django_assert_num_queries(n, pattern=NO_SAVEPOINT_PATTERN):
   ...
```

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -412,6 +412,7 @@ Example
   :param num: expected number of queries
   :param connection: optional non-default DB connection
   :param str info: optional info message to display on failure
+  :param str pattern: optional regex pattern to filter queries to take into account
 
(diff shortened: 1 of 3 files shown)


bluetech:
It will be better to propose this change to [Django's version](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) first.

PR #1053, opened 2023-03-22
Update database.rst

theilgaard:
Avoid confusion around single tests and test runs/sessions.


bluetech:
The transactions being referred to in this sentence actually do isolate per test, not per test session.

PR #1060, opened 2023-04-19
Fix verbose configuration option value check

stephane:
This commit corrects the check for the 'verbose' configuration option value. Previously, it was checked if the value was strictly greater than 0, but the use of '-v' flag sets the value to 0.

**Edit**: My original patch changed to `verbose = config.getoption("verbose") >= 0` in `fixtures.py`.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -600,7 +600,7 @@ def _assert_num_queries(
                 sqls = (q["sql"] for q in context.captured_queries)
                 msg += "\n\nQueries:\n========\n\n" + "\n\n".join(sqls)
             else:
-                msg += " (add -v option to show queries)"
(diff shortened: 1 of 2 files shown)


stephane:
May be the right fix is to change the message "(add -v option to show queries)" to "(add -vv option to show queries)"

bluetech:
> but the use of '-v' flag sets the value to 0.

I may be confused but I'm pretty sure passing `-v` sets verbosity to 1, not 0. If so, then the current code is correct. Perhaps you are passing `-q` somewhere? A `-q` reduces the verbosity by 1, so `-q -v` ends up 0.

stephane:
I didnt use other flags...

bluetech:
Maybe it's in your pytest config or in PYTEST_ADDOPTS envvar.

PR #1064, opened 2023-05-12
Add a django_set_urlconf fixture

orf:
Currently overriding the Django urlconf requires a marker, which also requires defining the urlconf at the root of the test module. If you have many tests that need this (with differing patterns and views) this can get quite annoying. 

This doesn't seem to be necessary, as the Django urlconf setting can be an object and it can be changed at runtime.

This PR adds a `django_set_urlconf` fixture that allows you to define the urlconf within the test and set it directly with the fixture:

```python
def test_django_set_urlconf(django_set_urlconf, client) -> None:
    def my_view(request):
        return http.HttpResponse(status=200, content="Success!")

    urlpatterns = (urls.path("test/view", my_view, name="test_view"),)
    django_set_urlconf(urlpatterns)
    assert b'Success!' in client.get("test_view").content
```

### DIFF
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -542,7 +542,7 @@ Example
     from django import http, urls
 
     def test_can_override_a_root_urlconf(django_set_urlconf):
-        def my_view(request: http.request.HttpRequest) -> http.response.HttpResponse:
+        def my_view(request):
             raise http.Http404()
 
         my_urlpatterns = (urls.path("test/view", my_view, name="test_view"),)
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -545,8 +545,9 @@ Example
         def my_view(request):
             raise http.Http404()
 
-        my_urlpatterns = (urls.path("test/view", my_view, name="test_view"),)
-        django_set_urlconf(my_urlpatterns)
+        django_set_urlconf([
+            urls.path("test/view", my_view, name="test_view")
+        ])
         assert urls.reverse("test_view") == "/test/view"
 
 Automatic cleanup
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -108,7 +108,8 @@ dynamically in a hook or fixture.
 
 .. py:function:: pytest.mark.urls(urls)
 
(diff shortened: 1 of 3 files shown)


docs/helpers.rst
```diff
@@ -529,6 +529,26 @@ the ``django_mail_dnsname`` fixture, which defaults to
...
+
+Example
+"""""""
+
+::
+
+    from django import http, urls
+
+    def test_can_override_a_root_urlconf(django_set_urlconf):
+        def my_view(request: http.request.HttpRequest) -> http.response.HttpResponse:
```
adamchainz, line 545: drop the type hints, since most users don't use them
```suggestion
        def my_view(request):
```

docs/helpers.rst
```diff
@@ -529,6 +529,26 @@ the ``django_mail_dnsname`` fixture, which defaults to
...
+            raise http.Http404()
+
+        my_urlpatterns = (urls.path("test/view", my_view, name="test_view"),)
+        django_set_urlconf(my_urlpatterns)
```
adamchainz, line 549: A list is more typical for `urlpatterns`, and using inline is maybe a little clearer?

```suggestion
        django_set_urlconf([
            urls.path("test/view", my_view, name="test_view")
        ])
```

docs/helpers.rst
```diff
@@ -529,6 +529,26 @@ the ``django_mail_dnsname`` fixture, which defaults to
 "fake-tests.example.com".
 
 
+``django_set_urlconf``
+~~~~~~~~~~~~~~~~~~~~~~
+
+This fixture allows setting the ``settings.ROOT_URLCONF`` to a value created within the test.
```
adamchainz, line 535:
```suggestion
Call this fixture to set ``settings.ROOT_URLCONF`` to a list of urlpatterns created within the test.
This keeps the URLs local unlike :ref:`pytest.mark.urls`.
```

(that reference probably needs adjusting to work)

adamchainz (changes requested):
I think this feature is pretty sensible to add.

Can you add a reference from the `pytest.mark.urls` docs to the new fixture?

orf:
Done, thank you @adamchainz 🚀

PR #1070, opened 2023-07-13
Fix issue with {% if model.related %} where related is None. Fixes #1059

### LINKED ISSUE #1059: FAIL_INVALID_TEMPLATE_VARS interferes with {% if model.one_to_one_field %}
Assuming I have the following model:

```python
class Person(models.Model):
  pass

class Purpose(models.Model):
  person = models.ForeignKey(Person, on_delete=models.CASCADE, related_name="purpose")
```

and the following template (assuming `person` is an instance of `Person` with no associated `Purpose`

```jinja2
        {% if person.purpose %}
            yes
        {% else %}
            no
        {% endif %}
```

Expected: "no"
Actual: "yes" if FAIL_INVALID_TEMPLATE_VARS is true, "no" otherwise

As far as I can see, person.purpose raises `RelatedObjectDoesNotExist`, which is then caught by django and converted into the `string_if_invalid`, which is set to an `InvalidVarException` instance.
The `IfNode` then uses `if match:` to determine whether to render the content, with match being our `InvalidVarException`.

Please let me know if I'm holding it wrong, but especially since this deviates from the regular behavior it looks like a bug to me

voidus:
I couldn't get the tests to run on my machine with either the makefile, pytest directly or tox, so I'm trusting the github actions.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -627,6 +627,34 @@ def _fail_for_invalid_template_variable():
             else:
                 return msg
 
+        def __bool__(self) -> bool:
+            # This is needed to avoid problems with conditionals on related objects.
+            #
+            # Given:
+            #   class Person(models.Model):
+            #     pass
+            #
+            #   class Purpose(models.Model):
+            #     person = models.ForeignKey(Person, on_delete=models.CASCADE, related_name="purpose")
+            #
+            # and the following template (assuming person is an instance of Person with no
+            # associated Purpose)
+            #
+            #   {% if person.purpose %}
+            #       yes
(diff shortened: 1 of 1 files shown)


voidus:
Ok got it this works `PYTHONPATH=$PWD DJANGO_SETTINGS_MODULE=pytest_django_test.settings_sqlite_file ./build/venv/bin/python -m pytest`

voidus:
Ok this is complete nonsense because `RelatedManagers` are truthy, and thus `bool(model.related)` is always `True`. I was looking for `model.related.exists()`.

Running the tests is still janky though. But sorry for the noise.

Issue #928, opened 2021-05-08
Expose TransactionTestCase.available_apps functionality


bluetech:
https://docs.djangoproject.com/en/3.2/topics/testing/advanced/#django.test.TransactionTestCase.available_apps

See also: PR #575

Note: Django marks this as private API so we would probably also mark it as such if we choose to add it.

bluetech:
Fixed by #1071

PR #1071, opened 2023-08-01
Expose available_apps in django_db marker

sullivan-sean:
Right now in a multi-schema environment, TransactionTestCases may fail to flush if there are foreign key dependencies between schemas because the test case will flush the DB in `_fixture_teardown` (see [source](https://github.com/django/django/blob/24068058a63c506c300629fcc491601abc968926/django/test/testcases.py#L1135-L1143)) but will not cascade truncation by default.

I've encountered this issue in a multi-tenant environment, and another user has posted about a similar issue [here](https://forum.djangoproject.com/t/transactiontestcase-available-apps-allow-cascade-true-db-cannot-be-flushed/18111).

Setting `available_apps` will enable `CASCADE` during truncation per the django docs ([link](https://docs.djangoproject.com/en/4.1/topics/testing/advanced/#django.test.TransactionTestCase.available_apps))

This PR exposes `available_apps` so that this experimental feature of `TransactionTestCase` can be used in `pytest-django` directly.

Addresses https://github.com/pytest-dev/pytest-django/issues/928

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -157,7 +157,7 @@ def _django_db_helper(
             reset_sequences,
             databases,
             serialized_rollback,
-            available_apps
+            available_apps,
         ) = validate_django_db(marker)
     else:
         (
@@ -165,7 +165,7 @@ def _django_db_helper(
             reset_sequences,
             databases,
             serialized_rollback,
-            available_apps
+            available_apps,
         ) = False, False, None, False, None
 
     transactional = transactional or reset_sequences or (
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -18,8 +18,9 @@ if TYPE_CHECKING:
     import django
 
     _DjangoDbDatabases = Optional[Union["Literal['__all__']", Iterable[str]]]
-    # transaction, reset_sequences, databases, serialized_rollback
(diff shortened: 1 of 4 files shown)


pytest_django/fixtures.py
```diff
@@ -156,14 +157,16 @@ def _django_db_helper(
             reset_sequences,
             databases,
             serialized_rollback,
+            available_apps
```
bluetech, line 160:
```suggestion
            available_apps,
```

pytest_django/fixtures.py
```diff
@@ -156,14 +157,16 @@ def _django_db_helper(
...
         ) = validate_django_db(marker)
     else:
         (
             transactional,
             reset_sequences,
             databases,
             serialized_rollback,
-        ) = False, False, None, False
+            available_apps
```
bluetech, line 168:
```suggestion
            available_apps,
```

pytest_django/plugin.py
```diff
@@ -389,6 +389,7 @@ def get_order_number(test: pytest.Item) -> int:
                     reset_sequences,
                     databases,
                     serialized_rollback,
+                    available_apps
```
bluetech, line 392:
```suggestion
                    available_apps,
```

bluetech (approved):
Thanks, this looks good!

PR #1074, opened 2023-09-29
Drop support for Python 3.7 (EOL since June 2023)

akx:
Someone needs to adjust this project's Required Checks to drop the 3.7 requirement :)


bluetech:
Thanks!

PR #1076, opened 2023-10-24
Make sure InvalidVarException.fail is reset

xavfernandez:
after a test using ignore_template_errors marker.

Otherwise, the first test set `fail` to `False`, never to be changed again.

### DIFF
--- a/tests/test_environment.py
+++ b/tests/test_environment.py
@@ -107,6 +107,47 @@ def test_invalid_template_variable(django_testdir) -> None:
         'django.template.loaders.filesystem.Loader',
         'django.template.loaders.app_directories.Loader',
     )
+    """
+)
+def test_invalid_template_variable_marker_cleanup(django_testdir) -> None:
+    django_testdir.create_app_file(
+        "<div>{{ invalid_var }}</div>", "templates/invalid_template_base.html"
+    )
+    django_testdir.create_app_file(
+        "{% include 'invalid_template_base.html' %}", "templates/invalid_template.html"
+    )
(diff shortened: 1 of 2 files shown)


tests/test_environment.py
```diff
@@ -101,6 +101,66 @@ def test_ignore(client):
...
+        'django.template.loaders.filesystem.Loader',
+        'django.template.loaders.app_directories.Loader',
+    )
+    ROOT_URLCONF = 'tpkg.app.urls'
+    """
+)
+def test_invalid_template_variable_marker_cleanup(django_testdir) -> None:
+    django_testdir.create_app_file(
+        """
+        from django.urls import path
```
bluetech, line 116: I think it will be easier to have the tests render the template directly using e.g. [`render_to_string`](https://docs.djangoproject.com/en/4.2/topics/templates/#django.template.loader.render_to_string) instead of setting up views and requests and everything.

bluetech (commented):
Good catch @xavfernandez, I'm surprised no one had noticed this bug before.

I left a suggestion for simplifying the test.

Also, scrolling up a bit I see that `_fail_for_invalid_template_variable` has a similar issue, although it's less consequential since it's a `session` fixture. But it would be good to fix it as well.

tests/test_environment.py
xavfernandez: Test updated

xavfernandez:
I also added cleanup to `_fail_for_invalid_template_variable` as you suggested

bluetech (approved):
Thanks!

PR #1077, opened 2023-10-26
Some assorted tweaks

bluetech:
Some minor maintenance changes (see the commits).


adamchainz:
Thank you for the burst of maintenance 👍 TIL Sphinx nitpicky !

PR #1079, opened 2023-10-27
Pyprojectification

bluetech:


adamchainz (commented):
LGTM, nice work getting to Mypy strict mode.

PR #1086, opened 2023-11-01
Add Python 3.12 support

foarsitter:
Without any changes this project supports Python 3.12. So version 4.6.0 is already compatible with Python 3.12

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,10 @@ jobs:
           python: '3.11'
           allow_failure: false
 
+        - name: py312-dj50-postgres-xdist-coverage
+          python: '3.12'
+          allow_failure: false
+
         - name: py311-dj50-postgres-xdist-coverage
           python: '3.11'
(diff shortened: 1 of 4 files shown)


adamchainz:
We need to add 3.12 to the test matrix. I think we should wait until we have Django 5.0 tested, because it is the only Django version that officially supports Python 3.2. (For now, that is. Django 4.2 will probably add support at some point.)

foarsitter:
Unofficially pytest-django already supports 3.12 so in my opinion there is no reason to wait for 5.0. I would point to https://forum.djangoproject.com/t/django-4-2-lts-compatibility-with-python-3-12/24508/3 to tell support for 3.12 is comming to 4.2 but you are the author of that posts.

adamchainz:
After merging #1087, I rebased this branch and added a commit with the changelog note + github actions run.

foarsitter:
@adamchainz LGTM, thanks!

PR #1044, opened 2023-01-24
WIP: Add Django 4.2 support

bluetech:
TODO:
- Currently blocked on https://github.com/jazzband/django-configurations/pull/349.
- I also want to add some testing for the new psycopg3 backend (with/without server-side bindings).

PR #1087, opened 2023-11-01
Add Django 5.0 support

adamchainz:
Copied the changes from #1044 for this new Django version.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,10 @@ jobs:
           python: '3.11'
           allow_failure: false
 
+        - name: py311-dj50-postgres-xdist-coverage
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -69,6 +69,10 @@ jobs:
           python: '3.11'
           allow_failure: false
 
+        - name: py311-dj50-postgres-xdist-coverage
+          python: '3.11'
+          allow_failure: false
+
```
adamchainz, line 75: I couldn’t figure out the logic for what is on the test matrix here, so I only added one entry.

This can be expanded to Python 3.12 with #1086.

bluetech (approved):

PR #1088, opened 2023-11-06
Run non-test workflows with Python 3.12

adamchainz:
Use the latest, fastest Python for linting, docs, and publishing.

PR #1089, opened 2023-11-06
Move docs to use Sphinx's minimal Makefile

adamchainz:
This has been the default [since Sphinx 1.5](https://www.sphinx-doc.org/en/master/changes.html#release-1-5-released-dec-5-2016): “Use make-mode of sphinx-quickstart by default. To disable this, use -M option”


adamchainz:
As noted on #1088, Codecov seem to be having problems.

bluetech (approved):

PR #1097, opened 2023-11-10
Avoid mutable global state in `SettingsWrapper`

bluetech:


pytest_django/fixtures.py
```diff
@@ -498,7 +497,9 @@ def async_rf() -> django.test.AsyncRequestFactory:
 
 
 class SettingsWrapper:
-    _to_restore: ClassVar[list[Any]] = []
+    def __init__(self) -> None:
+        self._to_restore: list[django.test.override_settings]
```
adamchainz, line 501: The type declaration can remain at the class level, like in a stub. IMO a little more readable.

bluetech: My idea was to make it look like `self._to_restore: list[django.test.override_settings] = []`

adamchainz: Yeah fair, I guess type checkers support it.

PR #1101, opened 2023-12-22
django_username_field example should use create_user not create

zsoldosp:
so that the password is hashed and `client.login` etc. works as expected


bluetech (approved):
Thanks!

PR #1109, opened 2024-01-29
Added assertMessages() from django.contrib.messages.

sdolemelipone:
Hi, let me know if you'd like this implemented differently and I can amend. This is my first time contributing to pytest-django.

Thanks!

### DIFF
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -230,15 +230,14 @@ if TYPE_CHECKING:
     ):
         ...
 
-    if USE_CONTRIB_MESSAGES:
-
-        def assertMessages(
-            response: HttpResponseBase,
-            expected_messages: Sequence[Message],
-            *args,
-            ordered: bool = ...,
-        ) -> None:
-            ...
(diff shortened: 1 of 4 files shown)


pytest_django/asserts.py
```diff
@@ -213,6 +230,16 @@ def assertNumQueries(
...
 
+    if USE_CONTRIB_MESSAGES:
+
+        def assertMessages(
+            response: HttpResponseBase,
+            expected_messages: Sequence[Message],
+            *args,
+            ordered: bool = ...,
+        ) -> None:
+            ...
```
bluetech, line 241: This condition doesn't have an affect inside of TYPE_CHECKING, so let's remove it.

```suggestion
    # Added in Django 5.0.
    def assertMessages(
        response: HttpResponseBase,
        expected_messages: Sequence[Message],
        *args,
        ordered: bool = ...,
    ) -> None:
        ...
```

bluetech (approved):
Thanks!

sdolemelipone:
Great, you're welcome :smiley:

PR #1112, opened 2024-02-09
Fix env-var name for parallel tox tests

meshy:
When running tests in parallel, we need to add a suffix to Django DB names to avoid clashes. This previously used `TOX_PARALLEL_ENV`, but Tox no longer sets this environment variable.

Tox does set an environment variable called `TOX_ENV_NAME`, which works for this purpose, so use that instead.


meshy:
Oh dear, this broke more tests that I accounted for.
I've moved this into draft for now.

PR #1113, opened 2024-02-13
A fixture which gives access to all databases in tests

dferens:
Hi,

At my company we have a problem that we have to use mark on every test:
```python
@pytest.mark.django_db(databases="__all__")
def test_something(db):
   ...

@pytest.mark.django_db(databases="__all__")
def test_other(db):
    ...

# + a bunch of tests which needs to be marked as well
```

I'd like to propose a solution:
```python
def setup_db(db, django_db_all_databases):
    pass

def test_something(setup_db):
    ...

def test_other(setup_db):
    ...
```

Let me know what you think. If you approve this, I will improve this PR with tests, docs, docstrings and so on. I wanted to know your opinion first.

Thanks!

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -44,6 +44,7 @@ __all__ = [
     "db",
     "transactional_db",
     "django_db_reset_sequences",
+    "django_db_all_databases",
     "django_db_serialized_rollback",
     "admin_user",
     "django_user_model",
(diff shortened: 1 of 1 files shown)


dferens:
Hey, can anyone look into this?

bluetech:
I would not want to merge something like this since it encourages a bad practice. So my suggestion, if you want to do this, is to put the following in an initial `conftest.py` file:

```py
import pytest

@pytest.hookimpl(tryfirst=True)
def pytest_collection_modifyitems(items):
    marker = pytest.mark.django_db(databases="__all__")
    for item in items:
        item.add_marker(marker)

PR #1103, opened 2024-01-06
Upgrade GitHub Action setup-python

cclauss:
https://github.com/actions/setup-python/releases

PR #1119, opened 2024-04-12
Configure dependabot to check for updates to GitHub Actions every weekday

flaeppe:
Ref: https://github.com/pytest-dev/pytest-django/pull/1103#issuecomment-1880164519


adamchainz (approved):

bluetech:
Thanks! I think daily is not really necessary, so I'll change to monthly and merge.

PR #1120, opened 2024-04-12
Avoid running database migrations for `SimpleTestCase`

### LINKED ISSUE #1049: `SimpleTestCase` triggers database migrations
The default implementation of `SimpleTestCase` disables database queries per default, which is why I'm thinking that migrating the database could conditionally happen depending on the value of `SimpleTestCase.databases`?

> [SimpleTestCase](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.SimpleTestCase) disallows database queries by default. This helps to avoid executing write queries which will affect other tests since each SimpleTestCase test isn’t run in a transaction. If you aren’t concerned about this problem, you can disable this behavior by setting the databases class attribute to '__all__' on your test class.

Ref: https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.SimpleTestCase.databases

flaeppe:
Closes #1049

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -559,6 +559,8 @@ def _django_setup_unittest(
     try:
         TestCaseFunction.runtest = non_debugging_runtest  # type: ignore[method-assign]
 
+        # Don't set up the DB if the unittest does not require DB.
+        # The `databases` propery seems like the best indicator for that.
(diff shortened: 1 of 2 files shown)


bluetech (approved):
Makes sense, thanks!

PR #1121, opened 2024-04-25
fail-on-template-vars: be more discreet with ignore_template_errors

xavfernandez:
When using the `@pytest.mark.ignore_template_errors` on a test, we could expect the test to return to its original behavior (just like running them without `--fail-on-template-vars`).

This would ease the use of the option since ignored tests are less likely to crash.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -644,8 +644,9 @@ def _fail_for_invalid_template_variable():
     class InvalidVarException:
         """Custom handler for invalid strings in templates."""
 
-        def __init__(self) -> None:
+        def __init__(self, *, origin_value: str = "") -> None:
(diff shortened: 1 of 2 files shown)


pytest_django/plugin.py
```diff
@@ -644,8 +644,9 @@ def _fail_for_invalid_template_variable():
     class InvalidVarException:
         """Custom handler for invalid strings in templates."""
 
-        def __init__(self) -> None:
+        def __init__(self, *, origin_value: str = "") -> None:
```
bluetech, line 647: I don't think the default is needed

```suggestion
        def __init__(self, *, origin_value: str) -> None:
```

bluetech (approved):
Thanks, makes sense!

PR #1128, opened 2024-07-09
Added import to database documentation sample code

pzarfos:
This PR updates a code sample in the documentation when using an existing database for the tests.

I ran across this omission when setting up my own project... I copied the original sample code into my project, and then Python complained that the `settings.DATABASES` was not defined.
So I added a one-liner `from django.conf import settings` to the code sample, and now it runs as is.


bluetech (approved):
Thanks!

PR #1129, opened 2024-07-10
fail-on-template-vars: modernize stack inspection code

xavfernandez:
inspect.stack() returns a list of namedtuple (or retrocompatible objects) since Python 3.5+: let's use the named attribute.

cf https://docs.python.org/3/library/inspect.html#inspect.stack

And once we have access to a FrameInfo object/namedtuple, access to its frame object and its f_locals member should not need to iterate on all its members:
https://docs.python.org/3/reference/datamodel.html#frame-objects

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -686,7 +686,6 @@ def _fail_for_invalid_template_variable():
             frameinfo = reduce(
                 lambda x, y: y if y.function == "render" and "base.py" in y.filename else x, stack
             )
-            # finding only the frame locals in all frame members
             # ``django.template.base.Template``
             template = frameinfo.frame.f_locals["self"]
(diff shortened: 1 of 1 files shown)


pytest_django/plugin.py
```diff
@@ -673,16 +671,11 @@ def _get_origin() -> str | None:
 
             # finding the ``render`` needle in the stack
             frameinfo = reduce(
-                lambda x, y: y[3] == "render" and "base.py" in y[1] and y or x, stack
+                lambda x, y: y if y.function == "render" and "base.py" in y.filename else x, stack
             )
-            # assert 0, stack
-            frame = frameinfo[0]
             # finding only the frame locals in all frame members
```
bluetech, line 676: Outdated comment

```suggestion
```

PR #1142, opened 2024-09-02
ci: use trusted publishing for release

bluetech:
See https://docs.pypi.org/trusted-publishers/. Loosely based on pytest CI.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,43 @@
+name: deploy
+
+on:
+  push:
+    tags:
+      - "*"
+
+# Set permissions at the job level.
+permissions: {}
+
+jobs:
+  package:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 10
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Build and Check Package
+        uses: hynek/build-and-inspect-python-package@2dbbf2b252d3a3c7cec7a810e3ed5983bd17b13a  # v2.8.0
+
+  deploy:
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && github.repository == 'pytest-dev/pytest-django'
+    needs: [package]
+    runs-on: ubuntu-22.04
+    environment: deploy
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      # For trusted publishing.
+      id-token: write
+
+    steps:
+      - name: Download Package
+        uses: actions/download-artifact@v4
+        with:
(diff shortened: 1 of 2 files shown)


bluetech:
We'll see if it works on release, might need fixups...

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+
+      - name: Build and Check Package
+        uses: hynek/build-and-inspect-python-package@2dbbf2b252d3a3c7cec7a810e3ed5983bd17b13a  # v2.8.0
+
+  deploy:
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags') && github.repository == 'pytest-dev/pytest-django'
```
webknjaz, line 25: FYI, there's a `github.ref_name` available now.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,43 @@
...
+
+    steps:
+      - name: Download Package
+        uses: actions/download-artifact@v4
+        with:
+          name: Packages
+          path: dist
+
+      - name: Publish package
+        uses: pypa/gh-action-pypi-publish@8a08d616893759ef8e1aa1f2785787c0b97e20d6 # v1.10.0
```
webknjaz, line 43: With this version, it's possible to opt in to uploading digital attestations, FYI.

.github/workflows/main.yml
```diff
@@ -126,34 +124,3 @@ jobs:
...
-        with:
-          python-version: '3.12'
-
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install --upgrade build
-
-      - name: Build package
-        run: python -m build
```
webknjaz, line 153: It's usually nice to keep it in the main workflow and put the job building the artifacts in front of the tests. Then, it'd be possible to test whatever's going to be uploaded, as opposed to building something new and potentially differing.

.github/workflows/deploy.yml
bluetech: Not sure about `github.ref_name` but maybe `github.ref_type` is better for this?

.github/workflows/deploy.yml
bluetech: Will look into it for the next release

.github/workflows/main.yml
bluetech: Agree, but I didn't want to complicate/slow down the test CI for now, but wanted to at least do it right for the deploy workflow.

PR #1143, opened 2024-09-02
Drop support for EOL Django 3.2, 4.1, add support for 5.1

bluetech:


bluetech:
Newer Djangos seem to have an issue with psycopg2cffi, so using the opportunity to switch to psycopg3.

PR #1148, opened 2024-09-26
Lock/unlock of db breaks if pytest is executed twice in the same process (fixes #1147)

### LINKED ISSUE #1147: Unblock not cleaned up at the end of pytest execution?
I realize this is a bit of an edge case, but there seems to be something off about the cleanup for `pytest.mark.django_db`. If you run pytest twice in the same process all tests that are marked `django_db` fail with the error reproduced below.

A reproduction repo is at https://github.com/boxed/pytest_django_block_db_issue. Run [run_pytest_twice.py](https://github.com/boxed/pytest_django_block_db_issue/blob/main/run_pytest_twice.py) after initializing a venv with pytest-django, and django.

(The reason I'm doing this is because I'm working on the next version of mutmut, which is looking like it could go ~100 times faster than the current version, but this issue stops me from trying out out iommi)

I have tracked down the exception to `_blocking_wrapper` (which was annoyingly hard because of `__tracebackhide__`), but I'm at a bit of a loss as to what my next step is.

```
_______________________________________________________ ERROR at setup of test_issue ________________________________________________________

request = <SubRequest '_django_db_marker' for <Function test_issue>>

    @pytest.fixture(autouse=True)
    def _django_db_marker(request: pytest.FixtureRequest) -> None:
        """Implement the django_db marker, internal to pytest-django."""
        marker = request.node.get_closest_marker("django_db")
        if marker:
>           request.getfixturevalue("_django_db_helper")

venv/lib/python3.12/site-packages/pytest_django/plugin.py:533: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv/lib/python3.12/site-packages/pytest_django/fixtures.py:144: in django_db_setup
    db_cfg = setup_databases(
venv/lib/python3.12/site-packages/django/test/utils.py:206: in setup_databases
    connection.creation.create_test_db(
venv/lib/python3.12/site-packages/django/db/backends/base/creation.py:78: in create_test_db
    call_command(
venv/lib/python3.12/site-packages/django/core/management/__init__.py:194: in call_command
    return command.execute(*args, **defaults)
venv/lib/python3.12/site-packages/django/core/management/base.py:459: in execute
    output = self.handle(*args, **options)
venv/lib/python3.12/site-packages/django/core/management/base.py:107: in wrapper
    res = handle_func(*args, **kwargs)
venv/lib/python3.12/site-packages/django/core/management/commands/migrate.py:118: in handle
    executor = MigrationExecutor(connection, self.migration_progress_callback)
venv/lib/python3.12/site-packages/django/db/migrations/executor.py:18: in __init__
    self.loader = MigrationLoader(self.connection)
venv/lib/python3.12/site-packages/django/db/migrations/loader.py:58: in __init__
    self.build_graph()
venv/lib/python3.12/site-packages/django/db/migrations/loader.py:235: in build_graph
    self.applied_migrations = recorder.applied_migrations()
venv/lib/python3.12/site-packages/django/db/migrations/recorder.py:89: in applied_migrations
    if self.has_table():
venv/lib/python3.12/site-packages/django/db/migrations/recorder.py:63: in has_table
    with self.connection.cursor() as cursor:
venv/lib/python3.12/site-packages/django/utils/asyncio.py:26: in inner
    return func(*args, **kwargs)
venv/lib/python3.12/site-packages/django/db/backends/base/base.py:320: in cursor
    return self._cursor()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <DatabaseWrapper vendor='sqlite' alias='default'>, name = None

    def _cursor(self, name=None):
        self.close_if_health_check_failed()
>       self.ensure_connection()
E       RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.

venv/lib/python3.12/site-packages/django/db/backends/base/base.py:296: RuntimeError
```

boxed:


bluetech (changes requested):
Thanks for the patch, I can see the problem this is fixing.

I would prefer another fix. The problem occurs because, in order to block DB access by default, pytest-django runs an initial `block()` in `_setup_django()`, but doesn't restore it at the end, so it sticks around. The correct fix IMO is to add a `restore` to balance the `_setup_django` one, probably in a `pytest_unconfigure` hook. `DjangoDbBlocker` by itself is good.

Would you mind trying it?

boxed:
Hmm. Yea that seems better. Although there is something iffy with the code as it is now, because the setting of the original ensure connection function isn't guaranteed to set the original function every time. Maybe an assert that the `__module__` of the function is what we expect might be nice too?

boxed:
This change works on my test project and I think this is what you meant in your comment above.

bluetech (approved):
Thanks!

PR #1151, opened 2024-09-28
plugin: unconfigure should `restore` not `unblock`

bluetech:
`restore` pops the `block` in `_setup_django`, while `unblock` pushes an unblock. We want to leave things clean, so should `restore`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: docs test clean fix
 
 test:
-	tox -e py311-dj42-sqlite_file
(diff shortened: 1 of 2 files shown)


bluetech:
@boxed I didn't initially notice the `unblock` instead of `restore` in #1148.

boxed:
I tried `restore`. That doesn't work.

bluetech:
Hmm strange, I don't see how that can be. I made it a bit more aggressive now. Can you try it?

boxed:
Yea, that works.

PR #1162, opened 2024-11-08
fixtures: provide per-scope fixtures for unblocking database access

m000:
Allow the convenience of `db` fixture for all fixture scopes.

The `db` fixture is very convenient when writing fixtures that create objects. However, for non function-scoped fixtures using `db` is not possible. This PR adds `db_<scope>` fixtures that can be used to the same effect as `db` when creating fixtures that use other scopes.


m000:
The use of the intermediate `_django_db_helper*` fixtures could also be avoided, which would simplify the fixtures code a bit.

m000:
NVM. For anyone attempting the same thing, this has side-effects: by enabling database access from a wider scope, (of course) ALL tests in that scope will be allowed database access.

PR #1167, opened 2025-01-10
Set the `serialized_aliases` Argument of `setup_databases`.

xmatos-so:
The `serialized_aliases` argument of the `setup_databases` function determines what subset of aliases test databases should have their state serialized to allow usage of the `serialized_rollback` feature. If it's not provided, it defaults to aliases.

Django 5 removed the `SERIALIZE` test setting, as it can be inferred from the databases with the `serialized_rollback` option enabled. Django's default test runner, `DiscoverRunner`, uses logic similar to this to build the `serialized_aliases` set and pass it to `setup_databases`, as implemented in `django.test.runner.DiscoverRunner.run_tests`.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -302,18 +302,20 @@ def _disable_migrations() -> None:
     migrate.Command = MigrateSilentCommand
 
 
-def _get_serialized_aliases(tests: list) -> set:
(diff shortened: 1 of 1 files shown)


bluetech:
Thanks for the PR @xmatos-so. Something like this is now implemented in #1186 and will be included in the next release.

PR #1170, opened 2025-02-07
Adds using to assert_num_queries

kingbuzzman:
* Depend on https://github.com/pytest-dev/pytest-django/pull/1171

Adds quality of life to replace

```python
from django.db import connections

def test_test():
    with django_assert_num_queries(1, connection=connections["log"]):
        Model.objects.get()
```

with

```python
def test_test():
    with django_assert_num_queries(1, using="log"):
        Model.objects.get()
```

_I have multiple DBs and this is driving me nuts..._

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Compatibility
 ^^^^^^^^^^^^^
 
 * Added official support for Python 3.13.
+* Adds ``using`` argument to assert_num_query
 
 Bugfixes
 ^^^^^^^^
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -9,8 +9,9 @@ Compatibility
 
 * Added official support for Python 3.13.
 
-* Adds ``using`` argument to ``pytest_django.fixtures.django_assert_num_queries`` 
-  and `pytest_django.fixtures.django_assert_max_num_queries``
+* Added ``using`` argument to :fixture:`django_assert_num_queries` and
+  :fixture:`django_assert_max_num_queries` to easily specify the database
+  alias to use.
 
 Bugfixes
 ^^^^^^^^
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,7 +8,9 @@ Compatibility
 ^^^^^^^^^^^^^
 
 * Added official support for Python 3.13.
-* Adds ``using`` argument to assert_num_query
+
+* Adds ``using`` argument to ``pytest_django.fixtures.django_assert_num_queries`` 
+  and `pytest_django.fixtures.django_assert_max_num_queries``
 
 Bugfixes
 ^^^^^^^^
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -466,9 +466,9 @@ If you use type annotations, you can annotate the fixture like this::
 .. py:function:: django_assert_max_num_queries(num, connection=None, info=None, using=None)
 
   :param num: expected maximum number of queries
-  :param connection: optional non-default DB connection
+  :param connection: optional database connection
   :param str info: optional info message to display on failure
-  :param str using: optional the non-default DB connection name
+  :param str using: optional database connection name
 
 This fixture allows to check for an expected maximum number of DB queries.
 
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -428,7 +428,7 @@ Example
   :param num: expected number of queries
   :param connection: optional database connection
   :param str info: optional info message to display on failure
-  :param str using: optional database connection name
+  :param str using: optional database alias
 
 This fixture allows to check for an expected number of DB queries.
 
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -423,7 +423,7 @@ Example
 ``django_assert_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. py:function:: django_assert_num_queries(num, connection=None, info=None, using=None)
+.. py:function:: django_assert_num_queries(num, connection=None, info=None, *, using=None)
 
   :param num: expected number of queries
   :param connection: optional database connection
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -426,9 +426,9 @@ Example
 .. py:function:: django_assert_num_queries(num, connection=None, info=None, using=None)
 
   :param num: expected number of queries
-  :param connection: optional non-default DB connection
+  :param connection: optional database connection
   :param str info: optional info message to display on failure
-  :param str using: optional the non-default DB connection name
+  :param str using: optional database connection name
 
 This fixture allows to check for an expected number of DB queries.
 
--- a/docs/helpers.rst
+++ b/docs/helpers.rst
@@ -423,11 +423,12 @@ Example
 ``django_assert_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. py:function:: django_assert_num_queries(num, connection=None, info=None)
+.. py:function:: django_assert_num_queries(num, connection=None, info=None, using=None)
 
   :param num: expected number of queries
   :param connection: optional non-default DB connection
   :param str info: optional info message to display on failure
+  :param str using: optional the non-default DB connection name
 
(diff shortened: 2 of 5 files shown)


docs/helpers.rst
```diff
@@ -423,11 +423,12 @@ Example
 ``django_assert_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. py:function:: django_assert_num_queries(num, connection=None, info=None)
+.. py:function:: django_assert_num_queries(num, connection=None, info=None, using=None)
 
   :param num: expected number of queries
   :param connection: optional non-default DB connection
   :param str info: optional info message to display on failure
+  :param str using: optional the non-default DB connection name
```
adamchainz, line 431: "default" would work, right? No need to say "non-default".

```suggestion
  :param connection: optional database connection
  :param str info: optional info message to display on failure
  :param str using: optional database connection name
```

docs/helpers.rst
```diff
@@ -462,11 +463,12 @@ If you use type annotations, you can annotate the fixture like this::
 ``django_assert_max_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. py:function:: django_assert_max_num_queries(num, connection=None, info=None)
+.. py:function:: django_assert_max_num_queries(num, connection=None, info=None, using=None)
 
   :param num: expected maximum number of queries
   :param connection: optional non-default DB connection
   :param str info: optional info message to display on failure
+  :param str using: optional the non-default DB connection name
```
adamchainz, line 471:
```suggestion
  :param connection: optional database connection
  :param str info: optional info message to display on failure
  :param str using: optional database connection name
```

pytest_django/fixtures.py
```diff
@@ -625,6 +627,11 @@ def _assert_num_queries(
     else:
         conn = connection
 
+    if using:
```
adamchainz, line 630: please add a check before line 625 that both `connection` and `using` are not provided

adamchainz (changes requested):
Seems like a valuable QOL and would actually align with [Django’s API](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries)

docs/helpers.rst
kingbuzzman: I wanted to keep the changes to a minimal and avoid changing things i also disagreed with, but it’s been like this for a long time; going against the grain is always rougher!

docs/helpers.rst
kingbuzzman: Ditto.

I’d change this if the maintainer wants it.

pytest_django/fixtures.py
kingbuzzman: I thought about this originally, and decided against it at this junction, it’s a new argument that is the deciding factor. A case where both would be introduced at the same time would not be an issue for a long time.. and again, wanted to keep the changes to a minimal.

I suppose i can add a warning, which is what i intended to do in the first place… i don’t know i’m in two minds. 😇

docs/helpers.rst
kingbuzzman: LOL You’re a member. lol my bad!

pytest_django/fixtures.py
kingbuzzman: Ok i’ll do this! I thought you were some rando’, 😇

pytest_django/fixtures.py
```diff
@@ -617,6 +619,7 @@ def _assert_num_queries(
     exact: bool = True,
     connection: Any | None = None,
     info: str | None = None,
+    using: str | None = None,
```
bluetech, line 622:
```suggestion
    *,
    using: str | None = None,
```

pytest_django/fixtures.py
```diff
@@ -606,6 +607,7 @@ def __call__(
         num: int,
         connection: Any | None = ...,
         info: str | None = ...,
+        using: str | None = ...,
```
bluetech, line 610:
```suggestion
        *,
        using: str | None = ...,
```

docs/helpers.rst
```diff
@@ -423,11 +423,12 @@ Example
 ``django_assert_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. py:function:: django_assert_num_queries(num, connection=None, info=None)
+.. py:function:: django_assert_num_queries(num, connection=None, info=None, using=None)
```
bluetech, line 426:
```suggestion
.. py:function:: django_assert_num_queries(num, connection=None, info=None, *, using=None)
```

docs/helpers.rst
```diff
@@ -423,11 +423,12 @@ Example
...
 
   :param num: expected number of queries
-  :param connection: optional non-default DB connection
+  :param connection: optional database connection
   :param str info: optional info message to display on failure
+  :param str using: optional database connection name
```
bluetech, line 431:
```suggestion
  :param str using: optional database alias
```

docs/helpers.rst
```diff
@@ -462,11 +463,12 @@ If you use type annotations, you can annotate the fixture like this::
 ``django_assert_max_num_queries``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. py:function:: django_assert_max_num_queries(num, connection=None, info=None)
+.. py:function:: django_assert_max_num_queries(num, connection=None, info=None, using=None)
```
bluetech, line 466:
```suggestion
.. py:function:: django_assert_max_num_queries(num, connection=None, info=None, *, using=None)
```

docs/helpers.rst
```diff
@@ -462,11 +463,12 @@ If you use type annotations, you can annotate the fixture like this::
...
 
   :param num: expected maximum number of queries
-  :param connection: optional non-default DB connection
+  :param connection: optional database connection
   :param str info: optional info message to display on failure
+  :param str using: optional database connection name
```
bluetech, line 471:
```suggestion
  :param str using: optional database alias
```

pytest_django/fixtures.py
bluetech: Not sure if this before or after the update, but in any case, what @adamchainz meant is to add a "mutually exclusive" check before the `if connection is None: ...` part. Also we shouldn't warn, just raise an error, it's better to be strict than to silently ignore a parameter.

So something like this:

```py
from django.db import connections
from django.db import connection as default_conn

if connection is not None and using is not None:
    raise ValueError('The "connection" and "using" parameter cannot be used together')

if connection is not None:
    conn = connection
else if using is not None:
    conn = connections[using]
else:
   conn = default_conn
```

pytest_django/fixtures.py
```diff
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import os
+import warnings
```
kingbuzzman, line 6:
```suggestion
```

pytest_django/fixtures.py
```diff
@@ -617,14 +620,22 @@ def _assert_num_queries(
     exact: bool = True,
     connection: Any | None = None,
     info: str | None = None,
+    *,
+    using: str | None = None,
 ) -> Generator[django.test.utils.CaptureQueriesContext, None, None]:
-    from django.test.utils import CaptureQueriesContext
+    from django.db import connections
+    from django.db import connection as default_conn
```
kingbuzzman, line 627:
```suggestion
    from django.db import connection as default_conn
    from django.test.utils import CaptureQueriesContext
```

docs/changelog.rst
```diff
@@ -9,6 +9,9 @@ Compatibility
 
 * Added official support for Python 3.13.
 
+* Adds ``using`` argument to ``pytest_django.fixtures.django_assert_num_queries`` 
+  and `pytest_django.fixtures.django_assert_max_num_queries``
```
bluetech, line 13:
```suggestion
* Added ``using`` argument to :fixture:`django_assert_num_queries`  and
  fixture:`django_assert_max_num_queries` to easily specify the database
  alias to use.
```

bluetech (approved):
Looks good, thanks!

bluetech:
Soon!

PR #1171, opened 2025-02-07
Fixes CI issue with main + py11

kingbuzzman:
As mentioned in the title, [django does not support py11 in the latest (main) branch](https://github.com/django/django/blob/main/pyproject.toml#L8) and its causing that particular test to fail.

Let me know if you want me to move things around.

_This came about because i wanted to do https://github.com/pytest-dev/pytest-django/pull/1170 and ran into this issue there._

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -103,17 +103,21 @@ jobs:
           python: '3.9'
           allow_failure: false
 
+        - name: py312-djmain-sqlite-coverage
+          python: '3.12'
+          allow_failure: true
+
         - name: py312-dj51-sqlite-xdist-coverage
           python: '3.12'
+          allow_failure: false
+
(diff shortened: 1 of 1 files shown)


adamchainz:
The expected checks need changing before this can be merged - @bluetech would you mind doing this?

webknjaz:
Perhaps, integrate alls-green to avoid this in the future?

bluetech:
Yes, the the required checks thing is a periodic bother. I'll try alls-green, thanks for the suggestion.

> I haven't merged anything here so I want to be sure you're okay with the change.

For the record, I'm okay with it :)

webknjaz:
I think it's more annoying when the matrix factors change, though.

kingbuzzman:
@bluetech merge me up scotty!!

kingbuzzman:
Thank you!

PR #1172, opened 2025-02-08
ci: use alls-green so we can have a single required check

bluetech:


webknjaz (approved):
Looks right, once you drop that intentional failure.

For pytest, I recommended waiting for a while so that existing PRs would start reporting the new check before it's required. But this seems to have just a few active ones that would be easy to rebase. So it's fine to add it to the branch protection right away...

tests/test_asserts.py
```diff
@@ -71,3 +71,4 @@ def test_sanity() -> None:
         pass
 
     assert assertContains.__doc__
+"LINT FAILURE"
```
kingbuzzman, line 74: Fixes linter issue

```suggestion


"LINT FAILURE"
```

PR #1177, opened 2025-02-10
Adds ability to capture all the db queries at once

kingbuzzman:
When dealing with multiple databases, in my case we have over 6. We don't much care to capture db1's queries vs db2 vs db3... we, holistically, care more about "how many queries where made in total", and then if the number is off go an explore the exact case by the query count is off.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -597,6 +597,41 @@ def _live_server_helper(request: pytest.FixtureRequest) -> Generator[None, None,
     live_server._live_server_modified_settings.disable()
 
 
+class CaptureAllConnectionsQueriesContext:
+    """
+    Context manager that captures all queries executed by Django ORM across all Databases in settings.DATABASES.
+    """
+
+    def __init__(self):
+        from django.db import connections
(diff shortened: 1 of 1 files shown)


kingbuzzman:
@bluetech let me know if there is something you want me to change

bluetech:
This seems sensible to me.

One thing, pytest-django tries to mostly follow Django upstream, and just uses `django.test` underneath as you know. So ideally, this feature suggestion would go through Django upstream.

Would you consider sending your suggestion to the Django developer list (https://forum.djangoproject.com/c/internals/5), or open a [Django ticket](https://code.djangoproject.com/) to discuss it?

adamchainz:
Yes, this seems like a good feature to add upstream first. I would propose it extending `assertNumQueries` to allow `using` to accept both an iterable of database names, like `using={"default", "other"}`, and some special value for all databases, probably `using="__all__"`. There is precedent for `__all__` in [`TestCase.databases`](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.TestCase.databases).

kingbuzzman:
Reached out to [Django community](https://forum.djangoproject.com/t/add-ability-to-capture-all-the-db-queries-at-once-during-tests/39650) for feedback

PR #1179, opened 2025-03-06
Add support for Django 5.2

adamchainz:
It’s in beta since Feb 19: https://www.djangoproject.com/weblog/2025/feb/19/django-52-beta-1-released/ . Final release expected April 2nd.

Since we're already testing the main branch, few changes are needed.

I copied what to change from #1143.

I’m a bit confused about what entries we include on the testing matrix, so I just added a few that I thought would make sense.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,10 @@ jobs:
           python: '3.13'
           allow_failure: false
 
+        - name: py313-dj52-postgres-xdist-coverage
+          python: '3.13'
+          allow_failure: false
+
         - name: py313-dj51-postgres-xdist-coverage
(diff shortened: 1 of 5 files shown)


bluetech (approved):
LGTM, thanks.

> I’m a bit confused about what entries we include on the testing matrix, so I just added a few that I thought would make sense.

That's what I do :)

adamchainz:
> That's what I do :)

Okay, fair. Maybe we can find some way to make it smarter in the future.

One possible optimization, which I use in my projects, is to put all Python and DB versions in a matrix, and run tests for all matching Django versions. It’s much faster to rerun a bunch of tests than spin up a new GitHub Actions worker.

PR #1180, opened 2025-03-06
Remove testing on MyISAM

adamchainz:
MyISAM testing was added back in 2012 (#78) because it would have detected one bug in relation to transactional ordering.

MyISAM was not recommended back then and even less so now. It’s non-transactional nature has many implications and [Django’s database documentation](https://docs.djangoproject.com/en/stable/ref/databases/) caveats against using it.

I think we can remove the testing now and reduce the maintenance burden. As far as I can tell, it was only added retroactively to guard against that one bug, 13 years ago.


pytest_django_test/settings_mysql.py
```diff
@@ -11,7 +11,6 @@
         "PASSWORD": environ.get("TEST_DB_PASSWORD", ""),
         "HOST": environ.get("TEST_DB_HOST", "localhost"),
         "OPTIONS": {
-            "init_command": "SET default_storage_engine=InnoDB",
```
adamchainz, line 14: Was always unnecessary because it's been the default in MySQL for a long time.

pelme:
Looks good to me!

pelme (approved):

pelme:
I can't imagine many people being affected by dropping explicit myisam support but maybe it is worth mentioning in the release notes?

bluetech (approved):

adamchainz:
> I can't imagine many people being affected by dropping explicit myisam support but maybe it is worth mentioning in the release notes?

Sure, can’t hurt, done. I also added PR links to my changelog notes from today, could be useful going forwards.

Issue #1169, opened 2025-01-28
Simpler way to run `pytest` from `manage.py` script?


paduszyk:
[Docs](https://pytest-django.readthedocs.io/en/latest/faq.html#how-can-i-use-manage-py-test-with-pytest-django) suggest to define custom test runner to invoke `pytest` (thus, related `pytest-django` functionality) via the `manage.py` script.

I used the following snippet and everything works as expected:

```python
# manage.py
import sys
from pathlib import Path

import environ

from django.core.management import execute_from_command_line


def main():
    if (dotenv_path := Path(__file__).parent / ".env").exists():
        environ.Env.read_env(dotenv_path)

    if sys.argv[1:2] == ["test"]:
        import pytest

        sys.exit(pytest.main(sys.argv[2:]))

    execute_from_command_line(sys.argv)


if __name__ == "__main__":
    main()
```

What are the advantages of the solution presented in the docs compared to the one above? Thanks in advance for any feedback.

bluetech:
The only advantage is that it's a bit more integrated with Django, e.g. the command line parsing. But if what you did works for you, I don't see much problem with it!

kingbuzzman:
if anyone ever tried to debug this / add any arguments, this will get chaotic quick.

Personally, I use the `pytest` args `--last-failed` / `-x` / `-s` /  `-vv` / `--pdb` daily.  Not being able to specify them at will, because you dont want them on all the time, makes this solution very poor.
And `--pdbcls=IPython.terminal.debugger:TerminalPdb` is on by default inside `pyproject.toml`. [Example](https://github.com/kingbuzzman/django-squash/blob/036466f/pyproject.toml#L192)

But like @bluetech, said, if this solution works for you, 👍 

---

@bluetech Why isn't `PytestTestRunner` included in the library to lower the entry level? It would be much faster to add `TEST_RUNNER = 'pytest_django.runner.PytestTestRunner'` than it is to copy it somewhere into your own repo and maintain it -- if you need more customizing just override it with your own while inheriting from one provided by `pytest_django` -- no?

bluetech:
Thanks to #1182 you can now use `TEST_RUNNER = 'pytest_django.runner.TestRunner'`!

paduszyk:
@bluetech Thanks 👍🏻

PR #1182, opened 2025-03-14
Integrate pytest into django's manage.py test

kingbuzzman:
To lower the barier of entry for questions like https://github.com/pytest-dev/pytest-django/issues/1169

### DIFF
--- a/pytest_django/runner.py
+++ b/pytest_django/runner.py
@@ -1,0 +1,36 @@
+class PytestTestRunner:
+    """Runs pytest to discover and run tests."""
+
+    def __init__(self, verbosity=1, failfast=False, keepdb=False, **kwargs):
+        self.verbosity = verbosity
+        self.failfast = failfast
+        self.keepdb = keepdb
+
+    @classmethod
+    def add_arguments(cls, parser):
+        parser.add_argument(
+            '--keepdb', action='store_true',
+            help='Preserves the test DB between runs.'
+        )
+
+    def run_tests(self, test_labels, **kwargs):
+        """Run pytest and return the exitcode.
+
+        It translates some of Django's test command option to pytest's.
+        """
+        import pytest
+
+        argv = []
+        if self.verbosity == 0:
+            argv.append('--quiet')
+        if self.verbosity == 2:
+            argv.append('--verbose')
+        if self.verbosity == 3:
+            argv.append('-vv')
+        if self.failfast:
+            argv.append('--exitfirst')
+        if self.keepdb:
+            argv.append('--reuse-db')
+
+        argv.extend(test_labels)
+        return pytest.main(argv)
--- a/pytest_django/runner.py
+++ b/pytest_django/runner.py
@@ -1,9 +1,8 @@
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 
 if TYPE_CHECKING:
     from argparse import ArgumentParser
-    from typing import Any
 
 
 class PytestTestRunner:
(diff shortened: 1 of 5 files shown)


pytest_django/runner.py
```diff
@@ -0,0 +1,35 @@
...
+        """
+        import pytest
+
+        argv = []
+        if self.verbosity == 0:
+            argv.append("--quiet")
+        if self.verbosity == 2:
+            argv.append("--verbose")
+        if self.verbosity == 3:
+            argv.append("-vv")
```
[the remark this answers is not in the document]
meschac38700: @kingbuzzman Yes, I've seen that.

You have a failing test. since beyond verbosity greater than or equal to 4, according to the logic of the tests, the verbosity option is ignored

This approach might be more efficient. It also fixes the test.

```Python
MAX_VERBOSITY = 3

if self.verbosity == 0:
    argv.append("--quiet")
elif self.verbosity > 0 and self.verbosity <= MAX_VERBOSITY:
    _verbosity = "v" * (self.verbosity -1)
    argv.append(f"-{_verbosity}"
```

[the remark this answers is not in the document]
kingbuzzman: Eh. Simpler, like you said `-vvvvvv` does nothing, and the `manage.py`'s `argparser` limits it already to max `3`, no need to do more.

kingbuzzman:
@bluetech thoughts?

pytest_django/runner.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def __init__(
+        self, verbosity: int = 1, failfast: bool = False, keepdb: bool = False, **kwargs: Any
+    ) -> None:
+        self.verbosity = verbosity
+        self.failfast = failfast
+        self.keepdb = keepdb
+
+    @classmethod
+    def add_arguments(cls, parser: ArgumentParser) -> None:
```
bluetech, line 16:
```suggestion
    def add_arguments(cls, parser: ArgumentParser) -> None:
        parser.add_argument(
            "--failfast",
            action="store_true",
            help="Stops the test suite after the first failure.",
        )
```

tests/test_runner.py
```diff
@@ -0,0 +1,28 @@
+from unittest.mock import Mock, call
+
+import pytest
+
+from pytest_django.runner import PytestTestRunner
+
+
+@pytest.mark.parametrize(
+    *(
```
bluetech, line 9: Can you remove the `*(`? (It seems redundant)

bluetech (commented):
Thanks @kingbuzzman. The PR is well done!

I agree this can be useful in some situations. We can extend support for further `manage.py test` arguments in the future.

I left some comments, PTAL.

pytest_django/runner.py
kingbuzzman: Keep in mind, i applied this, and had to remove it. `failfast` is a standard option already inside of the django `test`, duplicating it causes an error to be raised.

tests/test_runner.py
kingbuzzman: NM... looks like its happy..

kingbuzzman:
@bluetech let me know if there is anything else you want in the PR. Thank you!

pytest_django/runner.py
bluetech: The reason I suggested it is that I saw that Django's `DiscoverRunner` has it, so assumed we need it too. But if not, great!

bluetech (approved):
Thanks again @kingbuzzman!

kingbuzzman:
@bluetech can we close https://github.com/pytest-dev/pytest-django/issues/1169 ?

PR #1186, opened 2025-03-31
Only setup and serialize databases as needed by tests

### LINKED ISSUE #1184: Avoid slow database serialization when possible
I have a large project where the `serialize_db_to_string` function from Django takes around 5s on every call to `pytest`. This DB has some pre-loaded fixtures so it doesn't start the test run empty.

The relevant Django code is:

https://github.com/django/django/blob/02a5cbfe76382da2a0414df17017185be5bd47f9/django/test/utils.py#L231-L239

This is called from pytest-django `django_db_setup` session fixture:

https://github.com/pytest-dev/pytest-django/blob/b7328696ad7a5bfc308b60c909972a9166fce8a3/pytest_django/fixtures.py#L144-L148

We don't pass any of the `serialize` parameters so it defaults to serializing.

The Django `DiscoverRunner` doesn't always serialize:

https://github.com/django/django/blob/02a5cbfe76382da2a0414df17017185be5bd47f9/django/test/runner.py#L1081-L1089

It only serializes if there are any tests in the suite which use `serialized_rollback`:

https://github.com/django/django/blob/02a5cbfe76382da2a0414df17017185be5bd47f9/django/test/runner.py#L1044-L1056

We should try to do the same if possible.

bluetech:
Fix #1184

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,6 +10,17 @@ Compatibility
 * Added official support for Django 5.2 (`PR #1179 <https://github.com/pytest-dev/pytest-django/pull/1179>`__).
 * Dropped testing on MySQL’s MyISAM storage engine (`PR #1180 <https://github.com/pytest-dev/pytest-django/pull/1180>`__).
 
+Bugfixes
+^^^^^^^^
+
+* Stopped setting up and serializing databases on test session setup when not needed (the database is not requested / ``serialized_rollback`` is not used).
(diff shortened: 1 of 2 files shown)


kingbuzzman:
Doesn't this change cause tests to run in an unpredictable order now? -- (because of using `set`s)

bluetech:
I expanded the changelog a bit, I *suspect* this might cause some issues for some clever users, hopefully not.

PR #1187, opened 2025-03-31
Fixes issue with mail.outbox

kingbuzzman:


kingbuzzman:
@bluetech thoughts?

pytest_django/plugin.py
```diff
@@ -599,7 +599,8 @@ def _dj_autoclear_mailbox() -> None:
 
     from django.core import mail
 
-    del mail.outbox[:]
+    if hasattr(mail, "outbox"):
```
bluetech, line 602: This change LGTM. This is an autouse fixture and should just do nothing if locmem backend is not used, not error.

pytest_django/plugin.py
```diff
@@ -608,12 +609,13 @@ def mailoutbox(
     _dj_autoclear_mailbox: None,
 ) -> list[django.core.mail.EmailMessage] | None:
     """A clean email outbox to which Django-generated emails are sent."""
-    if not django_settings_is_configured():
```
bluetech, line 611: This change LGTM. If someone explicitly asked for this fixture and Django is not configured, either it is not meant to run and skip is good, or there's some misconfiguration and then `None` is not really expected.

pytest_django/plugin.py
```diff
@@ -608,12 +609,13 @@ def mailoutbox(
...
-        return None
+    skip_if_no_django()
 
     from django.core import mail
 
-    return mail.outbox  # type: ignore[no-any-return]
+    if hasattr(mail, "outbox"):
+        return mail.outbox  # type: ignore[no-any-return]
+    return None
```
bluetech, line 618: I'm not sure about returning `None`. The situation is that someone explicitly used the `mailoutbox` fixture, but didn't properly configure Django to actually fill it. Do they expect `None`? I think probably not.

So I think we should error in this case, with a proper message telling the user what they need to do to get the `mailoutbox` fixture to work (configure email backend to locmem), or if an error is too harsh, then a warning and return `[]`, this way we at least get rid of the pesky `None` that no one expects. I'd go with the error.

bluetech (commented):
Thanks for tackling this @kingbuzzman. Please see my comments.

pytest_django/plugin.py
kingbuzzman: I was on the fence about this one..

kingbuzzman (commented):
Changes bluetech wanted

bluetech (approved):
Thanks!

bluetech:
@kingbuzzman FYI, I sent you an invite to the pytest-django-developers github team. Doesn't mean you need to do anything, it just will allow you to close issues, merge approved PRs and stuff like that if you'd like.

kingbuzzman:
@bluetech Thank you!

Issue #1155, opened 2024-10-07
No obvious way to set `self.maxDiff` with `pytest_django.asserts`


r-thomson:
When using `pytest_django.asserts.assertXMLEqual`, the diff in the console is truncated, with the following message at the end:

```
Diff is ### characters long. Set self.maxDiff to None to see it.
```

It's not obvious to me what the right way to accomplish this with pytest-django is, since the `TestCase` for these functions is [defined internally](https://github.com/pytest-dev/pytest-django/blob/1ffc3239387bcfadbd8aecfdfacf3d6b52b947f8/pytest_django/asserts.py#L23-L25). Right now I'm just monkeypatching `unittest.case.TestCase.maxDiff`, but it'd be nice to have an obvious way of doing this.

bluetech:
Hmm, thanks for the issue. I figure we should plumb the pytest verbosity options into `maxDiff`.

bluetech:
Looked into it. Unfortunately what the way `pytest_django.asserts` is currently implemented, the assertion functions are just standalone functions, and are not bound to the actual running test. I don't see a way to implement this without a lot of complexity, sorry. I suggest switching to plain `assert` if possible.

roniemartinez:
We can temporarily monkey-patch this in `conftest.py` but looks like an ugly solution

```python
# top of conftest.py
import unittest

unittest.TestCase.maxDiff = None


# ... all other imports here
```

kingbuzzman:
What do you guys think of my naive approach https://github.com/pytest-dev/pytest-django/pull/1191

roniemartinez:
Looks good IMO, but to be honest, I would prefer that this goes to a configuration file (like `pyproject.toml`, `pytest.ini`)

kingbuzzman:
What would that look like @roniemartinez?

```
[tool.pytest.ini_options]
addopts = "..."
max_diff = ??????
```

roniemartinez:
Something like that. But would consult with a Pytest dev since this is for `pytest-django` and I am guessing that should be properly named (see https://pytest-django.readthedocs.io/en/stable/configuring_django.html)

avenet:
Proposed configuration:

```toml
[tool.pytest.ini_options]
django_asserts_max_diff = "None"
```

`None` enables unlimited diffs; the default remains `640`.

The name scopes this to `pytest_django.asserts`, without affecting pytest assertions or users’ Django `TestCase`s. Happy to rename if preferred.

PR #1191, opened 2025-04-06
Exposes the underlining TestCase instance to avoid using "asserts" module

kingbuzzman:
Related to issue: https://github.com/pytest-dev/pytest-django/issues/1155


kingbuzzman:
@bluetech thoughts?

bluetech:
I have a (vain?) hope that someday pytest will be thread safe (i.e. specifically, it should be possible to run multiple pytest instances concurrently in the same process). Currently it isn't, and even pytest-django adds some issues itself. But I think it better to avoid adding *new* ones.

I think if we decide that's it worth putting effort into this issue, we should find a way to make the assert "bind" to the test it is running under. And if there's no clean way to do that, then consider an alternative interface which does make it possible (e.g. a `django_assert` fixture or such which binds to the current test via `request` and exposes the assertions as methods, e.g. `django_assert.templateUsed()`).

kingbuzzman:
> But I think it better to avoid adding new ones.

This makes sense.

> find a way to make the assert "bind" to the test it is running under.

I have some ideas on how to make this work

tests/test_asserts.py
```diff
@@ -71,3 +74,46 @@ def test_sanity() -> None:
...
+def test_fixture_assert(django_testcase: django.test.TestCase) -> None:
+    django_testcase.assertEqual("a", "a")  # noqa: PT009
+
+    with pytest.raises(AssertionError):
+        django_testcase.assertXMLEqual("a" * 10_000, "a")
+
+
+class TestDjangoAssert(django.test.TestCase):
+    def test_fixture_assert(self, django_testcase: django.test.TestCase) -> None:
+        django_testcase.assertEqual("a", "a")  # noqa: PT009
```
kingbuzzman, line 88:
```suggestion
        assert django_test == self
        django_testcase.assertEqual("a", "a")  # noqa: PT009
```

kingbuzzman:
@bluetech thoughts how I can improve this?

bluetech:
@kingbuzzman That's a creative solution, I like it! I will definitely try to review this before the next release.

kingbuzzman:
@adamchainz  understand the making the fixture name smaller, but now sure about how the wrapper would work without frame hacking.. can you provide an example?

kingbuzzman:
What do you guys think of renaming "django_testcase" to just `djtc` it's 4 characters, similar to "self"? I'm open for suggestions

kingbuzzman:
> djt

i'm good with this

the only other thing is remove the warnings, so we don't cause developer stress? i mean, i don't disagree, but i would also love to get rid of this file in say 5 versions from now, or 10... i don't care about this arbitrary number. what is a good plan to deprecate this?

bluetech:
@kingbuzzman, coming from https://github.com/pytest-dev/pytest-django/pull/1223#discussion_r3749697850 and thinking about it a bit, it connects with the idea in this PR, or at least the non-user-facing part of it, of moving the test class into a fixture. I think doing that would help with my idea in the linked comment. So I'll experiment with that. If you want, I can do some rebasing and commit-splitting on this PR.

kingbuzzman:
@bluetech you have my blessing ❤️

PR #1193, opened 2025-04-14
Fix regression: pull in all databases instead of default database

### LINKED ISSUE #1192: No way to specify the database for fixtures
The only way to declare that a fixture requires a database is the `db` fixture. However, since #1186 this will only pull in the default database. As far as I can see, there is currently no way to specify a different database.

xi:
fixes #1192


pytest_django/fixtures.py
```diff
@@ -147,9 +147,7 @@ def _get_databases_for_test(test: pytest.Item) -> tuple[Iterable[str], bool]:
             databases = None
         else:
             return (), False
```
terencehonles, line 149: This code path is also problematic, and what I am running into

PR #1202, opened 2025-05-26
Remove Python 3.8 & Django 5.0

kingbuzzman:
Removes:
* Python 3.8
* Django 5.0
* Updates imports because of ruff linter (the minimum python change)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -71,6 +71,11 @@ jobs:
           python: '3.13'
           allow_failure: false
 
+        # Explicitly test min pytest.
+        - name: py313-dj52-sqlite-pytestmin-coverage
+          python: '3.13'
+          allow_failure: false
+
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -131,15 +131,6 @@ jobs:
           python: '3.11'
           allow_failure: false
 
-        - name: py38-dj42-sqlite-xdist-coverage
-          python: '3.8'
-          allow_failure: false
-
-        # Explicitly test min pytest.
```
bluetech, line 138: We should keep one job testing the `pytestmin`. Best to just bump this one to py39.

bluetech (approved):
Just one comment, otherwise looks good to me, thanks @kingbuzzman! Feel free to merge after the fix.

kingbuzzman:
@bluetech can we release a little minor for this? I can see this being a little annoying for folks.

PR #1203, opened 2025-05-26
Attempt to make codecov less important

kingbuzzman:


.github/workflows/main.yml
```diff
@@ -4,9 +4,9 @@ on:
   push:
     branches:
       - main
+    tags:
+      - "*"
   pull_request:
-    branches:
-      - main
```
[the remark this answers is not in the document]
kingbuzzman: This is moot. There will be no pull requests with tags...

pyproject.toml
```diff
@@ -103,6 +103,9 @@ include = [
     "pytest_django_test/*",
     "tests/*",
 ]
+omit = [
```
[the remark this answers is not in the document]
kingbuzzman: The other one was deleted.

.github/workflows/main.yml
```diff
@@ -50,102 +127,100 @@ jobs:
...
+          find downloaded-coverage-artifacts -type f -name 'coverage.*' -exec cp {} combined_coverage_data/ \;
+          coverage combine combined_coverage_data/*
+          coverage xml
+          coverage html
+          coverage report --format=markdown >> $GITHUB_STEP_SUMMARY
+
+      - name: Determine retention days
+        id: determine-retention-days
+        run: |
+          if [ "${{ github.ref }}" = "refs/heads/main" ] || [[ "${{ github.ref }}" == refs/tags/* ]]; then
```
bluetech, line 185: To avoid [injection attacks](https://docs.zizmor.sh/audits/#template-injection):

```suggestion
          if [[ "${GITHUB_REF}" = "refs/heads/main" ]] || [[ "${GITHUB_REF}" == refs/tags/* ]]; then
```

bluetech (commented):
There are quite a few changes in here, it's a bit hard for me to follow. If you could break them to multiple PRs (or alternatively, atomic commits in a single PR) that would be great!

.github/workflows/main.yml
kingbuzzman: Im going to make sure this doesnt happen again: https://github.com/pytest-dev/pytest-django/pull/1211

kingbuzzman: Woohoo! [Caught it!](https://github.com/pytest-dev/pytest-django/actions/runs/15413919234/job/43372108327?pr=1203#step:7:36)

kingbuzzman: Fixed!

pyproject.toml
```diff
@@ -118,8 +118,9 @@ branch = true
 [tool.coverage.report]
 include = [
     "pytest_django/*",
-    "pytest_django_test/*",
-    "tests/*",
```
webknjaz, line 122: It's actually highly discouraged to skip measuring tests:
https://web.archive.org/web/20250505160753/https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage.html

.github/workflows/main.yml
```diff
@@ -143,11 +147,85 @@ jobs:
...
+
+      - name: Upload HTML coverage report
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-html-report
+          path: htmlcov
+          retention-days: ${{ steps.determine-retention-days.outputs.retention_days }}
+
+      - name: Delete temporary coverage artifacts from run
+        uses: geekyeggo/delete-artifact@f275313  # 5.1.0
```
kingbuzzman, line 212:
```suggestion
        uses: geekyeggo/delete-artifact@ f275313e70c08f6120db482d7a6b98377786765b # 5.1.0
```

pyproject.toml
webknjaz: There should be a Codecov check that demands 100% coverage on the tests so that it's possible to reliably tell that there's no accidentally skipped tests or parts of the testing infra. Otherwise, there's a perception that things are being tested when they aren't.

.github/workflows/main.yml
```diff
@@ -25,6 +25,8 @@ jobs:
     timeout-minutes: 15
     permissions:
       contents: read
+    env:
+      MATRIX_NAME: ${{ matrix.name }}
```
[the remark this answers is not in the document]
kingbuzzman: I like it!

.github/workflows/main.yml
```diff
@@ -54,16 +56,18 @@ jobs:
 
       - name: Run tox
         run: tox -e "${MATRIX_NAME}"
```
kingbuzzman, line 58:
```suggestion
        run: tox
```

pyproject.toml
kingbuzzman: We're headed that direction, its anyone's guess when we arrive 😇

kingbuzzman:
@webknjaz  ?

webknjaz:
> I see extra context on the right in main, but seems a bit useless. what am i missing?

The PR views are different and sometimes inaccurate in Codecov. I recommend always consulting the linked `HEAD` commit page. Commit pages contain the uploads lists that are downloadable and linked to the corresponding jobs. Plus, PR views often misrepresent covered lines in patch widgets.

webknjaz:
> what am i missing from the codecov report, the summed up report vs individual ^

@kingbuzzman you're missing flags that can tell you exactly what environments covered each line. You should start setting them, though.

.github/workflows/main.yml
```diff
@@ -151,11 +155,85 @@ jobs:
...
+          include-hidden-files: true
+
+      - name: Upload HTML coverage report
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-html-report
+          path: htmlcov
+          retention-days: ${{ steps.determine-retention-days.outputs.retention_days }}
+
+      - name: Delete temporary coverage artifacts from run
```
webknjaz, line 219: This isn't really that necessary, as you already set them to expire in one day. Plus, it'd be impossible to inspect these files whenever you suspect something's wrong.

kingbuzzman: True. But its just noise, i figure i'd leave it "clean" -- but you're not wrong.

.github/workflows/main.yml
```diff
@@ -65,13 +65,17 @@ jobs:
...
-        uses: codecov/codecov-action@v5
+        run: mv .coverage coverage.${TOXENV}
+
+      - name: Upload temporary coverage artifact
+        if: contains(matrix.name, 'coverage')
+        uses: actions/upload-artifact@v4
         with:
-          fail_ci_if_error: true
-          files: ./coverage.xml
-          token: ${{ secrets.CODECOV_TOKEN }}
```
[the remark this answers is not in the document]
kingbuzzman: This is cool, i'm playing with it now, thanks!

.github/workflows/main.yml
webknjaz: It's just that I've recently seen a DIY implementation of this causing CI stability issues due to GH API or network being flaky and causing chaos. That memory is still fresh :)

But also, these are about inspectability.

If you want fewer artifacts, you can run `actions/upload-artifact/merge` and it'll do the deletion for you, replacing these with one artifact containing all the individual files.

kingbuzzman: Now that i like! Thanks!

kingbuzzman:
@webknjaz [this is why](https://github.com/pytest-dev/pytest-django/actions/runs/16937568450/job/47997777345?pr=1230) i want to remove codecov

most of the time it works. but it always breaks in the most inopportune times...

kingbuzzman:
@webknjaz from my understanding is that the token is expiring because the upload is taking long [randomly]

[see how it normally takes 2s](https://github.com/pytest-dev/pytest-django/actions/runs/16937568450/job/47997777311?pr=1230#step:9:1) vs [how it took 4m32s](https://github.com/pytest-dev/pytest-django/actions/runs/16937568450/job/47997777345?pr=1230#step:9:307) -- and because of _that_ the token expired.

PR #1204, opened 2025-05-26
Fix ruff PT001 warning

kingbuzzman:
[Fixes pytest warning](https://docs.astral.sh/ruff/rules/pytest-fixture-incorrect-parentheses-style/) to not declare our fixtures like `@pytest.mark.fixtures()` instead do `@pytest.mark.fixtures`


kingbuzzman:
Low hanging fruit!

PR #1208, opened 2025-06-02
pyproject.toml: switch from extras to dependency groups

bluetech:
See https://packaging.python.org/en/latest/specifications/dependency-groups/ More suitable for internal stuff than extras which are exposed externally.

Also moves all development dependency declarations from tox.ini to be in a single place, pyproject.toml.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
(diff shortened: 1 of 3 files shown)


kingbuzzman:
I think this broke something in CI https://github.com/pytest-dev/pytest-django/actions/runs/15401963387/job/43336475787

kingbuzzman:
After 24mins i had to cancel it manually

PR #1223, opened 2025-08-12
Add support for db fixture (test inside transaction) for asyncio tests

lode-braced:
In using and debugging pytest-django with async tests & fixtures, I ended up hacking together a way to get the db fixture working by enabling the transaction in the sync to async executor thread in which async orm queries run. 

This PR aims to integrate that hack as actual functionality into pytest-django.

The async work is done, I also worked to refine the sync side database access: not allowing db access to threads other than the main thread which is running the test (& transaction). That second part is causing some test failures I've yet been unable to fix on 3.9 & django 4.2, will try to fix or drop it from the PR.

**edit**: I managed to fix the issue with the test failures: my monkeypatches were dropping the self attribute. Ready for review.

Suggestions & feedback welcome.

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -203,6 +203,60 @@ def django_db_setup(  # noqa: PLR0917
                 )
 
 
+def _build_pytest_django_test_case(
+    *,
+    reset_sequences: bool,
+    serialized_rollback: bool,
+    databases: _DjangoDbDatabases,
+    available_apps: _DjangoDbAvailableApps,
+    transactional: bool,
+) -> type[django.test.TestCase]:
+    import django.test
+
+    if transactional:
+        test_case_class = django.test.TransactionTestCase
+    else:
+        test_case_class = django.test.TestCase
+
+    _reset_sequences = reset_sequences
+    _serialized_rollback = serialized_rollback
+    _databases = databases
+    _available_apps = available_apps
+
+    class PytestDjangoTestCase(test_case_class):  # type: ignore[misc,valid-type]
+        reset_sequences = _reset_sequences
+        serialized_rollback = _serialized_rollback
+        if _databases is not None:
+            databases = _databases
+        if _available_apps is not None:
+            available_apps = _available_apps
+
+        # For non-transactional tests, skip executing `django.test.TestCase`'s
+        # `setUpClass`/`tearDownClass`, only execute the super class ones.
+        #
+        # `TestCase`'s class setup manages the `setUpTestData`/class-level
+        # transaction functionality. We don't use it; instead we (will) offer
+        # our own alternatives. So it only adds overhead, and does some things
+        # which conflict with our (planned) functionality, particularly, it
+        # closes all database connections in `tearDownClass` which inhibits
+        # wrapping tests in higher-scoped transactions.
+        #
+        # It's possible a new version of Django will add some unrelated
+        # functionality to these methods, in which case skipping them completely
+        # would not be desirable. Let's cross that bridge when we get there...
+        if not transactional:
+
+            @classmethod
+            def setUpClass(cls) -> None:
+                super(django.test.TestCase, cls).setUpClass()
+
+            @classmethod
+            def tearDownClass(cls) -> None:
+                super(django.test.TestCase, cls).tearDownClass()
+
+    return PytestDjangoTestCase
+
+
 @pytest.fixture
 def _django_db_helper(
     request: pytest.FixtureRequest,
@@ -242,49 +296,13 @@ def _django_db_helper(
     )
 
     with django_db_blocker.unblock():
-        import django.db
-        import django.test
-
-        if transactional:
-            test_case_class = django.test.TransactionTestCase
-        else:
-            test_case_class = django.test.TestCase
-
-        _reset_sequences = reset_sequences
-        _serialized_rollback = serialized_rollback
-        _databases = databases
-        _available_apps = available_apps
-
-        class PytestDjangoTestCase(test_case_class):  # type: ignore[misc,valid-type]
-            reset_sequences = _reset_sequences
-            serialized_rollback = _serialized_rollback
-            if _databases is not None:
-                databases = _databases
-            if _available_apps is not None:
-                available_apps = _available_apps
-
-            # For non-transactional tests, skip executing `django.test.TestCase`'s
-            # `setUpClass`/`tearDownClass`, only execute the super class ones.
-            #
-            # `TestCase`'s class setup manages the `setUpTestData`/class-level
-            # transaction functionality. We don't use it; instead we (will) offer
-            # our own alternatives. So it only adds overhead, and does some things
-            # which conflict with our (planned) functionality, particularly, it
-            # closes all database connections in `tearDownClass` which inhibits
-            # wrapping tests in higher-scoped transactions.
-            #
-            # It's possible a new version of Django will add some unrelated
-            # functionality to these methods, in which case skipping them completely
-            # would not be desirable. Let's cross that bridge when we get there...
-            if not transactional:
-
-                @classmethod
-                def setUpClass(cls) -> None:
-                    super(django.test.TestCase, cls).setUpClass()
-
-                @classmethod
-                def tearDownClass(cls) -> None:
-                    super(django.test.TestCase, cls).tearDownClass()
+        PytestDjangoTestCase = _build_pytest_django_test_case(
+            reset_sequences=reset_sequences,
+            serialized_rollback=serialized_rollback,
+            databases=databases,
+            available_apps=available_apps,
+            transactional=transactional,
+        )
 
         PytestDjangoTestCase.setUpClass()
 
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import os
-from collections.abc import Callable, Generator, Iterable, Sequence
+from collections.abc import AsyncGenerator, Callable, Generator, Iterable, Sequence
 from contextlib import AbstractContextManager, contextmanager
 from functools import partial
 from typing import TYPE_CHECKING, Any, Protocol
@@ -258,7 +258,7 @@ def _build_pytest_django_test_case(
 
 
 @pytest.fixture
-def _django_db_helper(
+def _sync_django_db_helper(
     request: pytest.FixtureRequest,
     django_db_setup: None,  # noqa: ARG001
     django_db_blocker: DjangoDbBlocker,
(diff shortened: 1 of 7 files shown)


kingbuzzman (commented):
*This is not a code review*

From what I can briefly see; this looks great, i've personally had this issue come up, so happy you've taken this on!

tox.ini
```diff
@@ -10,6 +10,7 @@ envlist =
 [testenv]
 dependency_groups =
     testing
+				!dj42: async
```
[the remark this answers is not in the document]
kingbuzzman: Apologies, @lode-braced can you please sync with `main` once more.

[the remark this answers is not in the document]
lode-braced: done
... and fixed the resulting linting issue too ;)

tests/test_async_db.py
```diff
@@ -0,0 +1,52 @@
...
+
+else:
+    pytestmark = pytest.mark.asyncio
+    fixturemark = cast(MarkDecorator, pytest_asyncio.fixture)
+
+
+@pytest.mark.parametrize("run_number", [1, 2])
+@pytestmark
+@pytest.mark.django_db
+async def test_async_db(db, run_number) -> None:
```
kingbuzzman, line 25: Not needed
```suggestion
async def test_async_db(run_number) -> None:
```

tests/test_async_db.py
```diff
@@ -0,0 +1,52 @@
...
+    # test async database usage remains isolated between tests
+
+    assert await Item.objects.acount() == 0
+    # make a new item instance, to be rolled back by the transaction wrapper before the next parametrized run
+    await Item.objects.acreate(name="blah")
+    assert await Item.objects.acount() == 1
+
+
+@fixturemark
+async def db_item(db) -> Any:
```
kingbuzzman, line 35: Not needed
```suggestion
async def db_item() -> Any:
```

kingbuzzman (commented):
5/8 files reviewed. More to follow.

tests/test_db_thread_safeguards.py
```diff
@@ -0,0 +1,36 @@
...
+        except BaseException as exc:  # noqa: BLE001 - we want to capture exactly what is raised
+            captured[0] = exc
+
+    t = threading.Thread(target=worker)
+    t.start()
+    t.join()
+
+    assert captured[0] is not None, "Expected DB access in worker thread to raise an exception"
+    assert isinstance(captured[0], RuntimeError)
+    assert "only allowed in the main thread" in str(captured[0])
```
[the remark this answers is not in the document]
lode-braced: That's fair, I wrote this code a bit too rushed :)

Will clean it up in the new PR once these changes are finalized

pytest_django/plugin.py
```diff
@@ -831,18 +834,60 @@ def _dj_db_wrapper(self) -> django.db.backends.base.base.BaseDatabaseWrapper:
...
+                "modify your test fixtures to be async or use the transactional_db fixture."
+                "See https://pytest-django.readthedocs.io/en/latest/database.html#db-thread-safeguards for more information."
+            )
+        elif self._real_ensure_connection is not None:
+            self._real_ensure_connection(wrapper_self, *args, **kwargs)
+
+    def _unblocked_sync_only(self, wrapper_self: Any, *args, **kwargs):
+        __tracebackhide__ = True
+        if threading.current_thread() != threading.main_thread():
+            raise RuntimeError(
```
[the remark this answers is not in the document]
lode-braced: I'll make the second PR after finishing up work with this one.

lode-braced:
@kingbuzzman apologies for the silence, I've been on holiday. The PR should have addressed your remarks where possible, and I've removed the sync code, to be done in another PR once we get this one done. 

Can you have another look and close topics as needed?

bluetech:
I rebased this PR (before review). Had to fix some minor conflicts.

bluetech:
I split the `_build_pytest_django_test_case` refactor to a separate commit to make the 2nd commit easier to review. I will try to review tomorrow.

bluetech:
Thanks for the PR @lode-braced. I hoped to merge it for v4.12.0, but yeah, this is a complex topic -- need to get up to speed with asyncio, pytest-asyncio, asgiref, and how Django's own TestCase handles async, before I can evaluate the proposal in this PR. (methinks async was the biggest Python mistake ever, but that's besides the point :grinning: )

Here is my understanding so far after a few minutes of looking into it, which might not be entirely accurate yet:

The reason this PR is needed in the first place is that just using

```py
@pytest.mark.asyncio
async def test_it(db):
    await Item.objects.acreate(name="foo")
```

doesn't rollback the DB changes at the end of the test, i.e. the new `Item` sticks around.

The reason for that is (I'm mostly guessing here) the DB transaction BEGIN/ROLLBACK that `setUpClass` executes runs on a different thread than the `Item.objects.create`. And that's because the `Item.objects.acreate` doesn't have an `AsyncToSync` context to latch onto, so the `thread_sensitive` stuff isn't effective.

What Django's TestCase does is just wrap the test method itself with `AsyncToSync`, leaving all of the setups and teardowns sync. What this PR does instead (for async tests) is make the entire `_django_db_helper` async, but using `SyncToAsync` on the Django setups and teardowns (since they are sync). An advantage of this approach is that pytest fixtures can also be async and share the context.

pytest_django/fixtures.py
```diff
@@ -303,6 +321,106 @@ def tearDownClass(cls) -> None:
...
+    return transactional, reset_sequences, databases, serialized_rollback, available_apps
+
+
+@pytest.fixture
+def _django_db_helper(
+    request: pytest.FixtureRequest,
+    django_db_setup: None,  # noqa: ARG001
+    django_db_blocker: DjangoDbBlocker,  # noqa: ARG001
+) -> None:
+    asyncio_marker = request.node.get_closest_marker("asyncio")
```
bluetech, line 412: If we're going all in on pytest-asyncio, would be better to use [`pytest_asyncio.is_async_test(request.node)`](https://pytest-asyncio.readthedocs.io/en/stable/reference/functions.html#is-async-test) instead.

Although there are several async-support plugins for pytest, e.g. `anyio`. It *would* be nice to be agnostic, but maybe not possible.

lode-braced:
> Thanks for the PR @lode-braced. I hoped to merge it for v4.12.0, but yeah, this is a complex topic -- need to get up to speed with asyncio, pytest-asyncio, asgiref, and how Django's own TestCase handles async, before I can evaluate the proposal in this PR. (methinks async was the biggest Python mistake ever, but that's besides the point 😀 )
> 
> Here is my understanding so far after a few minutes of looking into it, which might not be entirely accurate yet:
> 
> The reason this PR is needed in the first place is that just using
> 
> ```python
> @pytest.mark.asyncio
> async def test_it(db):
>     await Item.objects.acreate(name="foo")
> ```
> 
> doesn't rollback the DB changes at the end of the test, i.e. the new `Item` sticks around.
> 
> The reason for that is (I'm mostly guessing here) the DB transaction BEGIN/ROLLBACK that `setUpClass` executes runs on a different thread than the `Item.objects.create`. And that's because the `Item.objects.acreate` doesn't have an `AsyncToSync` context to latch onto, so the `thread_sensitive` stuff isn't effective.
> 
> What Django's TestCase does is just wrap the test method itself with `AsyncToSync`, leaving all of the setups and teardowns sync. What this PR does instead (for async tests) is make the entire `_django_db_helper` async, but using `SyncToAsync` on the Django setups and teardowns (since they are sync). An advantage of this approach is that pytest fixtures can also be async and share the context.

Hi, thanks for looking at it! I'm happy to see some movement on it.

As a followup on your message:

The root problem is indeed the threads, but I can already save you some time on the _reason_ it runs on different threads, to supplement/correct your guesses: Django's testcase doesn't wrap the method with `async_to_sync`, as far as I could see at the time of writing the PR, it was setup to only run tests in sync scenarios. 

pytest-asyncio is what wraps _some_ work in an async loop, and once in an async loop, the async to sync takes its one thread pool executor which it shares in the loop's context, makes _one_ _new_ sync thread to run al the orm queries in.

In practice, the tests have two threads, running as follows:

T1 (the main thread): run sync fixtures, and async event loops for async fixtures/async tests: the async event loop is started in the main thread and "takes it over" to run the async event loop until complete. Then, Django orm work in the async event loop in T1 spawns a _new_, shared thread pool executor of size 1, on T2.

Since the django db setup is a sync fixture in pytest django, and that fixture starts the transaction, T1 has a transaction, T2 does not.

Two relevant things:

1. I do think async is quite neat, but do understand it can be a curve. If you'd like to schedule a call where I talk you through the parts that make up the problem here, happy to do so and spread the good word of async ;)
2. I built a better fix in https://github.com/lode-braced/pytest-django-asyncio, where instead of trying to make the transaction setup happen in T2, I made T1 and T2 share the same _connection_. Reason being that that also allows the use of sync fixtures to use the django orm for an async text, and given the explanation of the two threads above: you know that T1 is busy running the event loop while T2 is able to  run requests, so you don't have two threads *using* the db connection at the same time, which is the thing to avoid due to it being thread safe. I'll port over this _new_ fix into a new PR here too, up to you if you want to go with what is in this PR first or go with what I believe to be the now better option.

bluetech:
@lode-braced Thanks for the extra details, I will try to understand them soon :)

> Django's testcase doesn't wrap the method with async_to_sync, as far as I could see at the time of writing the PR, it was setup to only run tests in sync scenarios.

Just to note on this, Django has been doing this since async support was added:

https://github.com/django/django/blob/ae25a40be07e8a749edf526df37c93e59d4a22c9/django/test/testcases.py#L351-L353

I do think it's meant more as a convenience than a comprehensive solution.

lode-braced:
> https://github.com/django/django/blob/ae25a40be07e8a749edf526df37c93e59d4a22c9/django/test/testcases.py#L351-L353

Oh, TIL! Must have gotten my memory of the test setup code (sync) confused with the actual async test methods being allowed.

tests/test_async_db.py
```diff
@@ -0,0 +1,48 @@
...
+
+try:
+    import pytest_asyncio
+except ImportError:
+    pytestmark = pytest.mark.skip("pytest-asyncio is not installed")
+    fixturemark = pytest.mark.skip("pytest-asyncio is not installed")
+
+else:
+    pytestmark = pytest.mark.asyncio
+    fixturemark = cast(pytest.MarkDecorator, pytest_asyncio.fixture)
```
bluetech, line 18: I think instead of this granular skipping, we should skip the entire test file.

This can be done by calling `pytest.skip("pytest-asyncio is not installed", allow_module_level=True)` in the `except`. And then use `@pytest_asyncio.fixture` as normal to make the code more direct.

tests/test_async_db.py
```diff
@@ -0,0 +1,48 @@
...
+
+import pytest
+
+from pytest_django_test.app.models import Item
+
+
+try:
+    import pytest_asyncio
+except ImportError:
+    pytestmark = pytest.mark.skip("pytest-asyncio is not installed")
```
bluetech, line 13: BTW, not sure if you intended it, but `pytestmark` has a special meaning: https://docs.pytest.org/en/latest/reference/reference.html#globalvar-pytestmark

tests/test_async_db.py
```diff
@@ -0,0 +1,48 @@
...
+async def db_item() -> Any:
+    return await Item.objects.acreate(name="async")
+
+
+@pytest.fixture
+def sync_db_item() -> Any:
+    return Item.objects.create(name="sync")
+
+
+@pytest.mark.usefixtures("db_item", "sync_db_item")
```
bluetech, line 43: Why use `usefixtures` instead of request directly?

pytest_django/fixtures.py
```diff
@@ -303,6 +321,106 @@ def tearDownClass(cls) -> None:
...
+                reset_sequences=reset_sequences,
+                serialized_rollback=serialized_rollback,
+                databases=databases,
+                available_apps=available_apps,
+                transactional=False,
+            )
+
+            from asgiref.sync import sync_to_async
+
+            await sync_to_async(PytestDjangoTestCase.setUpClass)()
```
bluetech, line 360: I have a vague idea that, since pytest-asyncio and similar plugins already the sync/async-fixtures-within-async-test case by running them in the appropriate event loop, if we can make this stuff "just work" by running the django setup/teardown fixtures as pytest fixtures.

I'll see if it's viable and get back to you.

PR #1225, opened 2025-08-12
Create .editorconfig

kingbuzzman:
Adds:
* .editorconfig for auto setting tabs/auto trimming up of files
* editorconfig-checker action to check that all the rules are validated correctly
* Fixes all the files that the checker identified as an issue
  * I gave up on the rst for now, as that would be a LOT of changes -- this is a TODO!!!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,15 +19,6 @@ env:
 permissions: {}
 
 jobs:
-  editorconfig:
-    runs-on: ubuntu-24.04
-    steps:
-      - uses: actions/checkout@v5
-        with:
-          persist-credentials: false
-      - uses: editorconfig-checker/action-editorconfig-checker@0ab3427a3f09b17c21d489172245be511e0a16ea # v2
-      - run: editorconfig-checker
-
   test:
     runs-on: ubuntu-24.04
     continue-on-error: ${{ matrix.allow_failure }}
@@ -165,7 +156,6 @@ jobs:
 
     needs:
     - test
-    - editorconfig
 
     runs-on: ubuntu-24.04
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,7 @@ xdist = [
     "pytest-xdist",
 ]
 linting = [
+    "editorconfig-checker==3.2.1",
     "mypy==1.15.0",
     "ruff==0.9.5",
     "zizmor==1.9.0",
--- a/tox.ini
+++ b/tox.ini
@@ -48,6 +48,7 @@ commands =
     ruff check --diff {posargs:pytest_django pytest_django_test tests}
     ruff format --quiet --diff {posargs:pytest_django pytest_django_test tests}
     mypy {posargs:pytest_django pytest_django_test tests}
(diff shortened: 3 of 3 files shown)


tests/test_django_settings_module.py
```diff
@@ -195,19 +197,25 @@ def test_django_settings_configure(
...
-                                   'ENGINE': 'django.db.backends.sqlite3',
-                                   'NAME': ':memory:'
-                               }},
-                               INSTALLED_APPS=['django.contrib.auth',
-                                               'django.contrib.contenttypes',])
-
-            import pytest
-
-            pytest.main()
+        """
```
kingbuzzman, line 200:
```suggestion
        run="""
```

webknjaz:
I'd recommend putting the checker into tox or pre-commit configs since those would have better cross-env integrations and aren't locked to just the CI.

webknjaz:
It has a native integration with the pre-commit framework: https://github.com/editorconfig-checker/editorconfig-checker/blob/main/.pre-commit-hooks.yaml. So just add a `.pre-commit-config.yaml` pointing to the repo. Call `pre-commit run --all-files` in `tox.ini` and make sure you have a job calling that tox env in CI.

In my projects, it looks like this: https://github.com/ansible/awx-plugins/actions/runs/16881280962/job/47817646031.

Overall, having pre-commit invoked in CI through tox probably deserves a dedicated PR. Once it's in, adding any new linters and/or formatters would be going through the pre-commit framework and would work in CI, locally through tox and even if editor integrations somehow call pre-commit outside this infra.
Additionally, I recommend enabling https://pre-commit.ci in this repo to get a PR integration with formatting back-pushes.

webknjaz:
> PS. I did this quick PR to ensure the authors would catch issues like putting [tabs into places were the norm is spaces](https://github.com/pytest-dev/pytest-django/pull/1223/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R48) and I or others not to have to point it out.

Then, I'd definitelly go for the pre-commit framework as it has native fixers for this sort of things: https://github.com/ansible/awx-plugins/blob/12091e4/.pre-commit-config.yaml#L65-L68. Check out https://pre-commit.com/hooks.html for more fun stuff.

kingbuzzman:
@webknjaz I see, I get it; but @bluetech is going to have to authorize pre-commit.ci

Does it make sense to merge this in the interim, while we wait for the integration? Seems to me that the only thing we would remove is the changes to .github/workflows/main.yml -- everything else would stay in place. Plus, adding the hook config files.

What do you think?

webknjaz:
> is going to have to authorize pre-commit.ci

Note that his is a nice to have. Using it in GHA through tox isn't blocked as the GH App isn't needed for that. I'm rather surprised that there's no linting job in CI already.

kingbuzzman:
> Also, there's a wrapper that does work with pip install: [pypi.org/p/editorconfig-checker](https://pypi.org/p/editorconfig-checker). But I'd still go the pre-commit route. Though, maybe it's easier to use this in the scope of this PR until there's pre-commit based infra.

OH wow. I completely missed this! -- ok ok, let me patch it 😇

PR #1229, opened 2025-08-13
Fix CI HTTP 429 error

kingbuzzman:


kingbuzzman:
@webknjaz looks like readthedocs fixed it on their end. [main](https://github.com/pytest-dev/pytest-django/actions/runs/16937299256/job/47996897349) looks like its working fine now.

PR #1230, opened 2025-08-13
Update linter/test dependencies

kingbuzzman:


kingbuzzman:
The tests that are failing is because the readthedoc and codecov are both behaving badly today... pushing through

PR #1233, opened 2025-08-13
Fix: Make URL cache clearing thread-safe in `_django_set_urlconf`

pankaj-bind:
### Problem

When running tests in parallel with `pytest-xdist`, the `@pytest.mark.urls` fixture can cause a race condition. Multiple test workers may try to modify the global `ROOT_URLCONF` and clear the URL cache at the same time, leading to unpredictable behavior and intermittent test failures.

### Solution

This change introduces a `threading.Lock()` around the URL configuration and cache-clearing logic in the `_django_set_urlconf` fixture. This ensures that these operations are atomic, preventing multiple threads from interfering with each other and making the URL cache clearing thread-safe.

### DIFF
--- a/tests/test_urls.py
+++ b/tests/test_urls.py
@@ -107,3 +107,46 @@ def test_urls_cache_is_cleared_and_new_urls_can_be_assigned(
 
     result = django_pytester.runpytest_subprocess()
     assert result.ret == 0
+
+@pytest.mark.django_project(
+    extra_settings="""
+    ROOT_URLCONF = "empty"
+    """
+)
+def test_urls_concurrent(django_pytester: DjangoPytester) -> None:
+    "Test that the URL cache clearing is thread-safe with pytest-xdist."
+    pytest.importorskip("xdist")
+
+    django_pytester.makepyfile(
+        empty="urlpatterns = []",
+        urls1="""
+        from django.urls import path
+        urlpatterns = [path('url1/', lambda r: None, name='url1')]
+        """,
+        urls2="""
+        from django.urls import path
+        urlpatterns = [path('url2/', lambda r: None, name='url2')]
+        """,
+    )
+
+    django_pytester.create_test_module(
+        """
+        import pytest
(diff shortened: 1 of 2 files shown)


tests/test_urls.py
```diff
@@ -107,3 +107,47 @@ def test_something_else():
...
+    extra_settings="""
+    ROOT_URLCONF = "empty"
+    """
+)
+def test_urls_concurrent(django_pytester: DjangoPytester) -> None:
+    "Test that the URL cache clearing is thread-safe with pytest-xdist."
+    pytest.importorskip("xdist")
+
+    django_pytester.makepyfile(
+        empty="urlpatterns = []",
```
kingbuzzman, line 122: Not used
```suggestion
```

tests/test_urls.py
```diff
@@ -107,3 +107,47 @@ def test_something_else():
...
     result = django_pytester.runpytest_subprocess()
     assert result.ret == 0
+
+
+@pytest.mark.django_project(
+    extra_settings="""
+    ROOT_URLCONF = "empty"
+    """
+)
+def test_urls_concurrent(django_pytester: DjangoPytester) -> None:
```
kingbuzzman, line 117: I copied this test over to my `main` branch, and ran this on py3.13

```shell
count=0
while true; do
    count=$((count + 1))
    if ! pytest tests/test_urls.py -vv > /dev/null 2>&1; then
        echo "Test failed on iteration $count"
        break
    fi
    printf "."
done
```

it has not failed -- keeps going, after 100+ iterations (keep in mind, i only copied the test, not the "fix")

kingbuzzman (changes requested):
If your test is anything to go off of, this does not show a problem currently in the code as it stand. Can you elaborate more?

tests/test_urls.py
kingbuzzman: PS. I also ran this with `pytest -n auto` (same loop as above) -- no difference, i dont see the race condition

PR #1267, opened 2026-02-10
Adds 3.14 Free Threading support

kingbuzzman:
Removed  added 3.14 Free Threading support.
Removed `psycopg[binary]` in favor of `psycopg` (the full python implementation) to avoid issues with 3.14t


bluetech (commented):
Thanks @kingbuzzman.

Regarding 3.14t, I think https://github.com/pytest-dev/pluggy/pull/621#pullrequestreview-3387378019 is the current approach in pytest. I think until pytest itself adds 3.14t we should hold on it in pytest-django.

That said, pytest-django is pure python, 3.14t should work perfectly fine, as long as not trying to run pytest/pytest-django in multiple threads, as they are not thread safe.

Dropping Django 5.1 is good though.

PR #1271, opened 2026-02-26
Fix type hints of assert functions

mschoettle:
Updated type hints in assert functions to use Iterable instead of Collection. And have assertRaisesMessage return something to use as a context manager.

This is in line with `django-stubs`: https://github.com/typeddjango/django-stubs/blob/master/django-stubs/test/testcases.pyi

Version 4.12.0 raises the following errors with `mypy`:

```console
error: "assertRaisesMessage" does not return a value (it only ever returns None)  [func-returns-value]

error: Argument 2 to "assertQuerySetEqual" has incompatible type "QuerySet[Foo, Foo]"; expected "Collection[Any]"  [arg-type]
note: "QuerySet" is missing following "Collection" protocol member:
note:     __contains__
```

### DIFF
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -25,7 +25,7 @@ else:
     test_case = TestCase("run")
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Collection, Iterable, Iterator, Sequence
+    from collections.abc import Callable, Iterable, Iterator, Sequence
     from contextlib import AbstractContextManager
     from typing import Any, overload
 
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -25,7 +25,7 @@ else:
     test_case = TestCase("run")
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


kingbuzzman (approved):
LGTM, ~[looks like you got it from here 😉 ](https://github.com/typeddjango/django-stubs/blob/560137fcf9a3876e5b1da47982075bf2037f1405/django-stubs/test/testcases.pyi#L135)~ ah, right you added it in your description, eek. As you can see, I read code first then descriptions... foot, meet mouth. 😇

kingbuzzman:
Seems low risk. Merging.

mschoettle:
> Hum, if you want a release, can you make another PR and change the changelog please. This would go a long way in getting the release ready. Thats on me, I should have asked for this in the first place.

See #1274. Although someone will have to update it once the release date is known

mschoettle:
@bluetech Any chance we can get a release with this fix soon?

PR #1278, opened 2026-04-14
Complete type stubs for `assertTemplateUsed` / `assertTemplateNotUsed`

brianhelba:
mypy 1.20 (python/mypy#20350) changes how it types the return value of functions that return `None`: previously it used `Any` (suppressing downstream errors), now it propagates the actual `None` type. This means the existing `-> None` annotation on `assertTemplateUsed` / `assertTemplateNotUsed` causes type errors when these functions are used as context managers.

This adds `@overload` signatures to distinguish the two calling conventions, matching the pattern already used by `assertNumQueries` in this file:

* `response` is `HttpResponseBase`: direct assertion, returns `None`
* `response` is a `str` (shorthand for `template_name`): returns context manager
* `response` is `None` (`template_name` passed by keyword): returns context manager

### DIFF
--- a/pytest_django/asserts.py
+++ b/pytest_django/asserts.py
@@ -112,18 +112,69 @@ if TYPE_CHECKING:
         msg_prefix: str = ...,
     ) -> None: ...
 
+    # with assertTemplateUsed("template.html"): ...  # noqa: ERA001
+    @overload
+    def assertTemplateUsed(
+        response: str,
+        template_name: None = ...,
+        msg_prefix: str = ...,
+        count: int | None = ...,
+    ) -> AbstractContextManager[None]: ...
+
+    # with assertTemplateUsed(template_name="template.html"): ...  # noqa: ERA001
+    @overload
+    def assertTemplateUsed(
+        response: None = ...,
+        template_name: str | None = ...,
+        msg_prefix: str = ...,
+        count: int | None = ...,
+    ) -> AbstractContextManager[None]: ...
+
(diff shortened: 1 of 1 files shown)


brianhelba:
@kingbuzzman Please take a look.

brianhelba:
@kingbuzzman Could you take a look? This is a typing-only bugfix, to add support for Mypy 1.20+.

pytest_django/asserts.py
```diff
@@ -112,18 +112,69 @@ def assertFormSetError(
...
     ) -> None: ...
 
+    # with assertTemplateUsed("template.html"): ...  # noqa: ERA001
+    @overload
+    def assertTemplateUsed(
+        response: str,
+        template_name: None = ...,
+        msg_prefix: str = ...,
+        count: int | None = ...,
+    ) -> AbstractContextManager[None]: ...
```
bluetech, line 122: Technically it's not `None`, the context manager returns itself. I'll change that to `Any`.

bluetech (approved):
Thanks!

Consider also submitting this improvement to django-stubs, it can improve the typing there as well: https://github.com/typeddjango/django-stubs/blob/61a3af749523e9645861057588fe67230ac6b6dd/django-stubs/test/testcases.pyi#L119-L125

brianhelba:
@bluetech Thanks!

I took your suggestion to fix upstream too: https://github.com/typeddjango/django-stubs/pull/3564

PR #1280, opened 2026-05-21
feat: re-export Settings from pytest_django for typing (#1257)

SAY-5:
Closes #1257.

`SettingsWrapper` is now importable from the top-level `pytest_django` namespace so the `settings` fixture can be type-annotated without reaching into `pytest_django.fixtures`:

```python
from pytest_django import SettingsWrapper

def test_with_specific_settings(settings: SettingsWrapper):
    ...
```

Mirrors the existing pattern used for `DjangoAssertNumQueries` and `DjangoCaptureOnCommitCallbacks`.

- adds `SettingsWrapper` to `pytest_django.__all__`
- documents the typing usage under the `settings` fixture in `docs/helpers.rst`
- changelog entry under v4.12.1
- regression test in `TestSettings`

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,13 +1,20 @@
 Changelog
 =========
 
-v4.12.1 (unreleased)
+v4.13.0 (unreleased)
 --------------------
 
 Compatibility
 ^^^^^^^^^^^^^
 
 * Dropped support for Django 4.2 and 5.1.
+
+Improvements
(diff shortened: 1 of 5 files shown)


bluetech (approved):
Thanks!

I updated the PR according to my comments in the issue: https://github.com/pytest-dev/pytest-django/issues/1257#issuecomment-5167443336

PR #1275, opened 2026-03-07
Fix `--help`/`--version` when conftest imports models

brianhelba:
Previously, `pytest_load_initial_conftests` returned early when `--help` or `--version` was passed (added by #238), skipping `django.setup()` entirely. This caused conftest files with top-level Django model imports to fail with `AppRegistryNotReady`.

Instead of returning early, continue with normal Django initialization, allowing `--help`/`--version` to return properly when everything is valid.

If Django is somehow invalid (e.g. incorrect `DJANGO_SETTINGS_MODULE`), add a guard to still return gracefully if `--help`/`--version` is requested. Running real tests would still fail, but these basic informational arguments shouldn't be blocked.

Fixes #1152

PR #1285, opened 2026-07-28
Fix `--help`/`--version` when a conftest imports Django models

alimony:
## Problem

Running `pytest --help` or `pytest --version` fails to load a `conftest.py` that imports Django models at import time:

pytest.PytestConfigWarning: could not load initial conftests: .../conftest.py

The underlying cause is `AppRegistryNotReady: Apps aren't loaded yet.`, and it becomes a hard error under `filterwarnings = error`. Running the tests themselves works fine, because that path sets Django up first.

## Cause

`pytest_load_initial_conftests` returned early for `--help`/`--version` – added to keep these options working with an invalid `DJANGO_SETTINGS_MODULE` (#235) – which also skipped `django.setup()`. pytest still imports the initial conftests for these options, so any top-level model import then hits `AppRegistryNotReady`.

## Fix

Continue with normal Django initialization for `--help`/`--version` so conftests import cleanly. If Django can't be configured or set up (e.g. an invalid `DJANGO_SETTINGS_MODULE`), the failure is tolerated *only* for these options, so they keep working regardless of Django's state – preserving #235. A real test run still fails loudly.

## Tests

Adds a regression test in `tests/test_initialization.py` covering both `--help` and `--version`. It asserts a positive "conftest imported" marker rather than the absence of a warning, so it also covers `--version`, where pytest records but never prints the conftest-load failure.

## Notes

This overlaps with #1275, which addresses the same issue. Two differences: the error tolerance here catches any setup failure (not only `ImportError`), so a settings module that imports but is improperly configured also won't break `--help`; and the regression test guards `--version` explicitly.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -6,6 +6,7 @@ test database and provides some useful text fixtures.
 
 from __future__ import annotations
 
+import argparse
 import contextlib
 import inspect
 import os
@@ -312,9 +313,36 @@ def pytest_load_initial_conftests(
 
     options = parser.parse_known_args(args)
 
-    if options.version or options.help:
-        return
+    # pytest still imports the initial conftests for `--help`/`--version`, so
+    # Django is set up for them as well. Otherwise a conftest with top-level
+    # model imports fails with AppRegistryNotReady (issue #1152).
+    is_help_or_version = bool(options.version or options.help)
 
+    # Stashed up front, to be available even if the setup below fails.
+    report_header: list[str] = []
+    early_config.stash[report_header_key] = report_header
+    early_config.stash[blocking_manager_key] = DjangoDbBlocker(_ispytest=True)
+
+    try:
+        _initialize_django(early_config, options, args, report_header)
+    except Exception:
+        # `--help`/`--version` never run tests, so they must keep working on a
+        # broken configuration (issue #235); a real run still fails loudly.
+        if not is_help_or_version:
+            raise
+
+
+def _initialize_django(
+    early_config: pytest.Config,
+    options: argparse.Namespace,
+    args: list[str],
+    report_header: list[str],
+) -> None:
+    """Configure and set up Django from the pytest options/ini/environment.
+
+    Everything which can fail on a broken configuration lives here, so that
+    `pytest_load_initial_conftests` can tolerate it for `--help`/`--version`.
+    """
     django_find_project = _get_boolean_value(
         early_config.getini("django_find_project"), "django_find_project"
     )
@@ -347,9 +375,6 @@ def pytest_load_initial_conftests(
     ds, ds_source = _get_option_with_source(options.ds, SETTINGS_MODULE_ENV)
     dc, dc_source = _get_option_with_source(options.dc, CONFIGURATION_ENV)
 
-    report_header: list[str] = []
-    early_config.stash[report_header_key] = report_header
-
     if ds:
         report_header.append(f"settings: {ds} (from {ds_source})")
         os.environ[SETTINGS_MODULE_ENV] = ds
@@ -370,8 +395,7 @@ def pytest_load_initial_conftests(
         with _handle_import_error(_django_project_scan_outcome):
             dj_settings.DATABASES  # noqa: B018
 
-    early_config.stash[blocking_manager_key] = DjangoDbBlocker(_ispytest=True)
-
+    # Populates the app registry, which fails on a broken INSTALLED_APPS.
     _setup_django(early_config)
 
 
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -1,5 +1,7 @@
 from textwrap import dedent
 
+import pytest
+
 from .helpers import DjangoPytester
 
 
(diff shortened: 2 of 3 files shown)


alimony:
It seems like all PRs are currently blocked by a Django 6.2 incompatibility. I have a fix for it here: https://github.com/pytest-dev/pytest-django/pull/1286

kingbuzzman:
Please sync master

alimony:
> Please sync master

Done!

pytest_django/plugin.py
```diff
@@ -350,29 +355,36 @@ def _get_option_with_source(
...
-            dj_settings.DATABASES  # noqa: B018
+            with _handle_import_error(_django_project_scan_outcome):
+                dj_settings.DATABASES  # noqa: B018
 
-    early_config.stash[blocking_manager_key] = DjangoDbBlocker(_ispytest=True)
+        early_config.stash[blocking_manager_key] = DjangoDbBlocker(_ispytest=True)
 
-    _setup_django(early_config)
+        _setup_django(early_config)
+    except Exception:
```
kingbuzzman: Can we make this try/except block smaller? where does it actually fail?

alimony:
@kingbuzzman

I reworked the try block a little bit. Only two things can actually fail:

1. `dj_settings.DATABASES` – the forced settings load (`ImportError`/`ImproperlyConfigured` when DSM is missing or blows up on import – the original #235 case)
2. `_setup_django()` → `django.setup()` – settings load fine but `INSTALLED_APPS` is broken

The rest (header appends, `os.environ` writes, the stash) can't raise, so I moved the stash above the `try`. Bonus: no more re-indentation in the diff. Note `import configurations.importer` is now outside the `try` – let me know if you want it back in.

tests/test_initialization.py
```diff
@@ -60,3 +62,40 @@ def test_ds():
...
+    # conftests (pytest #13574), so it wouldn't exercise this path; passing
+    # it twice forces full startup, which imports the initial conftests.
+    args = [option, option] if option == "--version" else [option]
+    result = django_pytester.runpytest_subprocess(*args)
+
+    output = result.stdout.str() + result.stderr.str()
+    assert result.ret == 0
+    assert "conftest-imported-models" in output
+    assert "AppRegistryNotReady" not in output
+    assert "could not load initial conftests" not in output
```
kingbuzzman: Can we use `result.stdout.fnmatch_lines(..)` instead of the `... in output`?

alimony: Fixed.

pytest_django/plugin.py
```diff
@@ -363,16 +366,25 @@ def _get_option_with_source(
...
-        # ImproperlyConfigured if settings cannot be loaded.
-        from django.conf import settings as dj_settings
-
-        with _handle_import_error(_django_project_scan_outcome):
-            dj_settings.DATABASES  # noqa: B018
-
     early_config.stash[blocking_manager_key] = DjangoDbBlocker(_ispytest=True)
 
-    _setup_django(early_config)
+    try:
```
kingbuzzman: Tests still pass if i remove the try/catch
```diff
diff --git a/pytest_django/plugin.py b/pytest_django/plugin.py
index f8fe21e..9c1f43d 100644
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -368,23 +368,23 @@ def pytest_load_initial_conftests(
 
     early_config.stash[blocking_manager_key] = DjangoDbBlocker(_ispytest=True)
 
-    try:
-        if ds:
-            # Forcefully load Django settings, throws ImportError or
-            # ImproperlyConfigured if settings cannot be loaded.
-            from django.conf import settings as dj_settings
+    # try:
+    if ds:
+        # Forcefully load Django settings, throws ImportError or
+        # ImproperlyConfigured if settings cannot be loaded.
+        from django.conf import settings as dj_settings
 
-            with _handle_import_error(_django_project_scan_outcome):
-                dj_settings.DATABASES  # noqa: B018
-
-        # Populates the app registry, which fails on a broken INSTALLED_APPS.
-        _setup_django(early_config)
-    except Exception:
-        # `--help`/`--version` only print information and never run tests, so a
-        # broken Django configuration is tolerated here (issue #235); anything
-        # genuinely wrong still surfaces on a real test run.
-        if not is_help_or_version:
-            raise
+        with _handle_import_error(_django_project_scan_outcome):
+            dj_settings.DATABASES  # noqa: B018
+
+    # Populates the app registry, which fails on a broken INSTALLED_APPS.
+    _setup_django(early_config)
+    # except Exception:
+    #     # `--help`/`--version` only print information and never run tests, so a
+    #     # broken Django configuration is tolerated here (issue #235); anything
+    #     # genuinely wrong still surfaces on a real test run.
+    #     if not is_help_or_version:
+    #         raise
 
 
 @pytest.hookimpl(trylast=True)
```

kingbuzzman (changes requested):
Tested your changes, and your tests pass works without your try/except block which is concerning, please take a look.

alimony:
> Tested your changes, and your tests pass works without your try/except block which is concerning, please take a look.

I rearranged things and restructured both implementation (moving initialisation to its own function) and tests, try it again now.

bluetech (approved):
Thanks, this solution seems reasonable for now. The fundamental problem here is that the conftest using django is initial (loaded early, instead of during collection), but it rarely needs to be. But pytest doesn't currently provide ways to control this, whether a conftest is initial is just based on where it's located. Hopefully we can improve this in pytest core some day, then the fix would be for the user to make the conftest non-initial, instead of pytest-django ignoring errors.

Modern pytest actually handles `--version` early itself these days, but `--help` does load initial conftests since they might affect the help output.

I added @brianhelba from #1275 as co-author, hopefully you don't mind.

PR #1286, opened 2026-07-28
Handle Django 6.2 `ImproperlyConfigured` for an invalid DJANGO_SETTINGS_MODULE

alimony:
On Django 6.2 (the `djmain` CI jobs), an invalid `DJANGO_SETTINGS_MODULE` raises `django.core.exceptions.ImproperlyConfigured` instead of `ImportError`. `_handle_import_error` only caught `ImportError`, so on 6.2 the settings-import failure escaped it — pytest-django's guidance message ("pytest-django found a Django project…" / "did not search…") was dropped, and the error type changed.

This caught three tests that assert on that output: `test_ds_non_existent`, `test_django_project_found_invalid_settings`, and `test_django_project_scan_disabled_invalid_settings`.

Fix: catch `ImproperlyConfigured` alongside `ImportError` and re-raise as `ImportError` (as before), so the guidance is shown and behavior is consistent across Django versions. No test assertions changed. Verified against Django 4.2–6.2.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -8,6 +8,7 @@ Bugfixes
 ^^^^^^^^
 
 * Fixed type hints of assert methods to match actual signature (`PR #1271 <https://github.com/pytest-dev/pytest-django/pull/1271>`__)
+* Handled Django 6.2's ``ImproperlyConfigured`` (in addition to ``ImportError``) when the configured ``DJANGO_SETTINGS_MODULE`` cannot be imported, so pytest-django still shows its guidance message.
 
 v4.12.0 (2026-02-14)
 --------------------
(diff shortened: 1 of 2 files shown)


kingbuzzman (approved):

alimony:
@kingbuzzman "6.2" here is Django main: `stable/6.1.x` was forked at feature freeze, so main moved on to `VERSION = (6, 2, 0, "alpha", 0)`, and the `djmain` tox env installs main directly.

The confusing part is the version string:

```console
$ pip install https://github.com/django/django/archive/main.tar.gz
$ python -c "import django; print(django.get_version(), django.VERSION)"
6.2 (6, 2, 0, 'alpha', 0)
```

Sorry for the confusing title, "Django main" would have been clearer.

kingbuzzman:
I noticed the 6.2 and deleted my comment, It's fine. Thank you for the contribution  ❤️

PR #1293, opened 2026-08-01
Don't run `_pre_setup()` twice for transactional tests

vokracko:
Since Django 5.2, `TransactionTestCase.setUpClass()` calls `cls._pre_setup()` eagerly and sets `_pre_setup_ran_eagerly` so that whoever drives the test case can skip its own call, the way `unittest.TestCase.run()` does. `_django_db_helper()` calls `setUpClass()` and then `test_case._pre_setup()` unconditionally, so every `django_db(transaction=True)` test has its fixtures set up twice: the `serialized_rollback` restore, the `fixtures` load and the sequence reset all run again. On a suite of 121 transactional tests that was 567s against 445s, roughly 1s per test.

The first commit adds a test recording which frames call `_pre_setup()`. It passes on current `main` and asserts the duplicate: a plain `django_db` test records one call, a transactional one records `setUpClass` and then a second call from `_django_db_helper()`.

The second commit consumes `_pre_setup_ran_eagerly` in `_django_db_helper()` and flips that assertion to the single call. Verified against Django 5.2; the eager call is unchanged on 6.0 and main.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,7 @@ Bugfixes
 
 * Fixed type hints of assert methods to match actual signature (`PR #1271 <https://github.com/pytest-dev/pytest-django/pull/1271>`__)
 * Handled Django 6.2's ``ImproperlyConfigured`` (in addition to ``ImportError``) when the configured ``DJANGO_SETTINGS_MODULE`` cannot be imported, so pytest-django still shows its guidance message.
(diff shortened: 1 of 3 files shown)


kingbuzzman:
This makes sense to me. Verified your details, and this looks like a good change. Thank you!

kingbuzzman (approved):

PR #1294, opened 2026-08-03
Add missing type annotations for non_debugging_runtest

atiqur-rahman-pro:
…n_debugging_runtest

This PR fixes missing type annotations identified during strict type checking:

- **`pytest_django/asserts.py`**: Added missing parameter (`func: Callable[..., Any] | None = ...`) and return type (`-> AbstractContextManager[None] | None`) annotations to the `assertNumQueries` overload implementation.
- **`pytest_django/plugin.py`**: Added explicit type annotation (`self: TestCaseFunction`) for `non_debugging_runtest` parameter and removed unnecessary `noqa: ANN001`.

All mypy strict checks pass cleanly.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -588,7 +588,7 @@ def _django_setup_unittest(
 
     original_runtest = TestCaseFunction.runtest
 
-    def non_debugging_runtest(self) -> None:  # noqa: ANN001
+    def non_debugging_runtest(self: TestCaseFunction) -> None:
         self._testcase(result=self)
 
     from django.test import SimpleTestCase


kingbuzzman:
While the change looks good on the surface, can you provide an example where it was failing? (real world use case?)

atiqur-rahman-pro:
Hi @kingbuzzman,

Thanks for reviewing! The primary use case occurs when running **Mypy in strict mode (`mypy --strict` / `--disallow-untyped-defs`)** on downstream user codebases or type-checked pytest plugins.

### Real-world Example

Consider a Django project configured with `strict = true` in `pyproject.toml` or `mypy.ini`:

```python
# tests/test_queries.py
import pytest
from pytest_django.asserts import assertNumQueries
from myapp.models import User

def test_user_query_count() -> None:
    # 1. Context manager usage with strict typing
    with assertNumQueries(2):
        User.objects.filter(is_active=True).count()

    # 2. Callable invocation usage
    assertNumQueries(1, lambda: User.objects.get(id=1)) 

Mypy Error Output:

error: Function is missing a type annotation for parameter "func"  [no-untyped-def]
error: Function is missing a return type annotation  [no-untyped-def]

File: pytest_django/asserts.py [hunk 1]
```diff
@@ -227,11 +227,11 @@ def assertNumQueries(
 
     def assertNumQueries(
         num: int,
-        func=...,
+        func: Callable[..., Any] | None = ...,
```
bluetech: This should be an `@overload` rather than a param with default.

pytest_django/asserts.py
```diff
@@ -227,11 +227,11 @@ def assertNumQueries(
...
         *args: Any,
         using: str = ...,
         **kwargs: Any,
-    ): ...
+    ) -> AbstractContextManager[None] | None: ...
```
bluetech: The context manager is not exactly `None`, it actually returns itself. So `Any` would be better for now.

bluetech (commented):
Thanks for trying to improve this.

Regarding the `assertNumQueries`, if we want to be completely accurate we need to do something like django-stubs: https://github.com/typeddjango/django-stubs/blob/61a3af749523e9645861057588fe67230ac6b6dd/django-stubs/test/testcases.pyi#L176-L181. But that's somewhat complex, so I punted it. But `AbstractContextManager[Any]` would side-step the main complexity and still be an improvement.

The `non_debugging_runtest` looks good.

kingbuzzman (approved):
LGTM, but the name of the PR needs to change. Thoughts?

PR #1296, opened 2026-08-03
Update mypy dependencies

kingbuzzman:
- Bump mypy version
- Change type ignore comments from 'misc' to 'untyped-decorator' for clarity in test cases.


bluetech (approved):
Thanks!

PR #1298, opened 2026-08-03
Add django-upgrade dependency

kingbuzzman:
- Added `django-upgrade==1.31.1` to the linting dependencies in `pyproject.toml`.
- Updated `tox.ini` to include a command for running `django-upgrade` on specified directories.

These changes enhance the linting process by incorporating Django upgrade checks.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,6 +70,7 @@ xdist = [
     "pytest-xdist",
 ]
 linting = [
+    "django-upgrade==1.31.1",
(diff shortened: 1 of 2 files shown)


bluetech:
I'm not sure django-upgrade is a good fit here, since we need to be somewhat careful about our Django usage. So normally when Django adds a deprecation warning, we handle it with a conditional or such.

kingbuzzman:
I'll remove it, my thought here was more so we can catch it before hand. But I see your point

PR #1302, opened 2026-08-03
Refactor test_change_settings to simplify parameters

kingbuzzman:
Removed unnecessary parameters from test_change_settings.


kingbuzzman:
@bluetech am i missing something here? Why is `settings` needed here to begin with?

bluetech (commented):
Since the test is exactly the same as the one above, I assumed it was somehow part of the test.  So we should either leave it as-is (just in case it's intentional), or remove it entirely.

PR #1304, opened 2026-08-06
Fix `_pre_setup_ran_eagerly` on Django 5.0 and 5.1

### LINKED ISSUE #1303: [4.13.0] AttributeError: type object 'PytestDjangoTestCase' has no attribute '_pre_setup_ran_eagerly'
After upgrading from 4.12.0 I got a regression, now 395 of my tests fail with:

```
[3](https://github.com/wemake-services/django-modern-rest/actions/runs/31078669060/job/92542226654?pr=1215#step:7:6294)
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_acreate_token_without_expiry - AttributeError: type object 'PytestDjangoTestCase' has no attribute '_pre_setup_ran_eagerly'
```

https://github.com/wemake-services/django-modern-rest/actions/runs/31078669060/job/92542226654?pr=1215

This is Django version specific: only tests with Django 5.0 and Django 5.1 fail.
I see that the official support was dropped, but maybe we can still fix it? 
Packages that still support older Django versions would need either to stay on the older package version or to also drop older versions.

Will changing `if not PytestDjangoTestCase._pre_setup_ran_eagerly:` to `if not getattr(PytestDjangoTestCase, '_pre_setup_ran_eagerly', False):` solve it?

sobolevn:
Closes https://github.com/pytest-dev/pytest-django/issues/1303

Locally I double checked that:
- It now works with 5.0 and 5.1
- `False` is the correct default, with `True` I get errors like

```
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_create_token_explicit_expiry - AttributeError: 'PytestDjangoTestCase' object has no attribute 'atomics'
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_token_is_expired - AttributeError: 'PytestDjangoTestCase' object has no attribute 'atomics'
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_token_revoke - AttributeError: 'PytestDjangoTestCase' object has no attribute 'atomics'
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_hash_token_is_stored - AttributeError: 'PytestDjangoTestCase' object has no attribute 'atomics'
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_create_token_without_expiry - AttributeError: 'PytestDjangoTestCase' object has no attribute 'atomics'
ERROR tests/test_unit/test_security/test_token/test_app/test_token_model.py::test_create_token - AttributeError: 'PytestDjangoTestCase' object has no attribute 'atomics'
```

### DIFF
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -213,6 +213,8 @@ def _django_db_helper(
         yield
         return
 
+    from django import VERSION
+
     marker = request.node.get_closest_marker("django_db")
     if marker:
         (
@@ -289,7 +291,11 @@ def _django_db_helper(
         PytestDjangoTestCase.setUpClass()
 
         test_case = PytestDjangoTestCase(methodName="__init__")
-        if not PytestDjangoTestCase._pre_setup_ran_eagerly:
+        if VERSION >= (6, 0):
+            pre_setup_ran_eagerly = PytestDjangoTestCase._pre_setup_ran_eagerly
+        else:
+            pre_setup_ran_eagerly = getattr(PytestDjangoTestCase, "_pre_setup_ran_eagerly", False)
+        if not pre_setup_ran_eagerly:
             # For a TransactionTestCase, setUpClass() has already run _pre_setup() and set
             # this flag to say so.
             test_case._pre_setup()
--- a/pytest_django/fixtures.py
+++ b/pytest_django/fixtures.py
@@ -289,7 +289,7 @@ def _django_db_helper(
         PytestDjangoTestCase.setUpClass()
 
         test_case = PytestDjangoTestCase(methodName="__init__")
-        if not PytestDjangoTestCase._pre_setup_ran_eagerly:
+        if not getattr(PytestDjangoTestCase, "_pre_setup_ran_eagerly", False):
             # For a TransactionTestCase, setUpClass() has already run _pre_setup() and set
             # this flag to say so.
             test_case._pre_setup()


bluetech (approved):

kingbuzzman:
Im confused, why are we supporting 5.1/5.0?  [Support was dropped 8+ months ago](https://endoflife.date/django)

image

sobolevn:
Because it is a single line change and it makes live of many opensource projects that rely on `pytest-django` and older Django versions much easier :)

kingbuzzman:
While I agree with the sentiment, my main concern is that this sets a bad precedent. You now have a line of code living there that will likely be forgotten about until someone stumbles across it three years from now and asks, “Is it safe to remove this yet?”

More broadly, what’s the right lifecycle for a workaround like this? If the maintainer of the upstream package has already deemed the affected version outdated and unsupported, how long should we continue carrying a fix for it on our side? At some point, we risk accumulating compatibility code with no clear criteria for when it should be removed.

@bluetech

bluetech:
@kingbuzzman You are right, this is just a small courtesy.

For the risk of accumulating code, I modified @sobolevn PR slightly to make it conditional on an explicit django.VERSION check, which we occasionally check for and remove old stuff.

But the bigger thing is that pytest-django didn't really provide a nice way for users who support older Django versions to get a compatible pytest-django version. For that reason, I added in the latest version the [`django` extra](https://pytest-django.readthedocs.io/en/stable/tutorial.html#step-1-installation). So in the future we could tell users "add the django extra and your dependency resolver will do the right thing". But it can't help with existing pytest-django versions, only for future complainers :)

BTW, Django just changes to annual releases and all releases from now on will be LTS. That will simplify things a bit (no more support "gaps"). https://www.djangoproject.com/weblog/2026/aug/10/annual-release-cycle/

PR #1305, opened 2026-08-06
Support database transactions in async tests

### LINKED ISSUE #580: Database transactions with asyncio
The newly released Channels 2.0 Django package "relies on the `asyncio` library and native Python async support". [They recommend](http://channels.readthedocs.io/en/latest/topics/testing.html) to use `py.test` tests with the `asyncio` plugin.

Now, you can use the `db` and `django_db` fixtures with `asyncio` (when you use Channels' `database_sync_to_async` wrapper):

```python
def create_my_model():
    return MyModel.objects.create(name='foo')

@pytest.fixture
async def add_my_model_to_db(db):
    return await database_sync_to_async(create_my_model)()

@pytest.mark.asyncio
async def test_my_db_modification(add_my_model_to_db):
    assert add_my_model_to_db.name == 'foo'
    # ...
```

However, **transactions don't seem to work**. As a result all changes to the test database will be kept.

"Note that you can’t mix this with `unittest.TestCase` subclasses" (which `pytest-django` does internally – but I don't know if this is of importance concerning this issue, because [another `asyncio` issue](https://github.com/pytest-dev/pytest-asyncio/issues/75) seems to be related).

Are there plans to support transactions in async tests? Or should this be done in another package?

lode-braced:
Fixes #580.

Supersedes #1223.

# What

Share Django database connections between pytest's main thread and asgiref's thread-sensitive executor for pytest-asyncio tests.

This keeps direct async ORM calls, synchronous fixtures, and Django `AsyncClient` requests inside the transaction managed by pytest-django's `db` fixture.

# Why

pytest-django opens the test transaction on the main thread, while Django executes async ORM work on a synchronous executor thread. Because Django connections are normally thread-local, that work can use a different connection and escape the transaction rollback.

# Deviations

Unlike #1223, database setup and teardown remain synchronous on the main thread. Synchronous fixtures are also left on the main thread rather than moved onto the executor. The connection itself is temporarily shared between both threads.

Support is intentionally scoped to pytest-asyncio's default thread-sensitive executor, as used by direct async ORM calls and Django's test `AsyncClient`. Arbitrary user-created `ThreadSensitiveContext` executors and production ASGI execution are outside this PR's scope.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+Next
+----
+
+Bugfixes
+^^^^^^^^
+
+* Fixed database transaction isolation for asyncio tests using Django's async
+  ORM (`#580 <https://github.com/pytest-dev/pytest-django/issues/580>`__).
+
 v4.13.0 (2026-08-06)
 --------------------
 
--- a/docs/database.rst
+++ b/docs/database.rst
@@ -60,6 +60,30 @@ select using an argument to the ``django_db`` mark::
    def test_spam():
        pass  # test relying on transactions
 
+
+Async tests
+-----------
+
+Database access also works in async tests managed by
+`pytest-asyncio <https://pytest-asyncio.readthedocs.io/>`_. Mark the test for
+both asyncio and database access, then use Django's async ORM methods::
+
+   @pytest.mark.asyncio
+   @pytest.mark.django_db
+   async def test_async_user():
+       user = await User.objects.acreate(username="me")
+       assert user.username == "me"
+
+The :fixture:`db` fixture still wraps the test in a transaction and rolls it
+back afterward. Django runs async ORM operations on a dedicated synchronous
+executor thread. For an asyncio test, ``pytest-django`` shares the database
+connection between the main thread, where fixtures and transaction setup run,
+and that executor thread. This means synchronous fixtures and async test code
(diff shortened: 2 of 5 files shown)


lode-braced:
@bluetech here's the alternate, newer version of fixing interaction with pytest-asyncio, by having the main thread and the async thread pool executor share one connection.

This one allows for more "it just works" behaviour for users, allowing them to combine sync and async fixtures as they please. In my opinion the better experience, but up to you to compare it against #1223 and see what you like best. Note the previous pr is a lot larger as I am quite defensive there: given that sync/async fixture combining did not work there, I wanted to make sure we inform the users instead of having "silent weird behaviours" as it is currently for async tests.

bluetech:
This one seems to access very internal/private parts of `Local` and django's `connections`. That seems undesirable to me, for two reasons:
- It can break without notice
- It would be hard to anticipate the effects of messing with internals

But more to the code itself, I'm not sure I fully follow it, but IIUC even though the main & executor threads are different, they will still share a connection? That seems dangerous, since the `connections` `Local` is defined as "thread critical" for a reason, which is that least *some* DB backends (e.g. sqlite in [some configurations](https://docs.python.org/3/library/sqlite3.html#sqlite3.threadsafety)) require a connection to be pinned to the thread that created it.

PR #1309, opened 2026-08-16
Fix `mailoutbox` fixture initialization with database-dependent fixtures

### LINKED ISSUE #589: `mail.outbox` not being set?
Hey there, thanks for pytest-django! Appreciate all of the hard work, have used it quite a few times with no problems.

We're running into a bit of a snag where we upgraded some things and suddenly `mail.outbox` isn't retaining emails during tests. I've tried to figure out how exactly to use the `mailoutbox` fixture as documented [here](http://pytest-django.readthedocs.io/en/latest/helpers.html#mailoutbox) but I am having trouble using this.

Right now we're testing in a class and I have tried quite a few configurations to pass `mailoutbox` fixture. I.e. I updated the ini:

```ini
[pytest]
usefixtures = mailoutbox
```

Not exactly sure how to use that fixture though, I tried adding the positional argument `mailoutbox` to my class methods, still not able to access it.. I am sure I am missing something simple.

Would appreciate a push in the right direction -- and again, thanks so much for this project!

gh640:
Fixes #589

Ensure database setup completes before `mailoutbox` captures Django's `mail.outbox` to prevent `mailoutbox` from retaining a stale list when Django's `TestCase._pre_setup()` (that is invoked in `_django_db_helper` fixture) replaces `mail.outbox`.

---

This PR intentionally covers only database dependencies that are visible in pytest's fixture graph. It does not cover fixtures that dynamically request database access after `mailoutbox` has been initialized, for example:

```python
@pytest.fixture
def dynamically_requests_db(request):
    request.getfixturevalue("db")
```

Supporting that case would require a more invasive approach, such as preserving and synchronizing the existing outbox around `_django_db_helper` and Django's `_pre_setup()`. I am happy to change to the broader approach if maintainers prefer it.

### DIFF
--- a/pytest_django/plugin.py
+++ b/pytest_django/plugin.py
@@ -646,11 +646,17 @@ def _dj_autoclear_mailbox() -> None:
 
 @pytest.fixture
 def mailoutbox(
+    request: pytest.FixtureRequest,
     django_mail_patch_dns: None,  # noqa: ARG001
     _dj_autoclear_mailbox: None,
 ) -> list[django.core.mail.EmailMessage] | None:
     """A clean email outbox to which Django-generated emails are sent."""
     skip_if_no_django()
+
+    # Django's TestCase._pre_setup() replaces mail.outbox. Run the database
+    # helper first so mailoutbox returns the replacement list.
+    if "_django_db_helper" in request.fixturenames:
+        request.getfixturevalue("_django_db_helper")
 
     from django.core import mail
 
--- a/tests/test_fixtures.py
+++ b/tests/test_fixtures.py
@@ -804,6 +804,23 @@ def test_mail_again(mailoutbox) -> None:
     test_mail(mailoutbox)
 
 
+def test_mailoutbox_with_db(mailoutbox, db: None) -> None:  # noqa: ARG001
+    assert mailoutbox is mail.outbox
+
+
+@pytest.fixture
+def mail_sending_db_fixture(db: None) -> None:  # noqa: ARG001
+    mail.send_mail("subject", "body", "from@example.com", ["to@example.com"])
+
+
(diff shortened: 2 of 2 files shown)


pytest_django/plugin.py
```diff
@@ -646,12 +646,18 @@ def _dj_autoclear_mailbox() -> None:
...
     django_mail_patch_dns: None,  # noqa: ARG001
     _dj_autoclear_mailbox: None,
 ) -> list[django.core.mail.EmailMessage] | None:
     """A clean email outbox to which Django-generated emails are sent."""
     skip_if_no_django()
 
+    # Django's TestCase._pre_setup() replaces mail.outbox. Run the database
+    # helper first so mailoutbox returns the replacement list.
+    if "_django_db_helper" in request.fixturenames:
+        request.getfixturevalue("_django_db_helper")
```
kingbuzzman, line 659: Darn you're [right](https://github.com/django/django/blob/main/django/test/testcases.py#L388), i'm tempted to not do this and set the `mail.outbox = []` myself here, to _not_ instantiate the db.

@bluetech thoughts?

tests/test_fixtures.py
```diff
@@ -804,6 +804,23 @@ def test_mail_again(mailoutbox) -> None:
     test_mail(mailoutbox)
 
 
+def test_mailoutbox_with_db(mailoutbox, db: None) -> None:  # noqa: ARG001
```
kingbuzzman, line 807:
```suggestion
def test_mailoutbox_without_db(mailoutbox) -> None:  # noqa: ARG001
```

kingbuzzman: We should have this test to make sure we don't regress yet again 😇

PR #1310, opened 2026-08-25
Warn when --reuse-db is used with an in-memory sqlite database

### LINKED ISSUE #160: Raise warning when using SQLite3 in-memory test database with --reuse-db
Currently I've running several unittests with py.test that needs a database. The tests working fine.
But I#ve running with `--reuse-db`and the following database settings:

```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': 'db.sqlite3'
    }
}
```

The Tests will work fine, but the databse db.sqlite3 won't getting created / the file is missing while the tests run / after the tests.
I use django unittests and run py.test with the following: `py.test -v -s --ds=envisia.test.settings --reuse-db` also I'm running on Python 3.4.1, pytest 2.6.1 and Django 1.7rc2.
On PostgreSQL the database will be created and will exists..

Edit: Also PostgreSQL tests run faster even when I run sqlite3 on `:memory:` which is somehow strange?! But I think that's cause of PostgreSQL --reuse-db works.

ChrisJr404:
Closes #160.

Using `--reuse-db` with an in-memory sqlite database is a no-op: the database only lives for the duration of the process, so there's nothing to reuse or keep between runs. This trips people up (the default sqlite test database is in-memory), which is what the issue describes, and both Andreas and Daniel agreed a warning would be the right nudge.

So this emits a `PytestWarning` from `django_db_setup` when `--reuse-db` is requested (and `--create-db` isn't) and any of the databases being set up is an in-memory sqlite one. It only fires when you actually pass `--reuse-db`, so default runs are unaffected. Uses the same `vendor == "sqlite" and is_in_memory_db()` check already used for the live server.

Added tests for both the warning firing and staying quiet without the flag, plus a changelog line and a short note in the database docs.


pytest_django/fixtures.py
```diff
@@ -191,6 +191,24 @@ def django_db_setup(  # noqa: PLR0917
             **setup_databases_args,
         )
 
+    if django_db_keepdb and not django_db_createdb:
+        from django.db import connections
+
+        in_memory = sorted(
+            alias
+            for alias in aliases
+            if connections[alias].vendor == "sqlite" and connections[alias].is_in_memory_db()
```
kingbuzzman, line 200: [Wouldn't this by itself work?](https://github.com/django/django/blob/fdfbb711e3dd52683ae633f568761efe2735d5f0/django/db/backends/sqlite3/base.py#L333)
```suggestion
            if connections[alias].is_in_memory_db()
```

pytest_django/fixtures.py
```diff
@@ -191,6 +191,24 @@ def django_db_setup(  # noqa: PLR0917
             **setup_databases_args,
         )
 
+    if django_db_keepdb and not django_db_createdb:
+        from django.db import connections
+
+        in_memory = sorted(
```
kingbuzzman: To make it a bit clearer.
```suggestion
        in_memory_aliases = sorted(
```

pytest_django/fixtures.py
```diff
@@ -191,6 +191,24 @@ def django_db_setup(  # noqa: PLR0917
...
+            alias
+            for alias in aliases
+            if connections[alias].vendor == "sqlite" and connections[alias].is_in_memory_db()
+        )
+        if in_memory:
+            request.node.warn(
+                pytest.PytestWarning(
+                    "--reuse-db has no effect for the in-memory sqlite database(s) "
+                    f"{', '.join(repr(alias) for alias in in_memory)}: an in-memory "
```
kingbuzzman: Why the `repr`? Aren't these always [strings](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/conf/global_settings.pyi#L90) (and [here](https://github.com/typeddjango/django-stubs/blob/master/django-stubs/db/utils.pyi#L12))?

pytest_django/fixtures.py
```diff
@@ -191,6 +191,24 @@ def django_db_setup(  # noqa: PLR0917
...
         )
 
+    if django_db_keepdb and not django_db_createdb:
+        from django.db import connections
+
+        in_memory = sorted(
+            alias
+            for alias in aliases
+            if connections[alias].vendor == "sqlite" and connections[alias].is_in_memory_db()
+        )
```
kingbuzzman: or maybe we can try
```suggestion
        in_memory = any(
            connections[alias].is_in_memory_db()
            for alias in aliases
        )
```

pytest_django/fixtures.py
kingbuzzman: Nevermind, this is tightly coupled with sqlite's `DatabaseWrapper`

ChrisJr404:
Renamed the variable to `in_memory_aliases` as suggested.

On the `repr`: yes, the aliases are always strings, so it will not fail. I kept `repr` deliberately so the alias names come out quoted in the warning (e.g. `'default'`), which reads more clearly than bare names when several are listed. Happy to drop it to a plain join if you'd prefer the unquoted form.

